PR #43, opened 2016-03-06
Issue 41 - part 1

rurtle:
To fix history command for TSH


ppenna:
Implementation not complete.

PR #50, opened 2016-04-23
Implementing max size in region

Theldus:
With this pull, the maximum size of a region is dynamically defined as the region grows or shrinks.

Thus, it can be avoided in future implementations that the kernel be in doubt which region belongs to a particular address.


ppenna:
This partially fiz the problem, Nevertheless, I'll accept the merge.

Theldus:
I know, sorry about that, soon I will send another pull with the rest of the solution.

PR #51, opened 2016-04-26
Reorganizing init calls

rurtle:
Reorganizing initialization function calls by moving clock & FPU initialization as part of device initialization.


rurtle:
Missed minor changes in one of the files. Will modify & resend pull request.

PR #55, opened 2016-05-20
cmos-rtc

rurtle:
Fixed a bug so Nanvix is able to read CMOS register values properly. Added log in TTY driver to display bootup time (only supports GMT at the moment). Will remove this log once time system call is functional.


ppenna:
Merged in commit 0a662cf8b03aa0821baf08cef0148a267bcd3d6b.

PR #56, opened 2016-06-03
Time syscall

rurtle:
Time system call implementation - initial draft.


ppenna:
Pull request merged in commit f97c60b1dec395b0662aa17af361dae3bffde454.

Some bugs regarding the time() system call implementation were fixed, in the process:
        1. Bad return value using pointers in kernel land.
        2. Not checking for a valid memory region area in userspace.

PR #61, opened 2016-06-25
VBE Support on Nanvix

Theldus:
With this pull, Nanvix is now able to support VESA BIOS Extensions and is able to running at 1024x768 and 32-bit color, besides that, Nanvix continues to support the 80x25 mode and this is set as the default.

To use the VBE standard, just change the line
`#define VIDEO_MODE 0`
for
`#define VIDEO_MODE 1`, in the nanvix/config.h file.

To achieve this, some things have to be modified and/or reviewed, as:
- New address to be mapped in every process in the system.
- Rework the system console to support both video modes and not make it complex.
- Provide some way to draw pixels and characters on the screen.

and so on...

Other interesting points to highlight:
- The performance will be always slower than the 80x25, which is natural, but it is noticeable in a more clearly way in Bochs, so I highly recommend clear the screen when it fills instead of relying on the scrolling. Scrolling is a super heavy task and I still have to think in an optimized way to do it.

- Until now, does not work in real hardware, I used some primitives supported by Bochs/Qemu/VirtualBox to make it simple the support on VBE, as the switch between modes and the fixed physical address.

I believe that even so, the implementation is interesting and can be helpful in some cases.


Theldus:
Considering the direction the Nanvix will take, does not make sense to add VBE support and moreover, the code is not so well written and there are merge conflicts, so, this could be more interesting to a general purpose version of Nanvix in a future.

ppenna:
Perfetcly agree @Theldus

PR #70, opened 2016-10-15
Newlib support

Theldus:
With this pull, Nanvix now has full support to libc and libm through Newlib, provided by Red Hat.

To this, was made a merge with the Nanvix source tree's and the library, so that there is a good connection between the library and the system, moreover, headers and source are provided together with Nanvix, something like happens on BSD systems.

Thus, applications that use standard C library should be able to be compiled and run on Nanvix smoothly.

However, Newlib consists of many pieces of code and is originally made to a variety of systems and architectures, and then, there is a good deal of generic code for such a specific system as Nanvix and that will soon be removed.


ppenna:
Newlibc seems to be well integrated with Nanvix. However, this new library yields to larger binary files. Therefore, the initrd image had to be fine tuned and some user utilities to be removed from the standard system build. I have created an issue to track this bug.

Issue #73, opened 2016-10-15
ISO C Character Types Library <ctype.h>


ppenna:
**Description**

Provide an ISO C standard implementation of the Character Types Library \<ctype.h\>.

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/ctype.h.html

**Functions**

`int isalnum(int)`
`int isalpha(int)`
`int isblank(int)`
`int iscntrl(int)`
`int isdigit(int)`
`int isgraph(int)`
`int islower(int)`
`int isprint(int)`
`int ispunct(int)`
`int isspace(int)`
`int isupper(int)`
`int isxdigit(int)`
`int tolower(int)`
`int toupper(int)`

ppenna:
Under review.

ppenna:
Enhancement added in commit 11c31de21b627f4dbbf3a1960ece9773d809231b.

PR #80, opened 2016-10-19
Aligning ctype.h to ISO C

Theldus:
According to issue #73 was removed some dependent compiler macros and distinguished standard functions from extended functions.


ppenna:
Under review.

PR #144, opened 2017-08-09
Nanvix Educational Improvements

fjorgemota:
This PR changes a few things in Nanvix Educational:

- Adds support for GDB+Bochs (so disabling Bochs debugger);
- Creates ISO image without root;
- Compiles with Debug symbols so debugging is really nice (and without really needing ;
- Improves setup-toolchain.sh so it can run even on environments with low space available;
- Changes Bochs to use Curses UI (changing it's dependencies too);
- Add Dockerfile so we can compile and run Nanvix on Docker; 

Well, that's a bunch of changes..and as far I know it's working well here.

Feel free to review and comment on any changes. =)

Thanks!


fjorgemota:
Just warning that we are still working on some edges (like `strtol`/`atoi`) in this PR, so...please, do not merge it yet.

But the review of the rest of the code would be excellent anyway. =)

tools/dev/setup-toolchain.sh
```diff
@@ -28,13 +28,15 @@ wget "http://ftp.gnu.org/gnu/binutils/binutils-2.25.tar.bz2"
...
+apt-get install -y g++ doxygen genisoimage gdb
 
 # Export variables.
 export PREFIX=/usr/local/cross
 export TARGET=i386-elf
+export CFLAGS=-pipe
+export CXXFLAGS=-pipe
 sh -c "echo 'export TARGET=$TARGET' > /etc/profile.d/var.sh"
-sh -c "echo 'export PATH=$PATH:$PREFIX/bin' >> /etc/profile.d/var.sh"
+sh -c "echo 'export PATH=\$PATH:$PREFIX/bin' >> /etc/profile.d/var.sh"
```
ppenna: @fjorgemota '\' stands for character escaping?

tools/dev/setup-bochs.sh
```diff
@@ -28,16 +28,21 @@ export WORKDIR=$CURDIR/nanvix-toolchain
...
+apt-get install -y libncurses5-dev
 
 # Get bochs.
 wget "http://sourceforge.net/projects/bochs/files/bochs/2.6.8/bochs-2.6.8.tar.gz"
 
 # Build Bochs
 tar -xvf bochs-2.6.8.tar.gz
 cd bochs-2.6.8/
-./configure --enable-x86-debugger --enable-debugger --enable-debugger-gui
+patch < $CURDIR/tools/dev/gdbstub.patch
```
ppenna: @fjorgemota what bug this patch fixes?

fjorgemota: Hey!

This patch fixes a behavior in Bochs's GDB Stub that causes the signal zero to be emitted to GDB each time a page fault occurs with signal zero (which GDB cannot ignore).

Basically, this patch adds a condition to emit SIGSEGV (instead of signal zero) when Bochs stops because of a page fault (and other reasons not really clear, but definitely not related to step by step and/or  breakpoints). As SIGSEGV can be configured to be ignored by GDB,  with `handle SIGSEGV nostop nopass noprint` in GDB, we can finally ignore/block these stops caused by the page fault. =)

tools/dev/setup-toolchain.sh
fjorgemota: I did not understand the question.

This change basically causes `export PATH=$PATH:$PREFIX/bin` to be saved in `/etc/profile.d/var.sh`, instead of resolving `$PATH` based on the environment where `setup-toolchain.sh` ran, fixing issues where the environment defined by `sudo` is different from those where the user really is (which is my case).

An basic example: If someone has [pyenv](https://github.com/pyenv/pyenv) installed, loaded by `.bashrc` **before** using something like `source /etc/profile.d/var.sh` (which is my case), this change allow `$PREFIX/bin` to be *appended* to `PATH`, instead of replacing `PATH` with the `PATH` that `sudo` uses followed by `$PREFIX/bin`.

It's clear now?

PR #155, opened 2017-08-27
Performance counters support

Theldus:
This PR adds support for Intel performance monitoring counters, version 2. Note that it currently supports only 6 events* that are guaranteed to be supported in the Intel Core microarchitecture:
- UnHalted Core Cycles
- UnHalted Reference Cycles
- Instructions Retired
- LLC Reference
- LLC Misses
- Branch Instruction Retired
- Branch Misses Retired

(please refer to chapter 19.1 of the Intel Software Developer's Manual, volume 3, for a detailed events description).

\* New events can be easily added, although makes the code highly machine-dependent.


include/nanvix/config.h
```diff
@@ -28,7 +28,7 @@
 	/**@{*/
 	#define MACHINE_NAME "valhalla" /**< Machine name.              */
 	#define CPU                i386 /**< Cpu model.                 */
-	#define MEMORY_SIZE   0x1000000 /**< Memory size (in bytes).    */
+	#define MEMORY_SIZE   0x4000000 /**< Memory size (in bytes).    */
```
ppenna: @Theldus Should we really change memory size?

Theldus: If I remember correctly, I changed the memory size because the benchmarks I was using used a little over 16MB. You can leave the size as before, there are no problems :)

Theldus:
since everything looks fine, I've merged the PR

PR #158, opened 2017-08-29
Toolchain submodule

Theldus:
This PR uses the "toolchain" repository as a submodule, this way, we can manage better the toolchain used and not have to deal with another servers, like gnu.org for instance.

Besides that, this makes the tools more integrated with the system once we have the 'Nanvix' organization, all the stuff related to Nanvix can be bind here in some way.

The submodule 'i386' points to the master branch in 'toolchain' repository, so we can pull Intel or OpenRISC toolchain only when needed.

--

Even so, there are some things that I would like to point:
* This approach works fine only if the user get the source by cloning from github, not from a zip file.
* Since we are using submodule we can't get rid the source files after installing toolchain, otherwise, git client will complaim that we have local changes.
* The submodule size will always grow at every change we made and the total size will be larger than current toolchain size (i.e: gcc/gdb/binutils + history).

### DIFF
--- a/tools/dev/setup-toolchain.sh
+++ b/tools/dev/setup-toolchain.sh
@@ -43,7 +43,7 @@ sh -c "echo 'export TARGET=$TARGET' > /etc/profile.d/var.sh"
 sh -c "echo 'export PATH=$PATH:$PREFIX/bin' >> /etc/profile.d/var.sh"
 
 # Build binutils.
-cd binutils-2.25/
+cd binutils*/
 ./configure --target=$TARGET --prefix=$PREFIX --disable-nls
 make -j$num_cores all
 make install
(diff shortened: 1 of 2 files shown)


tools/dev/setup-toolchain.sh
```diff
@@ -21,56 +21,58 @@
 
 # Set working directory.
 export CURDIR=`pwd`
-export WORKDIR=$CURDIR/nanvix-toolchain
-mkdir -p $WORKDIR
+export WORKDIR=$CURDIR/toolchain/i386
```
ppenna: @Theldus  shouldn't it be export WORKDIR=$CURDIR/tools/dev/toolchain/i386

Theldus: @ppenna Okay, I'll change. I thought the script would be invoked from within the 'dev' folder.

PR #162, opened 2017-08-31
Travis toolchain image

Theldus:
Travis CI is now working with the toolchain. You just need to get my image from theldus/nanvix:compile, put in the official repository and update .travis.yml to point to there.


turbolay:
Wasn't already made in #154?

ppenna:
@turbolay we have updated the toolchain structure, and thus we had to update the docker image.

Issue #139, opened 2017-07-21
Kernel Error from Userland Inside a fork()


Mesh33:
**Brief**
The follow code causes a kernel panic.
```
static int sem_test(void)
{
	if (fork() == 0)
	{
		/* child */
		int *i;
		i = NULL;
		printf("%d \n",*i);
	}
	else
	{

	}

	return (0);
}
```
**Further Information**

* Accessing the pointer initialized to "NULL" will result in a kernel panic.
* If the code is inside the father, there will only be a memory violation.

ppenna:
Unit test introduced for tracking this error in commit https://github.com/nanvix/nanvix/commits/dev.

Theldus:
I think I understood the problem, it happens that the child process when it tries to access the invalid portion of memory, `do_page_fault()` is triggered with error code 5. Giving a quick look at error codes [here](http://wiki.osdev.org/Exceptions#Page_Fault), and looking at this snippet of code (src/kernel/arch/x86/exception.c):
```c
/* Validty page fault. */
if (!(err & 1))
{
	if (!vfault(addr))
		return;
}
	
/* Protection page fault. */
if (err & 2)
{
	if (!pfault(addr))
		return;
}
	
if (KERNEL_RUNNING(curr_proc))
{
	dumpregs(&s);
	kpanic("kernel page fault %d at %x", err, addr);
}
```
the code 101 (flags UWP - User/Page read/Protection violation) is not treated in either the first `if` nor the second one, and once `intlvl` is > 1, the kernel panic is fired.

I believe that error code 5 could be dealt in the same way as 7, but maybe you can check separately, for instance:
```c
if ( (addr == 0) && (err & 4) )
	goto sigsegv;
...
sigsegv:
	sndsig(curr_proc, SIGSEGV);
	die(((SIGSEGV & 0xff) << 16) | (1 << 9));
```

PR #169, opened 2017-09-05
fix KERNEL_WAS_RUNNING macro

ppenna:
Bug fix for issue https://github.com/nanvix/nanvix/issues/139.


fjorgemota:
Hey, here this PR causes Nanvix to be unable to boot. It simply stops at "kout is not initialized" and stops there. 

I've tried to debug the cause of the problem using GDB, but I only found that `sys_fork` is a frequent item in the backtraces while trying to locate the problem.

May it be related to the intlvl change?

mbcastro:
Any progress in this matter?

PR #181, opened 2018-03-30
Bad intlvl

ppenna:


Theldus:
Looks like the patch works fine in both branches, although in the `dev` seems to spent more time to complete the `test sched` test.

Regarding travis, I have no idea why the error, since the first build passes and the other not. Looking in the logs, seems that while the docker image was pulling, has occurred a timeout after 49 minutes (which is so much time) and the job was terminated. Maybe the Travis-CI experimented connection issues, slow down, or something like that.

PR #183, opened 2018-03-31
enhancing sched 3 test

ppenna:


mbcastro:
This solution slightly differs from the one proposed by @ppenna for the master branch. Please, take a look at: https://github.com/nanvix/nanvix/commit/3c9e651ddf5b23f0295e0540e459b80fb3d01b67

I think that the same solution should be applied to both branches, right?

Theldus:
Actually this PR is the very same solution, but I don't know why doesn't work in the `dev` branch.

mbcastro:
I think that the problem is related to the KERNEL_RUNNING macro

master brach: https://github.com/nanvix/nanvix/pull/183

mbcastro:
sorry! I've closed it :-(

PR #190, opened 2018-04-24
i386, bug fix: preserving registers between context switch

### LINKED ISSUE #184: Some Registers Are Not Preserved Upon Context Switch
**Description**

**x86**

In the x86 kernel, the ``switch_to`` routine does not save the EBX, ESI and EDI registers, even though the toolchain's ABI assumes it does so.

As a consequence, the preempted processes may potentially misbehave, or even the kernel itself may crash unexpectedly,

**or1k**

The same behaviour occurs in the or1k implementation, and the caller save registers must be saved before the context switch.

Theldus:
This PR fixes #184.

### DIFF
--- a/src/kernel/arch/x86/utilities.S
+++ b/src/kernel/arch/x86/utilities.S
@@ -211,6 +211,8 @@ switch_to:
 	movl curr_proc, %eax
 	pushfl
 	pushl %ebx
+	pushl %esi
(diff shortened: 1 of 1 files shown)


Theldus:
The or1k issue will be fixed inside the branch or1k, since it's in working stages yet.

PR #207, opened 2018-05-11
Add gcc eight support.

### LINKED ISSUE #198: Possible bug in unsetenv()?
The function `unsetenv()` in `src/lib/libc/stdlib/unsetenv.c` does the following:

`if ((*P = *(P + 1)) != '\0')`

However, `P` is declared as `register char **P;`. So, maybe the correct way of doing it would be:

`if (*(*P = *(P + 1)) != '\0')`

tarberd:
# What:
Update the compiler used in the toolchain.

# Why:
Good compiler features are missing from gcc 5.8. It can help to write better C. And since the project does not use the libc there should be no major problems on doing the upgrade.

# Fixes:
It changes some misleading indentations.
Fixes #198.

### DIFF
--- a/src/kernel/lib/kpanic.c
+++ b/src/kernel/lib/kpanic.c
@@ -53,5 +53,5 @@ PUBLIC void kpanic(const char *fmt, ...)
 	disable_interrupts();
 	
 	while(1);
-		halt();
+	halt();
 }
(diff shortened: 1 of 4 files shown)


fjorgemota:
One thing I think it's worth observing is that this PR **does not** break the compatibility between devs with old toolchains vs. new toolchain, because there is no new feature used (only valid C).

I will test installing the toolchain here on my machine and return feedback soon.

PR #219, opened 2018-05-21
Fix initrd overflow and bad initrd geometry

gbesnard:


gbesnard:
Oups, thanks for noticing the omission @Theldus.
Now that the binaries are compressed, do you think we shrink the image size again a little or do we keep some room to add some binaries in the future ?

PR #230, opened 2018-06-05
Enable Bochs safe optimizations on master

fjorgemota:
Reading [this page on Bochs documentation](http://bochs.sourceforge.net/doc/docbook/user/compiling.html), I found a flag that, according to Bochs changelog, enable some optimizations with improvements between 10% and 15% on performance.

This flag enables, according to Bochs developers, **only safe optimizations**, so it should not be a problem for nanvix itself.

I executed a few tests here but did not found a big improvement yet, probably the fact that I'm running  three bochs instances executing nanvix on my machine does not help. :smile: 

Please note that with this flag an error is shown about handlers chaining being not compatible with gdb stub. I do not know why this is flagged as an error, because Bochs itself compiles normally and the gdb integration works normally..Probably it should be flagged as a warning, but okay, it's safe to ignore anyway. :stuck_out_tongue:

### DIFF
--- a/tools/dev/setup-bochs.sh
+++ b/tools/dev/setup-bochs.sh
@@ -42,7 +42,7 @@ wget --no-check-certificate "http://sourceforge.net/projects/bochs/files/bochs/2
 tar -xvf bochs-2.6.9.tar.gz
 cd bochs-2.6.9/
 patch < $CURDIR/tools/dev/gdbstub.patch
-./configure --with-term --enable-gdb-stub
+./configure --with-term --enable-gdb-stub --enable-all-optimizations
 sed -i '/^\<LIBS\>/ s/$/ -lpthread/' Makefile
 make all
 make install


fjorgemota:
Hey.

I just wanna inform that I discovered that one of the features enabled by the added flag is dependent on a GCC attribute on each function that we want that feature enabled.

That feature is called x86 fast call, and works by removing the need to use the stack when working with functions with up to 2 integral arguments (by putting these arguments directly on registers..), as far I read. There is [some documentation here](https://gcc.gnu.org/onlinedocs/gcc/x86-Function-Attributes.html), but I will check that when I have some free time, but maybe we can optimize some intensive functions on the process scheduling.

What do you think? I accept opinions here 😄

Theldus:
It's an interesting feature.. but activating fastcall might not be a good idea, since we have assembly functions in the kernel that pass parameters through the stack, which would require rewriting all of them.

fjorgemota:
Initially my idea was marking as fastcall just some functions, mainly used on the scheduling algorithm (in C) because it is a really frequent operation. Even these operation would require to rewrite these assembly functions? I do not know much about the assembly stack of the Nanvix so I think that question is important.

Theldus:
> Initially my idea was marking as fastcall just some functions, mainly used on the scheduling algorithm (in C) because it is a really frequent operation. Even these operation would require to rewrite these assembly functions? I do not know much about the assembly stack of the Nanvix so I think that question is important.

It depends of what functions we are talking about, the `yield()` function doesn't receive any arguments, so I think that does not makes difference, the `switch_to` (asm), receives one argument...

> Anyway, I enabled the fastcall support anyway because it did not changed the Nanvix execution (I think that it really needs support from the compiler, right?), so even in the case that we decide to not use fastcall, as far I know, there is no problem to leave it enabled, right?

The docs do not say clearly why this feature is exclusive to x86 or if fastcall is the same fastcall we're thinking to be... I also would not know how much speedup we would have with using fastcall in a few points of the code.

Anyway, if the system works fine (without changes) with this flag enabled, I see no reasons why we shouldn't use it. Moreover, since we are talking about performance, it may be equally interesting to activate the gcc optimization flags (-O2, -O3 ...) by the variable `CXXFLAGS`, as described here:
http://peter.quantr.hk/2014/12/make-boots-runs-faster/. ;-)

fjorgemota:
Oh. I totally forgot that `switch_to` is a function implemented directly on assembly, sorry for that.

> The docs do not say clearly why this feature is exclusive to x86 or if fastcall is the same fastcall we're thinking to be...

I think it is just a calling convention related to x86 (maybe x86-64 has another option that is better suited for that architecture? I do not know..)

> Anyway, if the system works fine (without changes) with this flag enabled, I see no reasons why we shouldn't use it. Moreover, since we are talking about performance, it may be equally interesting to activate the gcc optimization flags (-O2, -O3 ...) by the variable CXXFLAGS, as described here:
http://peter.quantr.hk/2014/12/make-boots-runs-faster/. ;-)

I'll test that on this weekend! Of course, if the results were successful (and does not cause problems with GDB usage on any way), i'll update this PR. Thanks for the suggestion.

fjorgemota:
By the way, this PR is free to merge, as long someone more test it. =)

Theldus:
Awesome results, let us wait someone with merge powers review this =)

mbcastro:
I think that @ppenna should check that too but it seems that everything is good!

PR #231, opened 2018-06-12
Conforming low-level i386 functions to the threads branch

Theldus:
This PR fixes building issues in the i386 arch and conforms to the threads implementation.


gbesnard:
To me this work perfectly. Do you want to have a look @ppenna  ?
There is a travis error due to recent changes: nothing big, I would correct it as soon as it is merged.

PR #243, opened 2018-07-31
Update setup-toolchain.sh

mbcastro:
The xz-utils and make packages may not be installed in some distros, so it is not possible to decompress binutils and gcc and compile them. This patch adds the packages xz-utils and make to the required packages.


mbcastro:
I've just added the "make" package too.

PR #244, opened 2018-08-01
Priority Scheduler

luigisoares97:
Updated the scheduling algorithm to consider other fields than counter (priority and nice):

- Added a macro to calculate the process' effective priority (priority + nice - counter).

- Added a macro to verify whether or not we should update the next process to be executed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ $ sudo make image TARGET=i386
 If using or1k architecture:
 ```bash
 $ make nanvix TARGET=or1k
-$ sudo make image TARGET=or1k```
+$ sudo make image TARGET=or1k
 ```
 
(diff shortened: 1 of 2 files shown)


ppenna:
@Theldus actually me and @luigisoares97 found a bug on the priority scheduling test. In summary, we are writing to the HDD device which no longer exist in the dev branch. @luigisoares97 is working on the fix for that also.

Theldus:
Oh, I got it
You mean the `work_io()` function, right? which tries to read from /dev/hdd... which should not even exist in the dev branch.

ppenna:
Exactly that! We are working some fixed for that =) may I merge it?

Theldus:
Merged ;-)

PR #248, opened 2018-08-06
Setup arch linux

luigisoares97:
**Changes**

- Modified `<setup-toolchain.sh>` and `<setup-qemu.sh>` to work properly on both debian-based or arch-based systems.
- Updated qemu version.

### DIFF
--- a/tools/dev/setup-qemu.sh
+++ b/tools/dev/setup-qemu.sh
@@ -25,7 +25,7 @@
 #
 
 # QEMU.
-export QEMU_VERSION=3.0.0-rc3
+export QEMU_VERSION=2.12.0
 
 # Set working directory.
 export CURDIR=`pwd`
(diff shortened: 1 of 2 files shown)


ppenna:
@Theldus does it break compatibility for you?

luigisoares97:
@ppenna @Theldus I already downgrade qemu version. For some reason i was having trouble to install it by setup-qemu script but, even without the two missing dependencies that I found looking in arch repository, I tried at least five times now and it magically worked in all of them. So, I think that wait for GDB fixes is the best option for now.

Theldus:
@luigisoares97 I saw here that you remove the Qemu update plus the statements that check for the current distro. The latter you can keep if you want, and makes use of this PR to add conditional statements that check for the current Linux distribution, what do you think?

luigisoares97:
@Theldus Actually, this was a mistake and i'll revert it (thanks, didn't noticed that). Think I should change those ifs to a switch-case for better readability and so anyone could add your distros specifications properly (In my case, I always been using debian-based distros and just moved to an arch-based at this weekend, so i don't know anything about others distros commands).

Theldus:
Nice, so, I'll wait you push more commits and when we're ready to go, ping me and I will approve and merge this commit, ;-).
> so i don't know anything about others distros commands
It's important to keep 'bashisms' simple enough so we can run these scripts from any other distro, would even be better if we try to use only 'sh' stuff, since sh is supported by POSIX and available in more environments than bash. 

Regards the switch-case, I agree with you and I suggest you to add a default case that warns the user that the running distro is not officially supported and/or tested by us.

luigisoares97:
@Theldus Done. Reverted changes that I mistakenly made and added the default case to raise a warning. Think everything is ok now ;)

PR #249, opened 2018-08-07
Sched test 04

luigidcsoares:
**Changes**

- `<test.c>`:
    - `work_io`: modified for loop inside `work_io` function to reduce the high processing time
   - `sched_test4`: added a new unit test that spawns several process in tree-like structure (every process that isn't a leaf has exactly 2 childs) where processes priority grows as deep as they are in the tree.


Theldus:
Works fine here, @ppenna you can merge if you want.

Theldus:
I saw this, I thought it was intentional, @ppenna sorry for not saying anything.

ppenna:
Not a problem @Theldus  =)

PR #252, opened 2018-10-29
Add support for Arch on setup-toolchain and setup-bochs scripts

fjorgemota:
Hey. 

As I'm now an Arch user (well..Antergos, but it's based on Arch anyway)...I wanted to add support to run Nanvix on arch-based distros... and so I ported #248 from the `dev` branch to the `master` branch.

Please note that all the Ubuntu/Debian code is the same, and I adapted the code that runs on Arch to maintain the same order of the packages, so it's easier to maintain.

By the way, this is a feature commonly requested by students because a few students seem to prefer Arch, so this PR fixes that problem too.. and according to my tests everything runs fine even after this PR (well, at least on the Arch side)

Thank you!

### DIFF
--- a/tools/dev/setup-bochs.sh
+++ b/tools/dev/setup-bochs.sh
@@ -32,8 +32,23 @@ cd $WORKDIR
 export CFLAGS="-pipe -O3"
 export CXXFLAGS="-pipe -O3"
 
+
 # Get required packages.
-apt-get install -y libncurses5-dev
+DIST=$(uname -rv)
+
+case ${DIST,,} in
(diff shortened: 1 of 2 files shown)


fjorgemota:
I think most of the differences between this PR and the old one is that this PR installs dependencies for bochs instead of qemu, and there are a few packages that @mbcastro added a few time ago so Nanvix compiled easily on virtual machines that used a slimmed version of Ubuntu.

But really..that's it, in fact it's a very simple PR. I think it only needs a bit more testing to guarantee that everythign is OK.

PR #253, opened 2018-12-10
OpenRISC SMP Support

Theldus:
This PR introduces support to SMP in OpenRISC architecture, as well as threads support (introduced in `kernel-threads-dev` branch and merged here) for both OpenRISC (thanks to @gbesnard) and i386 with slightly changes.

The approach used makes use of master-slave processors. The master processor (usually id 0) receives messages from the slave via Inter Processor Interrupt (IPI) and thus treats all types of privileged accesses that the slave may have. The system always executes one process at a time and distributes the threads between the available cores. Therefore, slave cores basically spend most of their time in user mode, while the master waits for interrupts.

To achieve this, various modifications were required, from the thread structure, specific and individual structures of each core, mode of handling interrupts, race conditions and so on.

So far, the emulator used (Qemu) only supports 2 cores in the OpenRISC architecture, so no tests with a number greater than 2 have been made, although the current implementation is not restricted to a number of cores.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -26,10 +26,25 @@ E: contato@fjorgemota.com
 W: https://fjorgemota.com
 D: Bachelor Student in Computer Science at UFSC
 
+N: Guillaume Besnard
+E: guillaume.besnard@protonmail.com
+W: https://www.linkedin.com/in/guillaume-besnard-416281104
+D: Engineering student in Computer Science at UGA, paragliding and mountain addict
+
+N: João Vicente Souto
+E: joaovicentesouto@gmail.com
+W: https://www.linkedin.com/in/joaovicentesouto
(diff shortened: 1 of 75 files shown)


Theldus:
@ppenna, as you can see, there are merge conflicts. When attempting to resolve them, x86 performance degraded significantly by introducing the `PRIORITY` and` HIGHER_PRIORITY` macros in the process scheduler.

I pushed a draft branch called `or1k-merge-test`, which contains my changes that solves the merge conflict but introduce this behavior. Since you also worked on those macros, could you check the latest changes I made in pm/sched.c?

Theldus:
@ppenna I've synced with dev (by using the merge strategy -Xours) and everything works fine now, just the latest sched.c changes was discarded, as we have talked earlier.

PR #258, opened 2019-08-31
C Library Support for sleep() ad alarm()

alekfrohlich:
This PR is an expansion to nanvix's libc and bosch configuration
  scheme (mainly to bochs run script). Altough it includes a user
  binary (namely, sleep), it merely serves the purpose of ilustrate
  sleep's functionallity and shouldn't be included in default user
  software-package. The following sections describe what has been
  changed by this PR in detail.

  Please note that if you're seeing this then the commit is still
  a WIP (Work In Progress). Please refer to the Pull Request thread
  for up-to-date discussion on changes proposed here.

  * Changes to libc:

      This changes were the inital objective of the PR, however, as
      explained in [bochs configuration], the need for a more versatile
      run-script was also found.

      - alarm.c (new):

          This file represents libc's wrapper to the sys_alarm syscall
          as specified by unistd.h. The wrapper does not contain any
          extra return logic, it only forwards the return of sys_alarm.
          That's the case because kernel/sys/alarm.c already implements
          all of it according to:

          [RETURN VALUE] http://pubs.opengroup.org/onlinepubs/009695399/
                functions/alarm.html

          [lines 42-45] https://github.com/nanvix/nanvix/blob/master/src/
                kernel/sys/alarm.c

      - sleep.c (new):

          This file makes part of libc but it isn't a syscall-wrapper.
          Sleep is implementing by setting up an alarm (the same as
          described above) and then registering a signal_handler with a
          call to signal (using the libc wrapper). In contrast to
          linux's libc, this implementation of sleep doesn't return the
          unslept time in case of another signal (this should probably
          be changed before the merge).

          [RETURN VALUE] https://pubs.opengroup.org/onlinepubs/009695399/
                functions/sleep.html

  * Changes to user binaries:

      - sleep (new):

          Simple user binary,

                usage: sleep [seconds]

          Puts the current process to sleep for [seconds] seconds. Even
          though it was written as driver program, it should be
          following all the usual nanvix conventions (style-guides,
          usage, version, argparsing, etc...).

  * Changes to headers:

      As discussed in nanvix's teaching channel on slack, there was a
      problem regarding kernel namespace pollution. The work-around to
      this problem was to add include guards on unistd.h and pm.h
      wrapping the sleep declaration so there is only one defined at a
      given moment. (pm.h's during kernel compilation and unistd.h's
      during the compilation of libc).

      - nanvix/pm.h (modified):

          #ifdef __NANVIX_KERNEL__

          #define sleep (kernel's)

          #endif

      - unistd.h (modified):

          #ifndef __NANVIX_KERNEL__

          #define sleep (libc's)

          #endif

  * Changes to makefiles:

      As explained in [Changes to headers], the work-around for the
      kernel namespace pollution problem (sleep being defined both
      inside pm.h and unistd.h) was to include guard sleep's declaration
      in a way to make the includes mutually exclusive. Moreover, the
      makefile under ubin has been changed to account for the sleep user
      binary.

      - kernel/makefile:

        Now the kernel needs to be compiled with an extra define:

            CFLAGS += -D __NANVIX_KERNEL__

      - ubin/makefile:

        Add sleep rule, target in clean and all, marked as PHONY.

  * Changes to bochs configuration:

      As speculated and now confirmed, bochs's timer isn't real time,
      resulting in grotesque time discrepancies such as running (user
      binary):

          sleep 200 (seconds)

      And it taking only ~6 seconds to complete:

      [BOCHS CLOCK TIKING SUPER FAST] https://stackoverflow.com/questions/
                45845736/emulated-environments-hardware-clock-ticking-a-lot-
                faster-than-18-2-times-per

      [BOCHS TIMER IS NOT REAL TIME] https://wiki.osdev.org/Bochs

      After some web-browsing and some tweaking of bochs parameters,
      it's now capable of running in a state somewhat close to real time
      so that nanvix users do not notice the difference.

      To facilitate further bochs management, the run script was changed
      a bit.

      - run/run.sh:

        The run script now iterates over all script parameters, exitting
        in case of a help/version/unknown flag. Otherwise, the script
        builds a set of config values, DEBUG and RT (Real Time) as of
        now to be latter used by the update_boch function in the seddin'
        of bochsrc.txt.tpl. The set of flags could easily be expanded
        but the way in which they are evaluated (sed calls) should
        probably be modified with code reusability in mind.

      - run/bochs.txt.tpl:

        Added a #RT_ENABLED# marker to be replaced by sed on the line
        that configures clock.

Author: Alek Frohlich, Date: 31/08/2019.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ It relies on the Homebrew package manager (https://brew.sh), to get
 the necessary UNIX-like packages to build Nanvix.
 
 The general flow for building Nanvix on macOS from the root directory
-is as follows :
+is as follows:
 
 ```sh
 bash tools/dev/setup-toolchain.sh
@@ -57,13 +57,8 @@ make image
 
 Unlike the Linux-like systems procedure, this flow does not build
 bochs from source. Instead, it installs the bochs package using brew.
-As this package does not comme with gdb-stubs support, one last step
-will be to **remove** the following line from the
-`tools/run/bochsrc.txt.pl` file :
-
-```
-gdbstub: enabled=#GDBSTUB_ENABLED#, port=1234
-```
+As this package does not come with gdb-stubs support, you cannot run
+Nanvix with `--debug` option (see details below).
 
 ## Running
 
@@ -72,6 +67,9 @@ To run Nanvix, type the following command at the root directory:
 ```sh
 bash tools/run/run.sh
 ```
+
+The script above accepts some optional parameters to configure bochs.
+Please, run `bash tools/run/run.sh --help` for more details.
 
 ## License and Maintainers
 
--- a/include/nanvix/pm.h
+++ b/include/nanvix/pm.h
@@ -213,8 +213,14 @@
 	EXTERN int issig(void);
 	EXTERN void pm_init(void);
 	EXTERN void sched(struct process *);
+
+#ifdef __NANVIX_KERNEL__
+
 	EXTERN void sleep(struct process **, int);
-	EXTERN void sndsig(struct process *, int);
+	
+#endif /* __NANVIX_KERNEL__ */
+
+    EXTERN void sndsig(struct process *, int);
 	EXTERN void wakeup(struct process **);
 	EXTERN void yield(void);
 	
--- a/include/unistd.h
+++ b/include/unistd.h
@@ -190,7 +190,16 @@
 	extern int setuid(pid_t uid);
 	
 	extern pid_t setpgrp(void);
+
+#ifndef __NANVIX_KERNEL__
+
+    /*
+     * Puts the current process to sleep.
+     */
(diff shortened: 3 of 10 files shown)


ppenna:
@mbcastro before I proceed with the merge, could you double check that it builds fine for you too?

mbcastro:
> @mbcastro before I proceed with the merge, could you double check that it builds fine for you too?

The sed option '-i' was not working properly in MacOS. I've made some changes locally so it now works fine. @alekfrohlich: can you give me permissions to push into your forked repo? https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

mbcastro:
@ppenna : could you check if my workaround works on Linux?

PR #260, opened 2019-09-17
Add Check Before Switching to New Process

RamAddict:
# Description

This change prevents any unneeded context switch by checking if the current process is equal to next.

### DIFF
--- a/src/kernel/pm/sched.c
+++ b/src/kernel/pm/sched.c
@@ -116,5 +116,6 @@ PUBLIC void yield(void)
 	next->priority = PRIO_USER;
 	next->state = PROC_RUNNING;
 	next->counter = PROC_QUANTUM;
-	switch_to(next);
+	if (curr_proc != next)
+		switch_to(next);
 }


ppenna:
@mbcastro it works fine in Linux, does it in MacOS? I would say that this improvement is OK....

mbcastro:
Considering the default scheduler (round-robin), I think that the `current` process will always be different to the `next` one. This is due to the fact that there will be, at least, 2 processes in the system (`idle` and `init`). Thus, I believe that this pull request wouldn't take any effect.

ppenna:
@mbcastro I agree with you. However, this is a potential enhancement when Nanvix features another scheduler. What do you think? Should we merge it?

mbcastro:
I've just checked this situation with @RamAddict and we found out that the same process can be rescheduled several times in a row. For instance, the `IDLE` process can be rescheduled several times when all other processes are not in `PROC_READY` state. In this case, we don't need to call `switch_to(next)`. So, this pull request makes sense and I'm accepting it. thanks @RamAddict!

PR #261, opened 2019-10-08
fix(ubin): rename wrong "chmod" from "chown" utility's usage message

StanleySathler:
While trying to understand a bit more about how the `chown` utility works, I realized that it was using the `chmod` word in its usage message rather than the `chown` one - probably due a classic copy n' paste. :yum: 

My commit also removes a lot of empty spaces, but that happened due my personal editor configs. I left the removals in case you guys find it useful, but I can revert them if needed to keep the commit as straightforward / clean as possible.

And **thank you very much** for this awesome project. It has been helping me **a lot** to understand how operating systems work, alongside some books.

### DIFF
--- a/src/ubin/chown/chown.c
+++ b/src/ubin/chown/chown.c
@@ -1,16 +1,16 @@
 /*
  * Copyright(C) 2011-2014 Pedro H. Penna <pedrohenriquepenna@gmail.com>
- * 
+ *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation; either version 3 of the License, or
  * (at your option) any later version.
- * 
+ *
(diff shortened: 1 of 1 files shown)


src/ubin/chown/chown.c
```diff
@@ -42,24 +42,24 @@ static void version(void)
...
+	exit(EXIT_SUCCESS);
 }
 
 /*
  * Prints program usage and exits.
  */
 static void usage(void)
 {
-	printf("Usage: chmod [options] <owner> <file>\n\n");
+	printf("Usage: chown [options] <owner> <file>\n\n");
```
StanleySathler: This is what was really changed at all.

ppenna:
@StanleySathler thanks for the compliments and for the PR I shall review it soon!

PR #267, opened 2023-02-28
Add bash shebang to run setup scripts with others shells

GuilhermeWerner:
Some shells like zsh do not support the `${DIST,,}` directive used in the setup script. By adding the shebang at the top of the scripts, it specifies that they will only be executed by bash.

Instead of running `sudo bash tools/dev/setup-toolchain.sh`, we can simply use `sudo ./tools/dev/setup-toolchain.sh`


ppenna (approved):

ppenna:
Thanks for the contribution @GuilhermeWerner .

PR #271, opened 2024-01-11
Nanvix Dockerized Support

viniciusfdasilva:
* Dockerfile to create a container with Nanvix
* Dockerfile to create an Ubuntu container that runs Nanvix in the QEMU emulator upon starting a bash session! 
* The goal is to facilitate installation and avoid cluttering the host machine with installations.
* Tested implementation.


ppenna (approved):
Great, thanks!

PR #273, opened 2024-07-30
[kernel] E: Alloc usable frames conservatively

rafael0121:
## Description
This PR implements new functions to manipulate frame allocation and allow allocates usable frames conservatively, i.e., just memory marked as usable will be, all the rest, including untracked memorys are reserved.

### DIFF
--- a/kernel/src/mm/mod.rs
+++ b/kernel/src/mm/mod.rs
@@ -87,18 +87,26 @@ fn parse_memory_regions(
     while let Some(region) = memory_regions.pop_front() {
         if region.typ() == MemoryRegionType::Reserved || region.typ() == MemoryRegionType::Mmio {
             if PhysicalAddress::from_virtual_address(region.start()).is_ok() {
-                if region.typ() != MemoryRegionType::Usable {
-                    if let Ok(region) =
-                        TruncatedMemoryRegion::from_virtual_memory_region(region.clone())
-                    {
-                        physical_memory_regions.push_back(region);
-                    }
+                if let Ok(region) =
+                    TruncatedMemoryRegion::from_virtual_memory_region(region.clone())
+                {
+                    physical_memory_regions.push_back(region);
                 }
                 virtual_memory_regions
                     .push_back(TruncatedMemoryRegion::from_memory_region(region)?);
             } else {
                 other_virtual_memory_regions
                     .push_back(TruncatedMemoryRegion::from_memory_region(region)?);
+            }
+        } else {
+            if region.typ() == MemoryRegionType::Usable {
+                if PhysicalAddress::from_virtual_address(region.start()).is_ok() {
+                    if let Ok(region) =
+                        TruncatedMemoryRegion::from_virtual_memory_region(region.clone())
+                    {
+                        physical_memory_regions.push_back(region);
+                    }
+                }
             }
         }
     }
--- a/kernel/src/mm/phys/frame.rs
+++ b/kernel/src/mm/phys/frame.rs
@@ -118,6 +118,38 @@ impl FrameAllocator {
     ///
     /// # Description
     ///
+    /// Free all frames in the range `[start, end]`.
+    ///
+    /// # Parameters
+    ///
+    /// - `start`: Start page frame address.
+    /// - `end`: End page frame address.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, `Ok(())` is returned. Upon failure, an error is returned instead.
+    ///
+    pub fn clear_range(
+        &mut self,
+        region: &TruncatedMemoryRegion<PhysicalAddress>,
+    ) -> Result<(), Error> {
+        let start_frame_number: usize = region.start().into_frame_number().into_raw_value();
+        let end_frame_number: usize = start_frame_number + region.size() / mem::FRAME_SIZE - 1;
+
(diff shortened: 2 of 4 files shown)


kernel/src/mm/phys/frame.rs
```diff
@@ -176,4 +208,49 @@ impl FrameAllocator {
...
+        for index in start_frame_number..=end_frame_number {
+            match self.bitmap.set(index) {
+                Ok(()) => continue,
+                Err(e) if e.code == ErrorCode::ResourceBusy => continue,
+                Err(e) => {
+                        warn!("fail to try book frame {}", index);
+                        return Err(e);
+                }
+            }
+        }
```
ppenna, line 243: If we fail in the middle and return `Err()` we would leave the system in an inconsistent state (some bits were set).

To fix this we should either:
1) Rollback what was allocated.
2) Before allocating, checking if resources are available.

kernel/src/mm/phys/frame.rs
```diff
@@ -176,4 +208,49 @@ impl FrameAllocator {
...
+
+        Ok(())
+    }
+
+    ///
+    /// # Description
+    ///
+    /// Allocs all frames at once, without busy checks.
```
ppenna, line 251: There is a typo here. You may want to write "Allocates"

kernel/src/mm/phys/frame.rs
```diff
@@ -176,4 +208,49 @@ impl FrameAllocator {
...
+    ///
+    pub fn fill_storage(&mut self) -> Result<(), Error> {
```
ppenna, line 253: Could you add comments regarding the return value of this function?

kernel/src/mm/phys/mod.rs
```diff
@@ -64,6 +65,37 @@ static mut FRAME_ALLOCATOR_STORAGE: [u8; config::kernel::MEMORY_SIZE
 // Standalone Functions
 //==================================================================================================
 
+fn parse_physical_memory_regions(
```
ppenna, line 68: Could you add documentation to this function? to this function?

kernel/src/mm/phys/mod.rs
```diff
@@ -102,6 +134,8 @@ fn book_mmio_regions(
                 Ok(()) => {},
                 // Frame lies outside addressable physical memory.
                 Err(e) if e.code == ErrorCode::InvalidArgument => {},
+                // Frame already allocated, continue try booking.
+                Err(e) if e.code == ErrorCode::ResourceBusy => {},
```
ppenna, line 138: If this happens is actually an error. It means that another region has booked it and we are just trying to supress the error.

kernel/src/mm/phys/mod.rs
```diff
@@ -64,6 +65,37 @@ static mut FRAME_ALLOCATOR_STORAGE: [u8; config::kernel::MEMORY_SIZE
...
+    for region in physical_memory_regions {
+        match region.typ() {
+            MemoryRegionType::Usable => {
+                usable_physical_memory_regions.push_back(region);
+            },
+            MemoryRegionType::Reserved => {
+                reserved_physical_memory_regions.push_back(region);
+            }
+            _ => {
+                return Err(Error::new(ErrorCode::InvalidArgument, "Invalid memory type"));
```
ppenna, line 91: Could you please exapnd this to follow a pattern where we log the error message? Something like:

```rust
let reason: &str = "invalid memory type"
error!("parse_physical_memory_regions(): {:?}", reason)
return Err(Error::new(ErrorCode::InvalidArgument, &reason));
```

kernel/src/mm/phys/mod.rs
```diff
@@ -73,7 +105,7 @@ fn book_physical_memory_regions(
     // Book physical memory that is not usable.
     for region in physical_memory_regions.iter() {
         info!("booking: {:?}", region);
-        frame_allocator.alloc_range(region)?;
+        frame_allocator.try_alloc_range(region)?;
```
ppenna, line 108: Why should we try to alloc isntead of allocating it?

kernel/src/mm/mod.rs
```diff
@@ -87,19 +87,27 @@ fn parse_memory_regions(
...
                 other_virtual_memory_regions
                     .push_back(TruncatedMemoryRegion::from_memory_region(region)?);
             }
+        } else {
+            if region.typ() == MemoryRegionType::Usable {
+                if PhysicalAddress::from_virtual_address(region.start()).is_ok() {
+                    if let Ok(region) =
+                        TruncatedMemoryRegion::from_virtual_memory_region(region.clone())
+                    {
+                        physical_memory_regions.push_back(region);
```
ppenna, line 107: We should put these in a different list, to reduce potential bugs in later parsing functions.

kernel/src/mm/phys/frame.rs
rafael0121: Yes.

kernel/src/mm/phys/mod.rs
rafael0121: Sure.

kernel/src/mm/phys/mod.rs
rafael0121: Done.

PR #309, opened 2024-10-18
[doc] E: Add build and run process

### LINKED ISSUE #308: [doc] Update setup, build and run docs
## Description
We need to update the Nanvix docs, updating the setup and adding microvm build and run.

rafael0121:
## Description

Update setup and add instructions to run Nanvix with Microvm and QEMU.
This closes #308

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -1,46 +1,72 @@
-# Development Environment Setup
+# Setting up your Development Environment
 
-This document instructs you on how to setup your development environment.
+This document instructs you on how to setup your development environment to build and run Nanvix.
 
-> ⚠️ Follow these instructions on a fresh Ubuntu 22.04 system.
+> ⚠️ Some instructions instructions in this document assume that you have superuser privileges on the system.
 
-> ⚠️ These instructions assume that you have superuser privileges on the machine.
+## Table of Contents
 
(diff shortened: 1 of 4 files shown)


doc/setup.md
```diff
@@ -1,46 +1,122 @@
...
+
+This section instructs you on how to compile and run Nanvix. 
+
+## Using Microvm
+
+Compile Microvm:
+
+```bash
+cd $WORKDIR/nanvix/src/microvm
+make all
```
ppenna: This should be supported from root dir, if not, we need to bump the pointer for the build repository to the latest one.

rafael0121: Yes, it is. The Kernel too, but i chose this way to show that is possible.

ppenna: We should prefer building from the root directory, as build artifacts will be placed on the same output directory (ie `bin/`) as other build artifacts (ie, `kernel`).

If one builds components individually, build artifacts will be scattered across the repository. As a result, one could even end up with two binaries for `microm` (one in `bin/` and another in` src/microvm/bin`).

Could you please update instructions to guide people to *aways* build from the root directory?

rafael0121: Sure. I got it.

ppenna (approved):

PR #328, opened 2024-11-05
[doc] E: Fix default parameters on make run. Add suggested TIMEOUT.

LucasGdosR:
```bash
# Run Nanvix in QEMU with default parameters:
# ↓ Corrected this line  to agree with Makefile's default arguments ↓
# make TARGET=x86 MACHINE=qemu-pc VERBOSE=no RELEASE=no TIMEOUT=10 FEATURES= run
```

```bash
# ↓ Added this code that's ready to run on low-end machines ↓
make TIMEOUT=90 run
```

### DIFF
--- a/doc/run.md
+++ b/doc/run.md
@@ -17,9 +17,6 @@ This document instructs you on how to run Nanvix.
 # Run Nanvix in QEMU with default parameters:
 # make TARGET=x86 MACHINE=qemu-pc VERBOSE=no RELEASE=no TIMEOUT=10 FEATURES= run
 make run
-
-# TIMEOUT=<seconds>: Set the timeout for the run script. You may want to override the default value if your development environment is a low-end machine (ie: old CPU, HD disk, low RAM).
-make TIMEOUT=90 run
 ```
 
 ### List of Optional Run Parameters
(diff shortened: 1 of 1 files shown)


doc/run.md
```diff
@@ -15,8 +15,11 @@ This document instructs you on how to run Nanvix.
 
 ```bash
 # Run Nanvix in QEMU with default parameters:
-# make TARGET=x86 LOG_LEVEL=warn MACHINE=qemu-pc TIMEOUT=90 run
+# make TARGET=x86 MACHINE=qemu-pc VERBOSE=no RELEASE=no TIMEOUT=10 FEATURES= run
 make run
+
+# TIMEOUT=<seconds>: Set the timeout for the run script. You may want to override the default value if your development environment is a low-end machine (ie: old CPU, HD disk, low RAM).
```
ppenna: @LucasGdosR could you move this description to the section "List of Optional Run Parameters"?

Let's keep the section with default parameters as simple as possible, so we make people who are looking for copy-and-paste happy :)

LucasGdosR: Moved to the correct section and split the copy-paste code into two so people can copy just the bit they need.

ppenna: Looks good, thank you!

PR #434, opened 2025-01-16
[libs] E: creates config crate

### LINKED ISSUE #407: [config] Config Crate
## Context

The configuration options, such as the memory size of the underlying platform on which Nanvix runs, are currently spread across multiple crates. This distribution makes it difficult and cumbersome to manage and modify these settings. Here are some examples of where these configurations are located:

- https://github.com/nanvix/nanvix/blob/95705a10c97762988d9908005e05738636d19317/src/libs/sys/src/sys/config.rs#L16
- https://github.com/nanvix/nanvix/blob/95705a10c97762988d9908005e05738636d19317/src/microvm/src/config.rs#L14
- https://github.com/nanvix/nanvix/blob/95705a10c97762988d9908005e05738636d19317/build/scripts/run.sh#L59

## Proposed Solution

To streamline the management of these configuration options, we propose creating a new `libs/config crate`. This crate will centralize all common configuration options, making them easier to maintain and update. All other crates that require these configuration options should import them from `libs/config`.

Additionally, to facilitate easier sharing of these configurations with non-Rust code (such as shell scripts), we could introduce a `JSON` file. The `libs/config/build.rs` script would read this `JSON` file before building the crate, ensuring that the configurations are consistently applied across different parts of the project.

danbugs:
closes #407


src/kernel/src/mm/kheap.rs
```diff
@@ -14,12 +14,12 @@ use ::alloc::alloc::{
 use ::core::ptr;
 use ::sys::{
     arch::mem,
-    constants,
     error::{
         Error,
         ErrorCode,
     },
 };
+use config::constants;
```
ppenna: Could you please prefix it with `::` to keep consistency?

src/kernel/src/mm/phys/frame.rs
```diff
@@ -22,12 +22,12 @@ use ::sys::{
         self,
         paging::FrameNumber,
     },
-    constants,
     error::{
         Error,
         ErrorCode,
     },
 };
+use config::constants;
```
ppenna: Could you please prefix it with `::` to keep consistency?

ppenna (commented):
So far, so good!

src/kernel/src/mm/kheap.rs
danbugs: Done in [f588481](https://github.com/nanvix/nanvix/pull/434/commits/f58848116710b2cf3e7335e42815fc2cea5c5f1a) ✅

src/kernel/src/mm/phys/frame.rs
danbugs: Done in [f588481](https://github.com/nanvix/nanvix/pull/434/commits/f58848116710b2cf3e7335e42815fc2cea5c5f1a) ✅

src/libs/config/build.rs
```diff
@@ -0,0 +1,48 @@
+use std::env;
+use std::fs;
+use std::path::Path;
+
+fn main() {
+    // Read the TOML file
+    let kernel_config_content = fs::read_to_string("../../../config/kernel_config.toml").expect("Failed to read kernel_config.toml");
```
ppenna: We can create a symlink to build directory and avoid this hardcode.

src/libs/config/build.rs
```diff
@@ -0,0 +1,48 @@
...
+    if let Some(ipc_message_size) = kernel_config.get("ipc_message_size").and_then(|v| v.as_integer()) {
+        constants.push_str(&format!("pub const IPC_MESSAGE_SIZE: usize = {};\n", ipc_message_size));
+    }
+    constants.push_str("}\n");
+
+    // Write the generated file
+    fs::write(&dest_path, constants).expect("Failed to write kernel_config.rs");
+
+    // Inform Cargo to rerun the build script if the TOML changes
+    println!("cargo:rerun-if-changed=kernel_config.toml");
```
ppenna: No need to provide full/reative path?

src/libs/config/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// I/O port that is connected to the standard output of the virtual machine.
+    pub const DEFAULT_STDOUT_PORT: u16 = 0xe9;
+
+    /// I/O port that is connected to the standard input of the virtual machine.
+    pub const DEFAULT_STDIN_PORT: u16 = 0xea;
+
+    /// I/O port that enables the guest to invoke functionalities of the virtual machine monitor.
+    pub const DEFAULT_VMM_PORT: u16 = 0x604;
+}
```
ppenna, line 73: I would not move these here, as they are not config but constants (which are platform dependent)

src/libs/config/build.rs
danbugs: Done in [570a85c](https://github.com/nanvix/nanvix/pull/434/commits/570a85cd10c6775a7d73c26cda2546aa8a871414) ✅

src/libs/config/build.rs
danbugs: Fixed in [570a85c](https://github.com/nanvix/nanvix/pull/434/commits/570a85cd10c6775a7d73c26cda2546aa8a871414) ✅

src/libs/config/src/lib.rs
danbugs: I just noticed a lot of common usages for these constants hardcoded around the codebase. Figured it might be best to place them in one spot. microvm and Hyperlight HALs share most of these constants too.

PR #441, opened 2025-01-18
[libs] E: creates config crate

### LINKED ISSUE #407: [config] Config Crate

danbugs:
closes #407


src/kernel/src/hal/platform/pc/qemu/pc.rs
```diff
@@ -16,7 +16,7 @@
 ///
 pub fn shutdown() -> ! {
     unsafe {
-        ::sys::arch::io::out16(0x604, 0x2000);
+        ::sys::arch::io::out16(::config::hal::DEFAULT_VMM_PORT, 0x2000);
```
ppenna: `0x20000` is also a system-wide constant.

src/libs/config/config-files
ppenna, line 1: This symlink is wrong, and it is likely the reason why linting is failing. It should be:

`../../../config-files/`

To be consistent with the other crates, I would rather move `config-files` to `build/` and then create a symlink to `build`:

`../../../build/`

danbugs: Done in [5ff89e5](https://github.com/nanvix/nanvix/pull/441/commits/5ff89e5ea0cf344a6dabd2edd87e8b9c47b44fdd) ✅

src/kernel/src/hal/platform/pc/qemu/pc.rs
danbugs: Added constant in [5ff89e5](https://github.com/nanvix/nanvix/pull/441/commits/5ff89e5ea0cf344a6dabd2edd87e8b9c47b44fdd) ✅

src/libs/config/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    /// I/O port that is connected to the standard input of the virtual machine.
+    pub const DEFAULT_STDIN_PORT: u16 = 0xea;
+
+    /// I/O port that enables the guest to invoke functionalities of the virtual machine monitor.
+    pub const DEFAULT_VMM_PORT: u16 = 0x604;
+
+    /// Default VMM shutdown command
+    pub const DEFAULT_VMM_SHUTDOWN_CMD: u16 = 0x2000;
+}
```
ppenna: We should gate these constants behind the target platform feature (ie:  `microvm` and `hyperlight`). We should also change `Cargo.toml` of `kernel` and `microvm` to enable it accordingly. For instance, note that not all these constants are relevant to all platforms. We don't want to pollute the namespace.

danbugs: Done in [fa562bb](https://github.com/nanvix/nanvix/pull/441/commits/fa562bb4696cc3181aad5fb0673c6025621439cd) ✅

ppenna (approved):
LGTM

PR #448, opened 2025-02-10
[tests] F: added test_mmap_munmap_return_zeros to mm

### LINKED ISSUE #445: [test] System-Level Test for `mmap()`
## Context

The `mmap()` system call should always return pages filled with zeros, for security reasons.

We should have a system-level test for this.

## Proposed Solution

Add test to the Test Daemon (testd) that:

1. Maps a page with `mmap()`
2. Fills that page with ones
3. Unmaps the page with `unmap()`
4. Re-map same page with `mmap()`
5. Check if page is filled with zeroes

sameep-git:
Closes: #445 

1. Adding a test which maps to a page, fills it up with all 1s, and unmaps. Then, we map back to the same page and read the contents which should be all 0s.

### DIFF
--- a/src/tests/testd/src/mm/mod.rs
+++ b/src/tests/testd/src/mm/mod.rs
@@ -227,6 +227,78 @@ fn test_mmap_munmap_many_times_rolling() -> bool {
     true
 }
 
+///
+/// # Description
+/// 
+/// Attemps to test that mmap always returns pages filled with zeros
+/// 
+/// # Returns
+/// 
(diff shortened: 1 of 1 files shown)


ppenna (commented):
Thank you for submitting this PR.

Could you please modify the test to fill `PAGE_SIZE` bytes with ones, and then check if all those `PAGE_SIZE` bytes are zero?

ppenna (approved):

ppenna:
Merged 7f2bc7bb325591922f8f092b4d5601809ab900ac.

PR #452, opened 2025-02-18
 [benchmarks] E: Add benchmark routine to CI 

### LINKED ISSUE #451: [benchmarks] Create a benchmark routine for microvm in CI
## Context

It would be useful to implement a benchmark routine in CI for performance comparison and analysis during development and before a pull request.

## Proposed Solution

Create a new routine that will run after 'Test (microVM)'. This routine must run Nanvix kernel in microVM multiple times to collect performance data and make this data available in a chart.

rafael0121:
Closes #451.

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -58,8 +58,8 @@ function parse_boottime {
         do
             file=${DATA_PATH}/originalcsv/nanvix-boot_time-${memory}.csv
             paste -d ',' \
-                <(cat $file | grep '^\+,vm_run' | cut -d',' -f 5) \
-                <(cat $file | grep "^\+,vmm_creation" | cut -d',' -f 5) \
+                <(cat $file | grep '^+,vm_run' | cut -d',' -f 5) \
+                <(cat $file | grep '^+,vmm_creation' | cut -d',' -f 5) \
                 | awk -F',' '{print $1 + $2}' \
                 | sed "s/^/${memory},/g" \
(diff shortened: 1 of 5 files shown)


scripts/benchmark.sh
```diff
@@ -0,0 +1,95 @@
...
+
+for MEMORY_SIZE in 64 128 256 512 1024;
+do
+    compile_nanvix $MEMORY_SIZE
+    for IT in $( eval echo {1..$MANY_TIME} );
+    do
+        run_nanvix
+        print_progress $IT $MEMORY_SIZE
+    done
+done
```
ppenna: After running this loop, we need to reset the file to its original state

ppenna:
Instead of generating one plot per VM size, could we have a single plot with all results?

scripts/benchmark.sh
rafael0121: Okay, but what about the 'csvdata' folder? 
Maybe we should add a new CI step here to clear the entire environment after uploading artifacts. 
What do you think?

https://github.com/nanvix/nanvix/blob/1eaf7c13acc435fa1840349d0c2ac6cd25f1bb49/.github/workflows/x86-release.yml#L208-L216

ppenna: Yes, we should upload artifacts.

rafael0121:
> Instead of generating one plot per VM size, could we have a single plot with all results?

Yes, it looks right.

PR #537, opened 2025-05-10
[enhancement] Update hyperlight dependency across microvm and kernel crates

danbugs:
`nanvix/hyperlight` was almost 300 commits behind `hyperlight-dev/hyperlight`. This PR updates the Hyperlight dependency for Nanvix.

See individual commits for more info.

To use the new version of  `hyperlight-host/common` locally, after this merges, make sure to run `cargo update` at the root of the repo.

### DIFF
--- a/scripts/pipeline.sh
+++ b/scripts/pipeline.sh
@@ -13,7 +13,7 @@ TOOLCHAIN_DIR=${1:-$(pwd)/toolchain}
 TTY=${2:-/dev/null}
 
 # Timeout for each step.
-TIMEOUT=90
+TIMEOUT=300
 
 # Padding to force aligned formatting.
 PADDING=40
--- a/src/kernel/src/hal/platform/hyperlight/peb/mod.rs
+++ b/src/kernel/src/hal/platform/hyperlight/peb/mod.rs
@@ -52,27 +52,7 @@ pub enum OutBAction {
     _Log = 99,
     CallFunction = 101,
     _Abort = 102,
-    Magic = 103,
-}
-
-#[derive(Debug, Copy, Clone)]
-#[repr(C, packed)]
-pub struct HostFunctionDefinitions {
-    pub fbHostFunctionDetailsSize: u64,
-    pub fbHostFunctionDetails: u64,
-}
-
-#[derive(Debug, Copy, Clone)]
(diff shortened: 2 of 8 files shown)


src/kernel/src/hal/platform/hyperlight/peb/mod.rs
```diff
@@ -196,15 +165,27 @@ impl ProcessEnvironmentBlock {
...
+    fn debug_print(&mut self, message: &str) -> Result<(), Error> {
+        let port = OutBAction::DebugPrint as u16;
+        let data = message.as_bytes();
+
+        for chunk in data.chunks(4) {
+            let val = match chunk {
+                [a, b, c, d] => u32::from_le_bytes([*a, *b, *c, *d]),
+                [a, b, c] => u32::from_le_bytes([*a, *b, *c, 0x7F]),
+                [a, b] => u32::from_le_bytes([*a, *b, 0x7F, 0x7F]),
+                [a] => u32::from_le_bytes([*a, 0x7F, 0x7F, 0x7F]),
```
ppenna: Why `0x7f` and not `0xff`?

ppenna (approved):
LGTM

src/kernel/src/hal/platform/hyperlight/peb/mod.rs
danbugs: 0x7F padding is not printable in HL. 0xFF will output �

ppenna: Could you add a comment about this? This design decision is not obvious. For instance, I assumed that Hyperlight would scan the u32 for characters that are out of the printable ASCII range to handle the conversion manually.

scripts/pipeline.sh
```diff
@@ -13,7 +13,7 @@ TOOLCHAIN_DIR=${1:-$(pwd)/toolchain}
 TTY=${2:-/dev/null}
 
 # Timeout for each step.
-TIMEOUT=90
+TIMEOUT=300
```
ppenna, line 16: We should change in ci.py and Makefile as well.

danbugs: Done!

Issue #534, opened 2025-05-07
[kernel] Enable 64-bit Return Values for Kernel Calls


ppenna:
## Context

We should change the kernel call interface to enable 64-bit return values.

## Proposed Solution

Change the kernel call interface to use the `eax` and `edx` registers gor returning 64-bit values.

## References

- https://github.com/nanvix/nanvix/blob/8a8fb81c7c878111588858f82fdacb10332c442d/src/kernel/src/hal/arch/x86/hooks.S#L378-L414
- https://github.com/nanvix/nanvix/blob/8a8fb81c7c878111588858f82fdacb10332c442d/src/libs/sys/src/sys/kcall/arch/x86.rs

PR #539, opened 2025-05-15
[kernel] E: supporting 64bit kernel cals

ShayanDeSh:
#534 Adding support for 64bit kcals.

### DIFF
--- a/src/kernel/src/kcall/kcall_success.rs
+++ b/src/kernel/src/kcall/kcall_success.rs
@@ -44,3 +44,9 @@ impl From<KcallSuccess> for i32 {
         result.0
     }
 }
+
+impl From<KcallSuccess> for i64 {
+    fn from(result: KcallSuccess) -> Self {
+        result.0 as i64
+    }
+}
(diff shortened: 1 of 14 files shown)


src/libs/sys/src/sys/pm/pid.rs
ppenna, line 1: Could you add an empty line to the end of this file to conform with the other files in the source tree?

src/kernel/src/kcall/kcall_success.rs
ppenna, line 1: Could you add an empty line to the end of this file to conform with the other files in the source tree?

src/libs/error/src/lib.rs
```diff
@@ -24,7 +24,7 @@ use ::num_enum::TryFromPrimitive;
 ///
 #[derive(Copy, Clone, PartialEq, Eq, Debug, TryFromPrimitive)]
 #[num_enum(error_type(name = Error, constructor = invalid_error_code))]
-#[repr(i32)]
```
ppenna: We still need to have this as a `i32` to conform with a 32-bit ABI for error values. Could you revert this change and add a conversion from `TryFrom<i64>` instead?

PR #540, opened 2025-05-21
Fix a couple of nits in the documentation

csegarragonz:


scripts/pipeline.sh
```diff
@@ -39,7 +39,7 @@ do
             msg="$mode $step $target"
             printf "%-${PADDING}s" "$msg"
             start_time=$(date +%s%3N)
-            python3 scripts/ci.py --${mode} --toolchain-dir=$TOOLCHAIN_DIR --target-arch=x86 --log-level=trace --target-machine=$target --${step} --timeout $TIMEOUT > $TTY
+            python3 scripts/ci.py --${mode} --toolchain-dir=$TOOLCHAIN_DIR --target-arch=x86 --log-level=trace --target-machine=$target --${step} --timeout $TIMEOUT < /dev/null > $TTY
```
ppenna, line 42: I am curious to learn why is this required.

csegarragonz: I added some info on the commit message.

Basically, when running some of the combination of tests (e.g. `debug`, `test`, `qemu-isapc`) my `stdin` would become unusable after execution.

My hunch is that its got something to do with the way we call `qemu-*` in `./scripts/run.sh`, but I did not dig deep into it.

Issue #541, opened 2025-05-21
[build] Error installing javy with latest toolchain


csegarragonz:
## Description

I was building all the optional dependencies from scratch, and came across an error when installing javy. I think this is related to javy pinning to a stable rust version, and the stable version having moved.

In summary, you should be able to reproduce the issue just following our instructions to **install** javy:

```bash
git clone https://github.com/nanvix/javy
cargo build -p javy-plugin --target=wasm32-wasip1 -r

# This is the failing command
cargo install --path crates/cli
```

The error has something to do with the reference-values feature not being enabled, which [it is not](https://github.com/nanvix/javy/blob/dev/.cargo/config.toml). This is a similar issue to [this one](https://github.com/rust-lang/rust/issues/128475) which seems to be caused by LLVM 19 making `reference-types` a default feature (see thread [here](https://github.com/rust-lang/rust/pull/128511)).

This issue is [fixed](https://github.com/bytecodealliance/javy/pull/896) in upstream javy, and I have checked it works by branching nanvix/javy, rebasing to `v5.0.4` and running the tests locally.

We basically just need to rebase `nanvic/javy` to `v5.0.4`, but I wanted to open this issue first. (Also opening a PR for a rebase would not work very well neither).

cc @ppenna @danbugs

csegarragonz:
PS: I was thinking that it is tricky to catch this regressions here, as, afaict, there is no place where we track the javy version.

ppenna:
@csegarragonz I created a branch some while ago to start keeping track of the javy version that we support: `wip-enhancement-contrib-javy`

ppenna:
Javy is no longer shipped with Nanvix.

PR #542, opened 2025-05-21
[ci] E: Add CI Script for GHA

csegarragonz:
In this commit I add a GHA workflow script that replicates the behaviour of ./scripts/pipeline.sh and caches the built toolchain for subsequent runs.

The toolchain cache is invalidated once we bump a minor or major version (not patch).

A few gotchas:
* The tests are, currently, skipping the javy related stuff. We need to address #541 first.
* We could do more fine-grained caching (e.g. we could cache the `./opt` repo in `debug` and `release` modes, indexed by the submodule hash, and only re-build them if the submodule changes). I have done things like this in the past.

To-Do:
* The toolchain is not versioned correctly (empty version string).

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,24 @@
+# Copyright(c) The Maintainers of Nanvix.
+# Licensed under the MIT License.
+
+name: "Set-up Nanvix build and test environment"
+
+runs:
+  using: "composite"
+  steps:
+  - name: "Install APT dependencies"
+    shell: bash
+    run: |
+      sudo -E ./scripts/setup/ubuntu.sh --extra
+
+  - name: "Install Rust"
+    shell: bash
+    run: |
+      curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+
+  - name: "Add rust toolchains"
+    shell: bash
+    run: |
+      source $HOME/.cargo/env
+      rustup component add rust-src
+      rustup target add wasm32-wasip1
--- a/.github/workflows/github-hosted-ci.yml
+++ b/.github/workflows/github-hosted-ci.yml
@@ -1,0 +1,103 @@
+# Copyright(c) The Maintainers of Nanvix.
+# Licensed under the MIT License.
+
+name: "Nanvix CI/CD (GitHub Hosted)"
+
+on:
+  push:
+    branches: [dev]
+  pull_request:
+    branches: [dev]
+    types: [opened, synchronize, reopened, ready_for_review]
+
+defaults:
+  run:
+    shell: bash
+
+# Cancel previous running actions for the same PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  setup-toolchain:
+    runs-on: ubuntu-24.04
+    outputs:
+      cache-key: ${{ steps.extract-version.outputs.cache_key }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: "Extract version"
+        id: extract-version
+        run: |
+          # This uses the MAJOR.MINOR as key in the cache
+          VERSION=$(grep '^version' Cargo.toml | head -n1 | sed 's/.*"\(.*\)".*/\1/')
+          echo $VERSION
+          MAJOR_MINOR=$(echo "$VERSION" | cut -d. -f1,2)
+          echo "cache_key=toolchain-${MAJOR_MINOR}" >> $GITHUB_OUTPUT
+
+      - name: "Restore toolchain cache"
+        id: cache
+        uses: actions/cache@v4
+        with:
+          path: ./toolchain
+          key: ${{ steps.extract-version.outputs.cache_key }}
+
+      - name: "Prepare build environemnt if not cached"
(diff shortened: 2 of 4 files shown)


ppenna (commented):
Within a target we want to respect the following run order: lint, build and test. Furthermore fast fail could also be enabled.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+        uses: ./.github/actions/setup
+        with:
+          install_apt: "true"
+
+      - name: "Run ${{ matrix.mode }} (${{ matrix.step }} on ${{ matrix.target }})"
+        run: |
+          export TOOLCHAIN_DIR=$(pwd)/toolchain
+          echo "Running '${{ matrix.mode }}' in '${{ matrix.step }}' mode for '${{ matrix.target }}'"
+          python3 scripts/ci.py --${{ matrix.mode }} --toolchain-dir=${TOOLCHAIN_DIR} --target-arch=x86 --log-level=trace --target-machine=${{ matrix.target }} --${{ matrix.step }}
+        timeout-minutes: 10
```
ppenna: We need to increase this. Unfortunately compilation of cpython takes too long.

csegarragonz:
@ppenna thanks for your feedback. Why does the "lint", "build", "test" order need to be preserved? Is it just to make sure we fail fast if the earlier phases fail?

I will enable fail-fast once all the tests are green. For now it is useful to disable it so that we have a complete picture of the tests that are failing.

Even then, I would argue that disablin fail-fast is a good option in order to see everything that fails (in case something fails).

ppenna:
> @ppenna thanks for your feedback. Why does the "lint", "build", "test" order need to be preserved? Is it just to make sure we fail fast if the earlier phases fail?
> 
> I will enable fail-fast once all the tests are green. For now it is useful to disable it so that we have a complete picture of the tests that are failing.
> 
> Even then, I would argue that disablin fail-fast is a good option in order to see everything that fails (in case something fails).

- Free github accounts have a maximum monthly utilization of 2000 minutes. Enforcing that order would help us to stay in the budget.
- Test rules implicitly build thus we can avoid significantly higher run times if we would cache the build artifacts as well and then run tests:
https://github.com/nanvix/nanvix/blob/39c8fe6c22f3858f114f672a589b981ba9c6bf83/Makefile#L792-L806
https://github.com/nanvix/nanvix/blob/39c8fe6c22f3858f114f672a589b981ba9c6bf83/Makefile#L821-L837
https://github.com/nanvix/nanvix/blob/39c8fe6c22f3858f114f672a589b981ba9c6bf83/Makefile#L856-L867)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+      # Build
+      - name: "Run ${{ matrix.mode }} (build on ${{ matrix.target }})"
+        run: |
+          export TOOLCHAIN_DIR=$(pwd)/toolchain
+          echo "Running '${{ matrix.mode }}' in 'build' mode for '${{ matrix.target }}'"
+          python3 scripts/ci.py --${{ matrix.mode }} --toolchain-dir=${TOOLCHAIN_DIR} --target-arch=x86 --log-level=trace --target-machine=${{ matrix.target }} --lint
+        timeout-minutes: 30
+
+      # Test
+      - name: "Run ${{ matrix.mode }} (test on ${{ matrix.target }})"
```
ppenna: We should also set the default timeout for this job to be 30 minutes.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,122 @@
+# Copyright(c) The Maintainers of Nanvix.
+# Licensed under the MIT License.
+
+name: "Nanvix CI/CD"
+
+on:
+  push:
+    branches: [dev]
```
csegarragonz: @ppenna this script will now (only) run on commits to `dev` branch, or PRs against `dev` with the status below (i.e. not in draft).

what other configurations would you like to enable?

ppenna: Let's run it on the same matching branches as before:

- dev
- enhancement-*
- feature-*
- bugfix-*

csegarragonz: Ok. Note that this will still run if someone opens a PR against `dev` that does not begin with `enhancement/feature/bugfix`. I will see if I can add that somehow.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,118 @@
...
+
+      # Test
+      - name: "Run ${{ matrix.mode }} (test on ${{ matrix.target }})"
+        run: |
+          export TOOLCHAIN_DIR=$(pwd)/toolchain
+          echo "Running '${{ matrix.mode }}' in 'test' mode for '${{ matrix.target }}'"
+
+          # Check if KVM is enabled
+          sudo kvm-ok
+          sudo lsmod | grep kvm
```
csegarragonz: As an idea, we could do this in a separate step, and fail if the runner does not have KVM (it should not happen, but just in case).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,118 @@
...
+          # Add user to kvm group
+          sudo usermod -aG kvm $USER
+
+          # Must run tests in sudo to use KVM.
```
csegarragonz: We must remember to update this comment. We are not running as sudo, we are running as `$USER`, but this forces re-evaluating the group membership.

PR #558, opened 2025-05-30
Enhancements to the linuxd-gateway communication

csegarragonz:
In this PR I introduce an enhancement to the linuxd-gateway communication where we merge the size and body into a single message, instead of separating them in two.

### DIFF
--- a/src/daemons/linuxd/src/linuxd/mod.rs
+++ b/src/daemons/linuxd/src/linuxd/mod.rs
@@ -620,23 +620,13 @@ impl<'a> LinuxDaemon<'a> {
                     // NOTE: we don't check if the write operation is too big, because its size is
                     // already bound by the maximum payload size of the message.
                     let count: usize = request.count as usize;
-                    let length_buffer: [u8; mem::size_of::<u32>()] = (count as u32).to_le_bytes();
-                    match conn.write_all(&length_buffer) {
+                    match conn.send_message_to_gateway(&request.buffer[..count]) {
                         Ok(_) => {
-                            match conn.write_all(&request.buffer[..count]) {
-                                Ok(_) => {
-                                    debug!("wrote {count} bytes to the gateway");
-                                    WriteResponse::build(source, count as i32)
-                                },
-                                Err(e) => {
-                                    debug!("failed to write buffer to the gateway (error={e:?})");
-                                    // TODO: Check error conversion.
-                                    build_error(source, ErrorCode::ConnectionReset)
-                                },
-                            }
+                            debug!("wrote {count} bytes to the gateway");
+                            WriteResponse::build(source, count as i32)
                         },
                         Err(e) => {
-                            debug!("failed to write length to the gateway (error={e:?})");
+                            debug!("failed to write buffer to the gateway (error={e:?})");
                             // TODO: Check error conversion.
                             build_error(source, ErrorCode::ConnectionReset)
                         },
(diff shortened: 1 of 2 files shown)


src/libs/syscomm/src/lib.rs
```diff
@@ -155,6 +157,71 @@ impl SocketStream {
             },
         }
     }
+
+    /// Read one message from the gateway compising a header with a size
+    /// (u32 LE) and then the body with as many bytes as indicated in the header.
+    pub fn read_message_from_gateway(&mut self) -> Result<Vec<u8>, ::std::io::Error> {
+        let mut buffer = Vec::new();
+        // TODO: define constant
+        let mut tmp = [0u8; 64];
```
csegarragonz: @ppenna any suggestions on what would be a good constant here? (for the size of the `tmp` buffer).

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -677,73 +667,56 @@ impl<'a> LinuxDaemon<'a> {
...
-                                },
-                                Err(e) => {
-                                    debug!("failed to read from the gateway (error={e:?})");
-                                    // TODO: Check error conversion.
-                                    build_error(source, ErrorCode::ConnectionReset)
-                                },
+                            count
+                        };
+
+                        // TODO: we have already copied the bytes into message so we should be able
```
csegarragonz: I am leaving this comment as a future TODO. I think the PR can be merged anyway.

src/libs/syscomm/src/lib.rs
ppenna: It should be `::config::kernel::IPC_MESSAGE_SIZE`

src/daemons/linuxd/src/linuxd/mod.rs
ppenna: Could you create an issue for that and link it in the comment?

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -677,73 +667,56 @@ impl<'a> LinuxDaemon<'a> {
...
+                        debug!("read {count} bytes from the gateway");
+
+                        // Truncate read request to fit in the response buffer.
+                        let read_count: usize = if count > request.count as usize {
+                            warn!(
+                                "handle_read_request(): truncating payload (requested={}, \
+                                 actual={count})",
+                                { request.count },
+                            );
+                            request.count as usize
```
ppenna, line 688: Instead of logging a warning message we should fail. If failing is not possible, we should panic. People don't fix warning messages, but people do fix critical errors.

src/libs/syscomm/src/lib.rs
```diff
@@ -155,6 +157,71 @@ impl SocketStream {
...
+                Err(e) => return Err(e),
+            }
+        }
+    }
+
+    /// Read one message from the gateway comprising a header with a size
+    /// (u32 LE) and then the body with as many bytes as indicated in the header.
+    pub fn send_message_to_gateway(&mut self, buffer: &[u8]) -> Result<(), SocketError> {
+        let mut out_buffer = Vec::with_capacity(mem::size_of::<u32>() + buffer.len());
+        let buffer_len: u32 = buffer.len().try_into().unwrap();
```
ppenna: Do not unwrap. Fail with an error instead.

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: This is not an error. This warning only means that we will chunk the input message into different messages to the micro VM.

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: I removed the comment because I think we need to copy, as the downstream function only accepts a `[u8; 45]`, so we need to copy our `[u8]` into that.

PR #559, opened 2025-05-31
Add echo-rust-server-nostd Benchmark

csegarragonz:


src/benchmarks/echo-rust-server-nostd/src/main.rs
```diff
@@ -0,0 +1,61 @@
...
+use ::syscall::{
+    sys::types::ssize_t,
+    unistd,
+};
+
+//==================================================================================================
+// Constants
+//==================================================================================================
+
+const MAX_REQUEST_SIZE: usize = 4096;
```
ppenna, line 24: Could you add a comment here to explain the rationale behind this magic value?

src/benchmarks/echo-rust-server-nostd/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+# Copyright(c) The Maintainers of Nanvix.
+# Licensed under the MIT License.
+
+[package]
+name = "echo-rust-server-nostd"
+version.workspace = true
+license-file.workspace = true
+edition.workspace = true
```
ppenna: Could you make this Rust Edition 2024 (do not inhereit from workspace).

src/benchmarks/echo-rust-server-nostd/src/main.rs
csegarragonz: I just copied and pasted it from the `echo-rust-nostd` [benchmark](https://github.com/nanvix/nanvix/blob/dev/src/benchmarks/echo-rust-nostd/src/main.rs#L24), so I am not sure about the rationale.

If you can type the rationale here I will add it as a comment.

PR #573, opened 2025-06-06
[utils] Add `nanvix-bench` Utility

csegarragonz:
In this PR I introduce `nanvix-bench` a small benchmark script that we can use to benchmark some basic performance metrics for Nanvix.

Right now, the script supports three different benchmarks:

```bash
./bin/nanvix-bench.elf -benchmark [cold-start,warm-start,echo-breakdown]
```

The `cold-start` experiment supersedes `./scripts/benchmark.py`. We should be able to delete it, but I did not want to do it without approval. 

The `echo-breakdown` experiment breaks down the time a packet spends in different parts of the data-path by modifying message buffers in-place, and including a timestamp. I add a macro, `timestamp_message!` to abstract this functionality, gated behind the `timestamp-messages` feature flag. This means that to run the `echo-breakdown` benchmark you need to re-compile Nanvix with an additional flag:

```bash
make all RELEASE=yes LOG_LEVEL=panic TIMESTAMP_MSG=yes
```

The benchmarks are designed in a way that prevents them from being ran if Nanvix has been compiled with the wrong flags.

I am working on another benchmark, similar to `echo-breakdown` but without `linuxd`, but this PR is already big in itself, so happy to include it in a different PR.


ppenna (commented):
Can't we use https://doc.rust-lang.org/core/mem/macro.offset_of.html to automatically compute offsets?

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -711,6 +713,7 @@ impl<'a> LinuxDaemon<'a> {
                             }
                         }
                         // Push EoF message.
+                        #[cfg(not(feature = "timestamp-messages"))]
```
ppenna, line 720: You may want to add a comment here why this.

Issue #576, opened 2025-06-06
[tests] Thread Tests are Failing Non-Deterministically


ppenna:
## Description

The `nanvixd-thread-c` test is experiencing non-deterministic failures. We need to investigate the underlying cause and implement a fix.

## What We Know So Far

- The issue appears to occur in environments with a low number of CPU cores. While it cannot be reproduced locally, it consistently happens on the CI machine, which has only 2 cores.
- When the bug manifests, it causes the system execution to hang, eventually leading to test timeouts.
- The kernel consistently hangs at the same point: during the creation of the worker thread in the `pthread_cond_static_init()` test.
- ~The bug appears to happen only on debug builds.~ It happened in a release build (https://github.com/nanvix/nanvix/actions/runs/15643831521/job/44080087295, https://github.com/nanvix/nanvix/actions/runs/15883105934)
- The bug appears to happen only in the microvm.
- Bug can be reproduced by applying the following patch:
```
diff --git a/src/tests/thread-c/cond_static_init.c b/src/tests/thread-c/cond_static_init.c
index dbf81c91..6c8cef6a 100644
--- a/src/tests/thread-c/cond_static_init.c
+++ b/src/tests/thread-c/cond_static_init.c
@@ -41,6 +41,8 @@ static void *worker_thread(void *arg)
     // Check if worker argument matches the expected value.
     assert((size_t)arg == EXPECTED_WORKER_ARG);
 
+    sched_yield();
+
     // Wait for the main thread to signal the condition variable.
     assert(pthread_mutex_lock(&mutex) == 0);
     while (!initialized) {
```

## Logs

- [nanvix-v0.6.14-nanvixd-thread-c.elf.zip](https://github.com/user-attachments/files/20634708/nanvix-v0.6.14-nanvixd-thread-c.elf.zip)
- [nanvix-v0.6.16-nanvixd-thread-c.elf.zip](https://github.com/user-attachments/files/20639074/nanvix-v0.6.16-nanvixd-thread-c.elf.zip)
- [nanvix-v0.6.22-nanvixd-thread-c.elf.zip](https://github.com/user-attachments/files/20663087/nanvix-v0.6.22-nanvixd-thread-c.elf.zip)

ppenna:
Happened here: https://github.com/nanvix/nanvix/actions/runs/15883105934

PR #582, opened 2025-06-09
[ci] FIx slow and unintended behaviour of python-lint

csegarragonz:
In this PR I address two very annoying side-effects of the python-lint post-commit hook I introduced.

In the first commit I make the lint re-use a virtual environment, instead of creating a new one all the time. This should address the latency introduced at each commit.

In the second commit I refine the logic to check whether any files where modified by the lint, by only diffing on `*.py` files.

### DIFF
--- a/.githooks/post-commit
+++ b/.githooks/post-commit
@@ -7,6 +7,6 @@ set -euo pipefail
 
 make python-lint > /dev/null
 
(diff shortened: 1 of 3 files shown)


ppenna:
Issue https://github.com/nanvix/nanvix/issues/576 was triggered when running the CI.

PR #586, opened 2025-06-10
[microvm] Sets Up Plumbing for Snapshots

LucasGdosR:
Files changed:
1. VirtualProcessor (`src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs`): added more state (channel). Added new struct VirtualProcessorHandle (migrate to separate file?).
2. Emulator (`src/microvm/src/vmm/microvm/kvm/emulator.rs`): added more state (channel).
3. MicroVm (`src/microvm/src/vmm/microvm/microvm.rs`): passes more args to instantiate Emulator due to the added state. Added a method to get the VirtualProcessorHandle.
4. Vmm (`src/microvm/src/vmm/microvm/mod.rs`):
    - added exports to make structs visible to the IoThread (move the `io.rs` file around to avoid this?).
    - quick-fix1: unified methods `new` and `run`. This was done to be able to spawn a new thread for `run`, while saving the vCPU's thread handle inside the IoThread.
    - reduced state: the Vmm struct no longer holds the IoThread's handle.
    - created a new channel between the IoThread (receiver) and the Emulator (sender), which is passed to the MicroVm constructor.
    - quick-fix2: `new`'s return value should be that of `run`, so it can serve the purposes that `run`'s return value served. However, I just said screw error handling and returned Ok(()).
5. main (`src/microvm/src/main.rs`): since `new` and `run` were unified, main now calls only `new`.
6. IoThread (`src/microvm/src/io.rs`):
    - added more state for communication (handles / channel).
    - added a new enum for the OrchestratorStates in the snapshotting protocol (migrate to separate file?).
    - quick-fix3: modified the run method: instead of the infinite loop, now loop while the virtual processor's thread is running.
_________________________________________________________________________________
About the quick-fixes:
1. In order to pause the vCPUs, the orchestrator (IoThread) must have handles to the threads to send signals. The way it was, the main thread ran the vCPU, so there was no way for me to pass its handle to the IoThread. I unified `new` and `run` because the time to spawn the vCPU thread is in `run`, and the time the IoThread is spawned is in `new`. That was the least intrusive manner to pass the handle to the vCPU to the IoThread. I think some more refactoring is in order to maintain the methods separated.
2. Fixing this for real depends on the other quick-fixes. If things remain as they are, then the IoThread can propagate the vCPU thread's return code and everything will work out fine.
3. Is it ok to leave the IoThread doing this test every round, or is the overhead large?

### DIFF
--- a/src/microvm/src/main.rs
+++ b/src/microvm/src/main.rs
@@ -97,6 +97,7 @@ fn main() -> Result<ExitCode> {
         },
         None => None,
     };
+    
     let mut vmm: Vmm =
         Vmm::new(memory_size, &kernel_filename, initrd_filename, initrd_args, stderr, gateway)?;
 
--- a/src/microvm/src/main.rs
+++ b/src/microvm/src/main.rs
@@ -97,7 +97,7 @@ fn main() -> Result<ExitCode> {
         },
         None => None,
     };
-    
+
     let mut vmm: Vmm =
         Vmm::new(memory_size, &kernel_filename, initrd_filename, initrd_args, stderr, gateway)?;
 
--- a/src/microvm/src/main.rs
+++ b/src/microvm/src/main.rs
@@ -97,9 +97,17 @@ fn main() -> Result<ExitCode> {
         },
         None => None,
     };
-
+cfg_if::cfg_if! {
+    if #[cfg(feature = "hyperlight")] {
+        let mut vmm: Vmm =
+            Vmm::new(memory_size, &kernel_filename, initrd_filename, initrd_args, stderr, gateway)?;
+        let run = ||vmm.run();
+    } else {
+        let run = ||Vmm::quick_fix(memory_size, &kernel_filename, initrd_filename, initrd_args, stderr, gateway);
+    }
+}
     // Run virtual machine and check exit status code.
-    match Vmm::quick_fix(memory_size, &kernel_filename, initrd_filename, initrd_args, stderr, gateway)? {
+    match run()? {
         exit_status if exit_status != 0 => {
             let exit_code: u8 = match exit_status.try_into() {
                 Ok(code) => code,
--- a/src/microvm/src/main.rs
+++ b/src/microvm/src/main.rs
@@ -97,17 +97,11 @@ fn main() -> Result<ExitCode> {
         },
         None => None,
     };
-cfg_if::cfg_if! {
-    if #[cfg(feature = "hyperlight")] {
-        let mut vmm: Vmm =
-            Vmm::new(memory_size, &kernel_filename, initrd_filename, initrd_args, stderr, gateway)?;
-        let mut run = ||vmm.run();
-    } else {
-        let run = ||Vmm::quick_fix(memory_size, &kernel_filename, initrd_filename, initrd_args, stderr, gateway);
-    }
-}
+    let mut vmm: Vmm =
+        Vmm::new(memory_size, &kernel_filename, initrd_filename, initrd_args, stderr, gateway)?;
+
     // Run virtual machine and check exit status code.
-    match run()? {
+    match vmm.run()? {
         exit_status if exit_status != 0 => {
             let exit_code: u8 = match exit_status.try_into() {
                 Ok(code) => code,
--- a/src/microvm/src/main.rs
+++ b/src/microvm/src/main.rs
@@ -98,11 +98,8 @@ fn main() -> Result<ExitCode> {
         None => None,
     };
 
-    let mut vmm: Vmm =
-        Vmm::new(memory_size, &kernel_filename, initrd_filename, initrd_args, stderr, gateway)?;
-
     // Run virtual machine and check exit status code.
-    match vmm.run()? {
+    match Vmm::new(memory_size, &kernel_filename, initrd_filename, initrd_args, stderr, gateway)? {
         exit_status if exit_status != 0 => {
             let exit_code: u8 = match exit_status.try_into() {
                 Ok(code) => code,
--- a/src/microvm/src/main.rs
+++ b/src/microvm/src/main.rs
@@ -99,7 +99,7 @@ fn main() -> Result<ExitCode> {
     };
 
     // Run virtual machine and check exit status code.
-    match Vmm::new(memory_size, &kernel_filename, initrd_filename, initrd_args, stderr, gateway)? {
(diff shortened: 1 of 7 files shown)


LucasGdosR:
Changes:
- Changed the name of `new` to `quick_fix`, because there was an error that `new` didn't return `Self`;
- Used compile-time macros to make `main` compatible with hyperlight and microvm and the `new` + `run` = `quick_fix` change;
- Left the old version of `io.rs` as it were for hyperlight, and moved the modified `io.rs` into the vmm directory;
- Removed exports now that `io.rs` is inside the vmm directory;
- Fixed a bug when the `-gateway` command-line argument was absent.

src/microvm/src/vmm/microvm/kvm/emulator.rs
```diff
@@ -36,6 +38,8 @@ pub struct Emulator {
     input: Box<InputFn>,
     /// Output function used for emulating I/O port writes.
     output: Box<OutputFn>,
+    /// Channel to tell the orchestrator all vcpus have paused.
+    _paused_tx: Sender<Message>,
```
ppenna, line 46: Please add a comment saying why we need to keep this around. It would help others to accidently removing it from the structure.

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -155,12 +155,32 @@ impl Vmm {
...
+                // Get the virtual processor handle to pass to the I/O thread.
+                let vcpu_handle: VirtualProcessorHandle = microvm
+                    .lock()
+                    .map_err(|e| anyhow::anyhow!("failed to acquire lock {e:?}"))?
+                    .get_vcpu_handle(vcpu_thread_handle); 
+
+                // Spawn I/O thread.
+                let io_thread: JoinHandle<Result<()>> = IoThread::spawn(
+                conn, gateway_rx, gateway_tx, microvm, vcpu_handle, paused_rx);
+                io_thread.join().unwrap();
```
ppenna: Do not `unwrap()`. Fail with a log instead.

src/microvm/src/vmm/microvm/kvm/emulator.rs
LucasGdosR: Added:
// NOTE(gribel): when there are multiple vcpus, there will be a single stdin / stdout.
//   Having this channel in the Emulator instead of in each vcpu
//   reduces the number of messages between the orchestrator and the vcpus from O(n) to O(1),
//   with `n` being the number of vcpus.

src/microvm/src/vmm/microvm/mod.rs
LucasGdosR: Changed to:
```
            io_thread.join()
                    .map_err(|e| anyhow::anyhow!("failed to join I/O thread {e:?}"))?;
                Ok(0)
            }
            None => {
                vcpu_thread_handle.join()
                    .map_err(|e| anyhow::anyhow!("failed to join virtual processor thread {e:?}"))?
            }
```
Should it be `anyhow::bail!` instead or is this fine?

src/microvm/src/main.rs
ppenna, line 1: We should revert this.

src/microvm/src/vmm/microvm/kvm/emulator.rs
```diff
@@ -16,10 +16,12 @@ use crate::vmm::microvm::{
     },
 };
 use ::anyhow::Result;
-use std::sync::{
+use ::std::sync::{
```
ppenna, line 19: Thanks!

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -85,7 +91,7 @@ impl MicroVm {
     /// Upon successful completion, this method returns the MicroVM that was created. Otherwise, it
     /// returns an error.
     ///
-    pub fn new(memory_size: usize, input: Box<InputFn>, output: Box<OutputFn>) -> Result<Self> {
+    pub fn new(memory_size: usize, input: Box<InputFn>, output: Box<OutputFn>, paused_tx: Sender<Message>) -> Result<Self> {
```
ppenna, line 95: We should update documentation accordingly.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// Runs the I/O thread.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, empty is returned. Otherwise, an error is returned instead.
+    ///
+    pub fn run(&mut self) -> Result<u16> {
+        if let Some(mut vcpu_handle) = self.vcpu_handle.take() {
+            if let Some(thread_handle) = vcpu_handle.vcpu_thread.take() {
+                while !thread_handle.is_finished() {
```
ppenna, line 134: If we fail for any reason how would it impact a snapshot that is in progress?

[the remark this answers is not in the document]
ppenna: This is intentional for now.

LucasGdosR: For now, I'll assert those `if let Some(_)` statements are always true, as `io_thread.run()` is only called when they are true.

Failures around the vCPU threads do not affect a snapshot that is in progress, as they must be paused for the snapshot to start, and the snapshot must be finished for them to resume.

All of the communication inside the loop will also stop during snapshots. Those functions won't even be called, so they cannot fail. The IoThread will only poll for messages to resume the microvm, create a snapshot, or be killed. Off the top of my head, if creating a snapshot does not override an existing snapshot, it should be fine to kill the IoThread while a snapshot is in progress. If a `resume` message comes while a snapshot is in progress, it should either be staggered until the snapshot is done (via `join()` with a `JoinHandle` that's creating the snapshot), or be ignored entirely.

LucasGdosR:
I believe this pull request is ready to be merged. All the documentation is up-to-date, and it didn't break anything. For the next steps, should I create a new branch, or should I keep working on this one?

ppenna:
> I believe this pull request is ready to be merged. All the documentation is up-to-date, and it didn't break anything. For the next steps, should I create a new branch, or should I keep working on this one?

@LucasGdosR  Please rebase this branch with latest dev and open a PR with a branch that matches the name rules to trigger the ci (ie: `feature-microvm-snapshot`.

ppenna:
Superseded by #680

PR #603, opened 2025-06-14
[linuxd] Move Gateway Listener to Linuxd

csegarragonz:
In this PR I change the communication pattern between the gateway and linuxd by moving the listener socket inside linuxd and `connect`-ing from the gateway.

This means that we now have two listener sockets inside linuxd: one for the user VM, and one for the gateway. This forced a rename of the arguments in linuxd. We cannot use `bind-addr` anymore, as both sockets `bind`. I have picked `-uvm-bind-addr` and `-gateway-bind-addr`. I know user VM is not an extended terminology in the codebase, but iiuc we are converging towards it, so might as well rename it to the right thing. I am happy to use another name, though.

When refactoring `nanvixd` I ran into a few race conditions about UNIX socket files not being cleaned-up after executions (when things went wrong). This was a long-standing issue when benchmarking Nanvix. To address it cleanly, I implement the `Drop` trait for the `SocketListener` enum in `syscomm`. This forced a small API change, but nothing too majer.

### DIFF
--- a/src/daemons/linuxd/src/main.rs
+++ b/src/daemons/linuxd/src/main.rs
@@ -42,19 +42,10 @@ use ::flexi_logger::{
     FileSpec,
     Logger,
 };
-use ::signal_hook::{
-    consts::SIGINT,
-    iterator::{
-        Signals,
-        SignalsInfo,
-    },
-};
 use ::std::{
     env,
-    fs,
     str::FromStr,
     sync::Once,
-    thread,
 };
 use ::sys::{
     error::ErrorCode,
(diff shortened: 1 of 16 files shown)


src/daemons/linuxd/src/main.rs
```diff
@@ -102,87 +97,85 @@ pub fn main() -> Result<()> {
...
     let uvm_listener: SocketListener = match Socket::bind(uvm_bind_socket_type, uvm_sockaddr.clone()) {
         Ok(listener) => listener,
         Err(e) => {
-            error!("failed to bind to uVM socket address (error={e:?})");
+            error!("failed to bind to uVM socket address (address={}, error={e:?})", uvm_sockaddr.clone());
             anyhow::bail!("failed to bind to uVM socket address");
         },
     };
 
-    // Install signal handler.
```
csegarragonz, line 113: We do not need the signal handler anymore, because the `Drop` trait introduced in the previous commit will take care of the clean-up. Also note that this handler was an incomplete clean-up, as it was not covering the cases where an internal error happened after binding to a socket.

ppenna (commented):
LGTM. Before I proceed with the merge, could you rename UVM (and all variants) to user-vm (strings) or user_vm (code)? The UVM acronym may be misleading (make some people think of it as micro-vm or utility-vm)

src/utils/nanvixd/src/args.rs
```diff
@@ -68,8 +68,8 @@ impl Args {
                     i += 1;
                     keep_alive_timeout = Duration::from_secs(args[i].parse::<u64>()?);
                 },
-                _ => {
-                    return Err(anyhow::anyhow!("invalid argument"));
+                arg => {
```
ppenna, line 71: That's a good improvement :D Sometimes, I would even myself unsure why my CLI was wrong :D

PR #609, opened 2025-06-20
[utils] Add `warm-start-vmm` Benchmark to `nanvix-bench`

csegarragonz:
In this PR I introduce a new micro-benchmark similar to the warm-start one, where we only measure the time spent inside the VMM.

To do so, we connect the user VM to a UNIX stream that behaves like linuxd. We wait for the app to send us a ReadRequest, and then measure the time elapsed between sending the ReadResponse, and receiving a WriteRequest, which is the behaviour of the echo application.


ppenna (dismissed):
LGTM

PR #666, opened 2025-07-08
[build] Require TOOLCHAIN_DIR Env. Var. To Be Set

csegarragonz:
With the latest toolchain upgrades, it is no longer supported to have the toolchain in `$NANVIX_ROOT/toolchain`, instead it is required (by `./scripts/setup/rust.sh`) for `$TOOLCHAIN_DIR` to be out-of-tree. In the Makefile, however, we still provide a default value for `TOOLCHAIN_DIR` that is invalid.

In this commit I remove the default assignment to `$TOOLCHAIN_DIR`, and instead require the variable to be set in the calling environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,11 @@ else
 export IMAGE ?= nanvix.iso
 endif
 
+# Make sure toolchain directory is set in the environment.
+ifndef TOOLCHAIN_DIR
+$(error Environment variable TOOLCHAIN_DIR is not set.)
+endif
+
 #===================================================================================================
 # Directories
 #===================================================================================================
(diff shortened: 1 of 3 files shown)


Makefile
ppenna, line 1: As per documented here, we encourage the creation of a symbolic link to the target directory, as `pipeline.sh` requies that to work:

https://github.com/nanvix/nanvix/blob/dev/doc/setup.md#step-2-build-a-cc-cross-compiler-toolchain

My suggestion is to drop changes in `Makefile` and `setup.yml`.

csegarragonz: Oh! I had missed the updates in the documentation. That is an alternative way to fix my issue. I will close this PR then.

PR #667, opened 2025-07-08
[libs] Fix Reading Long Messages From Gateway

csegarragonz:
In this commit I fix a bug I introduced when writting ad-hoc methods to read/write from/to the gateway.

In particular, if the read buffer exceeded the IKC message size, we dropped the message with an error of "Message too large". This is incorrect, as we may want to read an arbitrarily long message from the gateway. Note that the chunking in small pieces happens in linuxd, not in the syscomm lib.

### DIFF
--- a/src/libs/syscomm/src/lib.rs
+++ b/src/libs/syscomm/src/lib.rs
@@ -183,8 +183,9 @@ impl SocketStream {
     /// (u32 LE) and then the body with as many bytes as indicated in the header.
     pub fn read_message_from_gateway(&mut self) -> Result<Vec<u8>, ::std::io::Error> {
         let mut buffer = Vec::new();
-        let mut tmp = [0u8; ::config::kernel::IPC_MESSAGE_SIZE];
+        let mut tmp = [0u8; config::syscomm::GW_READ_BUFFER_LEN];
         let u32_size = mem::size_of::<u32>();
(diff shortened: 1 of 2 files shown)


src/libs/syscomm/src/lib.rs
```diff
@@ -28,6 +28,12 @@ use ::std::{
     },
 };
 
+//==================================================================================================
+// Constants
+//==================================================================================================
+
+const MAX_MESSAGE_LEN: usize = 128 * 1024 * 1024;
```
ppenna: Could you add a documentation here that explains the rationale for the value of this constant? Also, could we move it to the config crate? We are putting all such constants there.

src/libs/syscomm/src/lib.rs
```diff
@@ -183,8 +189,10 @@ impl SocketStream {
     /// (u32 LE) and then the body with as many bytes as indicated in the header.
     pub fn read_message_from_gateway(&mut self) -> Result<Vec<u8>, ::std::io::Error> {
         let mut buffer = Vec::new();
-        let mut tmp = [0u8; ::config::kernel::IPC_MESSAGE_SIZE];
+        const READ_BUFFER_LEN: usize = 4096;
```
ppenna: Could you add a rationale for the value of this constant?

ppenna (approved):

PR #680, opened 2025-07-13
[microvm] Sets Up Plumbing for Snapshots

LucasGdosR:
# What the PR aims to do
This sets up "plumbing" for communications in a snapshotting protocol.

# How it's done
The `IoThread` becomes the coordinator of the protocol. It got:
- An internal state machine;
- A handle to the `MicroVm` to call snapshot methods (which don't exist yet);
- A handle to the vCPU thread to send signals;
- A `Sender` channel to issue commands pause and resume;
- A `Receiver` channel to receive a `paused` response (this is much like the `ready` response in 2PC);

The thread handle and the `Sender` channel might seem redundant, but they're not. I'm not entirely sure why yet.

The channels are passed down to their destinations in the `Emulator` and the `Vcpu`.

When there's no `-gateway` command-line argument, the main thread executes the microvm application directly. Otherwise, the main thread runs the `IoThread` and joins the application thread.


src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -30,10 +30,19 @@ use ::kvm_ioctls::{
...
+        mpsc::{
+            Receiver,
+            Sender,
+            self,
+        },
+        Mutex,
+    },
+    thread::JoinHandle,
 };
+use ::sys::ipc::Message; // Placeholder. Will probably be VcpuEvent instead of Message.
```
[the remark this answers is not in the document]
ppenna: @LucasGdosR could you create a unit structure for `VcpuEvent` and place a comment on it saying that it is a place holder for future implementations? This will make it clearer and prevent misuse of those fields ie: `paused_tx` through strong type checking. Better yet would be to also create an issue and link it in the unit structure definition as a `TODO` comment.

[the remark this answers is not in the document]
LucasGdosR: @ppenna I created the unit struct as well as the issue https://github.com/nanvix/nanvix/issues/686 and linked it in the comment.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -0,0 +1,264 @@
...
+    ///
+    /// # Returns
+    ///
+    /// Upon success, the vCPU exit code is returned. Otherwise, an error is returned instead.
+    ///
+    pub fn run(&mut self) -> Result<u16> {
+        // This function only gets called when both Options are Some.
+        if let Some(mut vcpu_handle) = self.vcpu_handle.take() {
+            if let Some(thread_handle) = vcpu_handle.vcpu_thread.take() {
+                while !thread_handle.is_finished() {
```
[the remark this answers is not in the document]
LucasGdosR: This is intentional for now.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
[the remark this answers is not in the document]
ppenna: Thanks!

src/microvm/src/vmm/microvm/io.rs
ppenna, line 1: As this file is specific to hyperlight, we should put it in the `hyperlight module:

```
+ src/vmm/
  + hyperlight/
    + mod.rs
    + io.rs
```

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -0,0 +1,264 @@
...
+//==================================================================================================
+// Enums
+//==================================================================================================
+
+///
+/// # Description
+/// 
+/// States relating to snapshots functionality.
+/// Snapshots may be loaded at PreBoot, and created at Paused.
+///
```
ppenna: Please remove trailing whitespace.

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -181,10 +187,28 @@ impl Vmm {
     /// Otherwise, it returns an error.
     ///
     pub fn run(&mut self) -> Result<u16> {
-        self.microvm
+        // If a gateway connection exists, the IoThread must have a handle to all vCPU threads.
+        if let Some(mut io_thread) = self.io_thread.take() {
+            let microvm = self.microvm.clone();
```
ppenna: Please add type annotation to be consistent with the code base.

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -181,10 +187,28 @@ impl Vmm {
...
+            let vcpu_thread_handle: JoinHandle<Result<u16>> = std::thread::spawn(move || microvm
+                .lock()
+                .map_err(|e| anyhow::anyhow!("failed to acquire lock {e:?}"))?
+                .run()
+            );
+            let vcpu_handle: VirtualProcessorHandle = self.microvm
+                .lock()
+                .map_err(|e| anyhow::anyhow!("failed to acquire lock {e:?}"))?
+                .get_vcpu_handle(vcpu_thread_handle);
```
ppenna: Please remove trailing whitespace.

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -284,4 +291,24 @@ impl MicroVm {
...
+    /// Returns the handle to the virtual processor (thread handle and sending channel).
+    /// 
+    /// # Parameters
+    /// 
+    /// - `join_handle`: the handle to the virtual processor's thread.
+    /// 
+    /// # Returns
+    /// 
+    /// A structure with the handle to the virtual processor's thread
+    /// and a channel to send messages to the virtual processor.
```
ppenna: Remove trailing whitespace.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -97,16 +126,40 @@ impl VirtualProcessor {
...
+    /// # Description
+    ///
+    /// Gets a handle to the virtual processor for snapshotting.
+    ///
+    /// # Returns
+    ///
+    /// If the handle was already taken, this method crashes. Otherwise, it returns the handle.
+    pub fn get_handle(&mut self, join_handle: JoinHandle<Result<u16>>) -> VirtualProcessorHandle {
+        let event_tx = self.event_tx.take().expect("vcpu handle already taken");
+        // NOTE (gribel): for a virtual multiprocessor implementation,
```
ppenna: Remove your alias from the comment.

src/microvm/src/vmm/microvm/kvm/emulator.rs
```diff
@@ -36,6 +38,12 @@ pub struct Emulator {
     input: Box<InputFn>,
     /// Output function used for emulating I/O port writes.
     output: Box<OutputFn>,
+    /// Channel to tell the orchestrator all vcpus have paused and are ready for snapshots.
+    // NOTE(gribel): when there are multiple vcpus, there will be a single stdin / stdout.
```
ppenna: Remove your alias from the comment.

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -284,4 +291,24 @@ impl MicroVm {
...
+    /// 
+    // NOTE(gribel): for a multiprocessor implementation, join_handle should not be passed.
```
ppenna: Remove your alias from the comment.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -97,16 +126,40 @@ impl VirtualProcessor {
...
+    ///
+    /// # Description
+    ///
+    /// Gets a handle to the virtual processor for snapshotting.
+    ///
+    /// # Returns
+    ///
+    /// If the handle was already taken, this method crashes. Otherwise, it returns the handle.
+    pub fn get_handle(&mut self, join_handle: JoinHandle<Result<u16>>) -> VirtualProcessorHandle {
+        let event_tx = self.event_tx.take().expect("vcpu handle already taken");
```
ppenna: Add type annotation here to be consistent with the codebase.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -97,16 +126,40 @@ impl VirtualProcessor {
...
 
+    ///
+    /// # Description
+    ///
+    /// Gets a handle to the virtual processor for snapshotting.
+    ///
+    /// # Returns
+    ///
+    /// If the handle was already taken, this method crashes. Otherwise, it returns the handle.
+    pub fn get_handle(&mut self, join_handle: JoinHandle<Result<u16>>) -> VirtualProcessorHandle {
```
ppenna: We should avoid panicking on leaf routines. It is always better to propagate the result and let higher-level functions to decide what to do.

Change the the signature of this function to return and `Option<VirtualProcessorHandle>` and return `None`  if `self.event_tx` was already taken. Alternatively, we could just `self.event_tx.clone()`.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -70,8 +78,29 @@ pub struct VirtualProcessor {
...
+///
+/// A placeholder for future implementations: https://github.com/nanvix/nanvix/issues/686
+///
+pub struct VcpuEvent { }
+
+///
+/// # Description
+/// 
+/// A structure that sends messages to pause or resume a virtual processor.
+///
```
ppenna: Remove trailing whitespace.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -97,16 +126,40 @@ impl VirtualProcessor {
         };
         fd.set_fpu(&fpu)?;
 
+        let (event_tx, event_rx) = mpsc::channel::<VcpuEvent>();
```
ppenna: Why can't you create this at the `microvm` level, keep the `event_tx` there and pass down to `vcpu` the `event_rx` ? Note that by doing so you remove the ownership  transfer introduced by `get_handle()`.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -70,8 +78,29 @@ pub struct VirtualProcessor {
...
+pub struct VcpuEvent { }
+
+///
+/// # Description
+/// 
+/// A structure that sends messages to pause or resume a virtual processor.
+/// 
+pub struct VirtualProcessorHandle {
+    _event_tx: Sender<VcpuEvent>,
+    pub vcpu_thread: Option<JoinHandle<Result<u16>>>,
```
ppenna: Please don't use public member fields for this. I know that it is convenient, but it empowers  people to do unintended things with that. Create get/set methods, the compiler will inline them.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -70,8 +78,29 @@ pub struct VirtualProcessor {
...
+}
```
ppenna: Please remove the trailing whitespace to be consistent with the codebase.

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -85,10 +89,7 @@ impl Vmm {
         let (vm_tx, gateway_rx) = mpsc::channel::<Message>();
         let (gateway_tx, memory_thread_rx) = mpsc::channel::<Message>();
         let (memory_thread_tx, vm_rx) = mpsc::channel::<Message>();
-
-        // Spawn I/O thread.
-        let _io_thread: Option<JoinHandle<Result<()>>> =
-            gateway_conn.map(|conn| IoThread::spawn(conn, gateway_rx, gateway_tx.clone()));
+        let (paused_tx, paused_rx) = mpsc::channel::<Message>();
```
ppenna, line 92: You could  create `event_tx` and `event_rx` here as well.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -0,0 +1,264 @@
...
+/// 
+/// States relating to snapshots functionality.
+/// Snapshots may be loaded at PreBoot, and created at Paused.
+/// 
+enum OrchestratorState {
+    PreBoot,
+    Running,
+    Pausing,
+    PausingAndOutputFlushed,
+    Paused,
```
ppenna: As these states are unused for now, could you just remove them? We don't want to have dead code around.

ppenna (commented):
We are getting there :D

src/microvm/src/vmm/microvm/io.rs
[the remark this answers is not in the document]
ppenna: This is intentional.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// Queue of incoming messages.
+    incoming: VecDeque<Message>,
+    /// Queue of outgoing messages.
+    outgoing: VecDeque<Message>,
+    /// Connection to the snapshot interface.
+    //_snapshot_gateway: Gateway, // where snapshot commands will come from
+    /// MicroVM handle to issue snapshots.
+    _microvm: Arc<Mutex<MicroVm>>,
+    /// State in the snapshotting protocol.
+    _state: OrchestratorState,
```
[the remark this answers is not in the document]
ppenna: This is intentional.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -0,0 +1,264 @@
...
+    microvm_rx: Receiver<Message>,
+    /// Gateway sender.
+    microvm_tx: Sender<Message>,
+    /// Queue of incoming messages.
+    incoming: VecDeque<Message>,
+    /// Queue of outgoing messages.
+    outgoing: VecDeque<Message>,
+    /// Connection to the snapshot interface.
+    //_snapshot_gateway: Gateway, // where snapshot commands will come from
+    /// MicroVM handle to issue snapshots.
```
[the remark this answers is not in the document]
ppenna: This is intentional.

src/microvm/src/vmm/microvm/io.rs
LucasGdosR: I ran the hyperlight tests after the change and everything seems to be working fine.

src/microvm/src/vmm/microvm/io.rs
LucasGdosR: I have removed all trailing whitespace thoroughly and checked the new input for it as well. I assume you have some automated tooling to check for it. Could you share it?

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
LucasGdosR: I think I noticed some leaf functions that cannot fail return `Result`, such as a `Struct::new` function that wraps the arguments with the struct. Is it ok (pun intended) to return the struct instead of a `Result` in these cases?

src/microvm/src/vmm/microvm/mod.rs
LucasGdosR: That makes sense. I implemented it.

However, since I still needed to create the channel in `Vmm::new()` and I only got the `JoinHandle` in `Vmm::run()`, I had to store `event_rx` temporarily anyway, and had to transfer its ownership when instantiating the `VirtualProcessorHandle`.

I could either store it in the `Vmm` (as I did), or the `IoThread`. This choice is cosmetic. I could, however, forego the concept of a `VirtualProcessorHandle` and store `event_rx` and the `JoinHandle` as separate members of the `IoThread`, which would remove the need for ownership transfer. Would this be better? Do you propose another solution?

src/microvm/src/vmm/microvm/io.rs
ppenna: The right tools for doing this will depend on the IDE that you use. I use this extension for VS Code: https://marketplace.visualstudio.com/items?itemName=shardulm94.trailing-spaces

ppenna: Trailing spaces are still here.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -0,0 +1,261 @@
...
+    _microvm: Arc<Mutex<MicroVm>>,
+    /// State in the snapshotting protocol.
+    _state: OrchestratorState,
+    /// Handles to issue pause / resume commands.
+    pub vcpu_handle: Option<VirtualProcessorHandle>,
```
ppenna: This should not be pub. Please create accessors for this.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -0,0 +1,261 @@
...
+///
+/// # Description
+///
+/// States relating to snapshots functionality.
+/// Snapshots may be loaded at PreBoot, and created at Paused.
+/// TODO: add `Running`, `Pausing`, `PausingAndOutputFlushed`, and `Paused` states.
+///
+enum OrchestratorState {
+    PreBoot,
+}
```
ppenna, line 77: We are missing new line here.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
ppenna: If a function may fail, it should return a `Result`. We don't want to `panic!()`, `unwrap()` nor `expect()`. If there are functions in the code base that are not following this guideline, we should fix their signature. As this deviates from this PR, such changes could be taken incrementally by other PRs.

src/microvm/src/vmm/microvm/mod.rs
ppenna: > This choice is cosmetic.

I disagree. In one, you end up with forcing ownership transfers. Let's go with the solution where you forgo the concept of a `VirtualProcessorHandle`.

As a side note, I believe that the best long-term architectural decision would be to spawn a "vcpu thread" (which is currently running as the "main thread"), and turn the "iothread" into the "main thread".

LucasGdosR: I wasn't clear. I agree with you on both points.

The cosmetic choice I alluded to was another way to implement it that also required an ownership transfer. Getting rid of the `VirtualProcessorHandle` struct and keeping both members independently is not a cosmetic choice, as it allows for no transfers. I will implement it right away. EDIT: done.

As to the side note, that is precisely the way I implemented it [here](https://github.com/nanvix/nanvix/blob/feature-microvm-snapshot-plumbing/src/microvm/src/vmm/microvm/mod.rs#L204-L213):
```
            // Spawn vCPU thread
            let vcpu_thread: JoinHandle<Result<u16>> = std::thread::spawn(move || {
                microvm
                    .lock()
                    .map_err(|e| anyhow::anyhow!("failed to acquire lock {e:?}"))?
                    .run()
            });

            // The vcpu_handle exists here. It must be put into the IoThread. Then, the IoThread must run.
            io_thread.set_vcpu_thread(vcpu_thread);
            // Run the `IoThread` as the main thread
            io_thread.run()
```
Notice there's no longer a `spawn` method.

I'm glad we're in agreement :)

ppenna: I would say that we are almost on the same page. I pushed some changes to this branch as I found the ownership for the VMM thread handle (previously vCPU thread) messy.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -0,0 +1,257 @@
...
+    ///
+    /// Runs the I/O thread.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, the vCPU exit code is returned. Otherwise, an error is returned instead.
+    ///
+    pub fn run(mut self) -> Result<u16> {
+        // Run the I/O thread until the VMM thread finishes.
+        while !self.vmm_thread.is_finished() {
```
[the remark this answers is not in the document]
ppenna: This is intentional

src/libs/syscall/src/pthread/bindings/pthread_setcancelstate.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright(c) The Maintainers of Nanvix.
```
csegarragonz, line 1: Why is this file needed here?

src/microvm/src/elf.rs
```diff
@@ -239,13 +239,15 @@ pub unsafe fn load(
             if vaddr + memsz > max_offset {
```
csegarragonz, line 239: Did anything change at all here? Can probably discard the changes.

src/microvm/src/gateway.rs
```diff
@@ -95,7 +95,11 @@ impl Gateway {
                         return Err(SocketError::new(Error::new(ErrorKind::InvalidData, reason)));
                     },
                 };
-                profiler::timestamp_message!(&mut message.payload, mem::offset_of!(syscall::LinuxDaemonMessage, payload) + mem::offset_of!(syscall::unistd::message::ReadResponse, buffer));
+                profiler::timestamp_message!(
```
csegarragonz, line 98: Ditto as before. This seems like an auto-formatting that does not belong to this PR. 

We eventually need to run `cargo fmt` in the whole codebase and enforce this as part of CI.

src/microvm/src/vmm/hyperlight/io.rs
```diff
@@ -171,7 +171,11 @@ impl IoThread {
     fn try_receive_from_microvm(&mut self) -> Result<()> {
         match self.microvm_rx.try_recv() {
             Ok(mut message) => {
-                profiler::timestamp_message!(&mut message.payload, std::mem::offset_of!(syscall::LinuxDaemonMessage, payload) + std::mem::offset_of!(syscall::unistd::message::WriteRequest, buffer));
+                profiler::timestamp_message!(
```
csegarragonz, line 174: Changes unrelated to this PR.

src/microvm/src/vmm/hyperlight/mod.rs
```diff
@@ -118,8 +124,10 @@ impl Vmm {
                     // PEB, I/O buffers, host fxn defs, guard pages, etc.
                     let reserved_pages = 11 * 4096;
 
-                    let used_memory =
-                        kernel_size + initrd_size + (heap_size + stack_size) as usize + reserved_pages;
+                    let used_memory = kernel_size
```
csegarragonz, line 127: Most of this formatting changes are unrelated to the PR.

src/microvm/src/vmm/microvm/io.rs
csegarragonz: This file seems to be scpecific to the `microvm` not `hyperlight` unless I am missing something.

src/microvm/src/vmm/microvm/kvm/emulator.rs
```diff
@@ -36,6 +38,12 @@ pub struct Emulator {
     input: Box<InputFn>,
     /// Output function used for emulating I/O port writes.
     output: Box<OutputFn>,
+    /// Channel to tell the orchestrator all vcpus have paused and are ready for snapshots.
```
csegarragonz, line 41: Who is the orchestrator? This is not a term that we use commonly around the codebase.

src/microvm/src/vmm/microvm/kvm/vmem.rs
```diff
@@ -13,6 +13,7 @@ use crate::{
     },
 };
 use ::anyhow::Result;
+use ::arch::mem::PAGE_SIZE;
```
csegarragonz, line 16: Changes in this file are unrelated to the PR.

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -205,7 +218,8 @@ impl MicroVm {
         if let Some((_, initrd_size)) = self.initrd {
             if initrd_size > max_initrd_size {
                 return Err(anyhow::anyhow!(
-                    "initrd is too large (initrd_size={initrd_size}, max_initrd_size={max_initrd_size:?})",
+                    "initrd is too large (initrd_size={initrd_size}, \
```
csegarragonz, line 221: Change unrelated to the PR.

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -133,11 +135,15 @@ impl Vmm {
 
         // Create a thread that reads from vm_rx and writes to vm_rx2.
         let memory_thread_tx: Sender<Message> = memory_thread_tx.clone();
-        let memory_thread: JoinHandle<Result<(), anyhow::Error>> =
-            std::thread::spawn(move || loop {
+        let memory_thread: JoinHandle<Result<(), anyhow::Error>> = std::thread::spawn(move || {
```
csegarragonz, line 138: Change is unrelated to this PR.

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -157,10 +163,38 @@ impl Vmm {
                         // No message available.
                     },
                 }
+            }
+        });
+
+        // The IoThread is not spawned at this point. Only some of its state is saved.
+        let io_thread: Option<IoThread> = gateway_conn.map(|conn| {
+            let microvm_: Arc<Mutex<MicroVm>> = microvm.clone();
```
csegarragonz, line 171: Why trailing `_`?

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -181,10 +215,18 @@ impl Vmm {
     /// Otherwise, it returns an error.
     ///
     pub fn run(&mut self) -> Result<u16> {
-        self.microvm
-            .lock()
-            .map_err(|e| anyhow::anyhow!("failed to acquire lock {e:?}"))?
-            .run()
+        // Check if a gateway connection exits to decide how to run the VMM.
+        if let Some(io_thread) = self.io_thread.take() {
```
csegarragonz, line 219: There seems to be a potential race condition here.

This method gets called in line 181, and it will try and see if the `io_thread` attribute is set. However, this attribute is only set after line 181, in line 197. So you have a race between the main thread populating self, and the thread being `spawned`.

The nested way in which these two methods are intertwined seems very error-prone.

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -232,12 +274,20 @@ impl Vmm {
 
             match input_queue.recv() {
                 Ok(mut msg) => {
-                    profiler::timestamp_message!(&mut msg.payload, mem::offset_of!(syscall::LinuxDaemonMessage, payload) + mem::offset_of!(syscall::unistd::message::ReadResponse, buffer));
+                    profiler::timestamp_message!(
```
csegarragonz, line 277: Changes in this file are unrelated to the PR.

csegarragonz (changes requested):
Many thanks for the PR @LucasGdosR ! I have a few comments.

First, there are a number of changes in this PR (mostly formatting) that are unrelated to this PR and seem to be automatic formatting from your IDE. Please can you undo all changes that are not related to this PR? We need to do a formatting sweep of the whole repository, but that is a known issue.

What is the difference between `(paused_tx/rx)` and `(event_tx/rx)` is not `pause` an event? What about `resume`?

Even though I understand why you made the changes in `microvm/mod.rs`, I do not agree with the design. I don't think the I/O thread should own the VMM thread, it should, semantically, be the other way around. If you are missing some handles, I would suggest trying to make these handles available other than flipping the ownership relationship. This flip, in fact, creates a race condition in your code that I point out below.

Also, it is likely that the I/O thread will always be there in the future, because we still need a control-plane thread that we can use to shutdown or snapshot the micro VM.

Lastly, could you update your PR description to include some high-level description of what the PR aims to do, and how it does it? It would be more helpful than just a file changelog.

LucasGdosR:
Thanks for the review, @csegarragonz. The formatting changes were done by @ppenna. I'll address each comment individually, but I just wanted to get that out of the way, since it's a recurring theme in the review.

src/microvm/src/vmm/microvm/kvm/emulator.rs
LucasGdosR: The orchestrator is the `IoThread`. It will coordinate the states and transitions for pausing and resuming the `MicroVM`, and for creating and loading snapshots. Would it be clear if I used `IoThread` instead of `orchestrator`?

src/microvm/src/vmm/microvm/mod.rs
LucasGdosR: @ppenna wrote this. I believe the trailing `_` might be unnecessary. It should be fine to shadow the microvm inside the lambda scope, and then use the previous identifier outside the lambda scope to instantiate the `Vmm`. Right?

src/microvm/src/vmm/microvm/mod.rs
LucasGdosR: The method that gets called in line 181 is `MicroVm::run()`, and this is `Vmm::run()`. The `io_thread` attribute is set in the same thread as the call to `Vmm::run()`, in line 197, so there's no race condition.

LucasGdosR:
> What is the difference between (paused_tx/rx) and (event_tx/rx) is not pause an event? What about resume?

`event` is the terminology I found for CPU events, and is used only for direct communication related to the vCPU. I replicated the terminology here. The events trigger actions, such as pausing. I think it's based on event-based architecture.

`paused` is a term I chose myself. It informs that the pause is complete. There's no `resume_tx/rx`, because the protocol I envision does not have a `Resuming` state, only a `Pausing` state.

> Even though I understand why you made the changes in microvm/mod.rs, I do not agree with the design. I don't think the I/O thread should own the VMM thread, it should, semantically, be the other way around.

The issue I have with that design is for multiprocessor VMMs. There's always a single `IoThread`, but there might be multiple vCPUs in the future. Initially I named the `vmm_thread` as `vcpu_thread`. Does this line of thinking about multiple vCPUs change your mind? If not, how would you propose I got access to the main thread's handle? I'm not too familiar with Rust, so I might be missing some alternative. Finally, as to the race condition you pointed out, I think `MicroVm::run()` and `Vmm::run()` probably got mixed up. Is that the case?

> Also, it is likely that the I/O thread will always be there in the future, because we still need a control-plane thread that we can use to shutdown or snapshot the micro VM.

I agree, but I didn't get the implications of that.

EDIT: updated PR description.

ppenna:
> Even though I understand why you made the changes in microvm/mod.rs, I do not agree with the design. I don't think the I/O thread should own the VMM thread, it should, semantically, be the other way around. If you are missing some handles, I would suggest trying to make these handles available other than flipping the ownership relationship. 

I agree.

src/microvm/src/vmm/microvm/mod.rs
csegarragonz: You are right, I got mixed up with the different `run` methods. No race condition.

csegarragonz:
> event is the terminology I found for CPU events, and is used only for direct communication related to the vCPU. I replicated the terminology here. The events trigger actions, such as pausing. I think it's based on event-based architecture.
> 
> paused is a term I chose myself. It informs that the pause is complete. There's no resume_tx/rx, because the protocol I envision does not have a Resuming state, only a Pausing state.

Does this mean that if we want to add another feature/function we will need to add another TX/RX channel? So that I understand better, who is at both ends of the communication for the `event_tx/rx` and `paused_tx/rx`?

> The issue I have with that design is for multiprocessor VMMs. There's always a single IoThread, but there might be multiple vCPUs in the future. Initially I named the vmm_thread as vcpu_thread. Does this line of thinking about multiple vCPUs change your mind? If not, how would you propose I got access to the main thread's handle? I'm not too familiar with Rust, so I might be missing some alternative. Finally, as to the race condition you pointed out, I think MicroVm::run() and Vmm::run() probably got mixed up. Is that the case?

In my mind there is one main VMM thread, that is the VM monitor. It controls the I/O thread (used for I/O) and the different vCPU threads. If you needed a handle on the running vCPU, I would suggest running each vCPU as a different thread, different than the main VMM thread. I think the fact that the main thread also happens to run the vCPU is just a simplification for the case where we only have one vCPU.

> I agree, but I didn't get the implications of that.

Right now you have different code-paths for the situation where there is a gateway present, or there is not. This will become unnecessary because we will always need to poll from external events. The fact that there are these two different code paths is partially what confused me and made me believe there was a race.

LucasGdosR:
> Does this mean that if we want to add another feature/function we will need to add another TX/RX channel? So that I understand better, who is at both ends of the communication for the `event_tx/rx` and `paused_tx/rx`?

`event_tx/rx` sends `VcpuEvent`s from the `IoThread` to the `Vcpu`. `paused_tx/rx` sends `Message`s from the `Emulator` to the `IoThread`. This is plumbing internal to `Vmm`.

I plan on adding more communication pathways between the `IoThread` and the outside. More specifically, I need a connection to `linuxd` and some kind of connection to receive commands related to snapshotting. Those communication pathways could be more flexible, like an HTTP server or something like it, and could be reused for new features.

These channels (event/paused/tx/rx), however, were not intended to be reused for other features. Would that be acceptable?

> In my mind there is one main VMM thread, that is the VM monitor. It controls the I/O thread (used for I/O) and the different vCPU threads. If you needed a handle on the running vCPU, I would suggest running each vCPU as a different thread, different than the main VMM thread. I think the fact that the main thread also happens to run the vCPU is just a simplification for the case where we only have one vCPU.

I see. Thanks for the suggestion. I'll take it and implement it this way during the weekend, most likely. I was hesitant to do something like that, since there'd be nothing for the virtual monitor thread to do right now with a vCPU thread executing the application. I'll take your and ppenna's judgement call on this and add a new thread.

> Right now you have different code-paths for the situation where there is a gateway present, or there is not. This will become unnecessary because we will always need to poll from external events. The fact that there are these two different code paths is partially what confused me and made me believe there was a race.

I'm sorry about that. As there's no running IoThread thread when there's no gateway, I assumed it wouldn't exist without a gateway. With the added vCPU thread, this should get cleared up.

ppenna:
Superseded by #690

PR #684, opened 2025-07-14
[ci] Increase Benchmark Iterations

ppenna:
In this commit I increased the number of benchmark iterations from 100 to 1000.


csegarragonz (approved):
LGTM!

PR #697, opened 2025-07-20
[microvm] Snapshot state machine transitions

LucasGdosR:
# What the PR aims to do
This sets up the protocol for snapshots. All the states are traversed, and all channels send and receive messages. It still needs to build a connection to a client to send the snapshot commands, and it doesn't actually pause and resume the MicroVM. It's just setting up the state transitions with a mocked up implementation.

# How it's done
`IoThread::run()` now has a different behavior depending on the state in the protocol. This method has a few `TODO`s related to connections that aren't set up yet.

`IoThread::try_receive_from_snapshot_client()` is a mocked method. It simulates how a client might send commands to advance the protocol. It does communicate with the VMM, so the channels between the `IoThread` and the `Vmm` are tested by this method. `Vmm::handle_command()` is mocked on the other end to make this test possible.

The rest of the methods are self explanatory by their own names.

The documentation still needs work. There's no need to review it at this time.

Note the PR is from `wip-feature-snapshots` to `feature-microvm-snapshot-plumbing-fresh`. I don't actually intend to merge the two. This is so the changes are fewer than if it were from `wip` to `dev`, to make it clearer to review.


src/microvm/src/vmm/microvm/io.rs
```diff
@@ -155,36 +167,85 @@ impl IoThread {
...
+                OrchestratorState::PausingAndOutputFlushed => {
+                    // TODO:
+                    // Try to receive `LinuxDaemonFlushed` from linuxd to advance the snapshotting protocol.
+                    // Also try to receive data from linuxd while the `LinuxDaemonFlushed` response doesn't arrive.
+                    // Should be something like the while-loop from the `Pausing` state.
+
+                    // Flush the incoming messages into the buffer.
+                    // NOTE: this match-statement is nearly identical to `try_receive_from_gateway()`.
+                    // It could be substituted if that method returned a Result<bool> instead.
+                    loop {
```
LucasGdosR, line 225: The match-statement inside this loop is a lot like the one from `try_receive_from_gateway()`. That method returns `Result<()>`. If it returned `Result<bool>` instead, with `true` for when a message is received, and `false` for when there's no message to receive, I could reuse it here. Should I modify the existing method so it returns a `Result<bool>` or should I keep this nearly duplicated code?

ppenna: >  Should I modify the existing method so it returns a Result<bool> or should I keep this nearly duplicated code?

Duplicating code is bad. It means more things to maintain :) Please update the signature of that method and use it accordingly. Note that this modification should be a standalone commit.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -155,36 +167,85 @@ impl IoThread {
...
+                            },
+                        }
+                    }
+                    self.state = OrchestratorState::Paused;
+                },
+                // In the `Paused` state, wait for commands to create a snapshot, resume execution, or something else (migration, kill VM).
+                OrchestratorState::Paused => {
+                    // NOTE: make this a blocking call instead (another method without `try` in the name).
+                    self.try_receive_from_snapshot_client()?;
+                },
```
LucasGdosR, line 247: In the `Paused` state, the `IoThread` shouldn't receive anything from the Linux daemon or from the MicroVM. In this case, should I include a blocking implementation, since there's only one possible source of messages, or should I make it a polling implementation?

I figure a blocking implementation might lead to bugs down the line if a new communication channel comes up that could deliver messages while in the `Paused` state, but at the same time that communication channel doesn't exist yet.

ppenna: > I figure a blocking implementation might lead to bugs down the line if a new communication channel comes up that could deliver messages while in the Paused state, but at the same time that communication channel doesn't exist yet.

From this statement either:
- I don't buy that "since there's only one possible source of messages".
- I believe that if "a blocking implementation might lead to bugs down the line" the protocol is not behaving as it should. 

In the paused state, there should be no inflight messages.

LucasGdosR: There should be inflight messages for creating snapshots or resuming execution, at the very least. I figured I could block while waiting for them, since I know from which connection they'll come.

ppenna: We may not be on the same page about what I mean by in-flight messages. Let me rephrase that:

While paused, there should be no outstanding messages from the other communication channels.

I believe that I also dont understand the buggy scenario that you are describing. Can you try to elaborate more?

LucasGdosR: Ah, I see. Indeed, there shouldn't be outstanding messages from the other communication channels. There should only be messages arriving from a snapshotting client, which might be either "create snapshot" messages, or "resume" messages, or some other message which would have no effecf (e.g.: "pause"). Since the IoThread can only receive messages from the snapshot client, it could use a blocking receive.

The buggy scenario involves adding more features that require the IoThread to potentially receive messages from a new channel while in the paused state. In that case, blocking while waiting for a "create snapshot" / "resume" command might leave the new channel hanging indefinitely.

ppenna: > The buggy scenario involves adding more features that require the IoThread to potentially receive messages from a new channel while in the paused state.

We will not keep adding new channels. We have the ones for the data path, and you are now introducing the one for the control path.

LucasGdosR:
Superseded by #701

Issue #691, opened 2025-07-19
[utils] Boot Time Benchmark in nanvix-bench Is Failing


csegarragonz:
## Description

The `-boot-time` microbenchmark in `nanvix-bench` is not measuring what we think it is. In fact, every execution is erroring out.

This is because in `nanvix-bench` we unconditionally set the `-gateway` argument of `microvm.elf`, but in the `-boot-time` experiment we do not use `linuxd`, so there is no gateway socket.

## How to Reproduce

To reproduce, just comment out the stdout/stderr re-direction of the user VM to `/dev/null`. You may apply the following patch:

```
diff --git a/src/utils/nanvix-bench/src/main.rs b/src/utils/nanvix-bench/src/main.rs
index 0a982046..c5f14fb3 100644
--- a/src/utils/nanvix-bench/src/main.rs
+++ b/src/utils/nanvix-bench/src/main.rs
@@ -205,8 +205,8 @@ impl Benchmark {
         let nanovm_cmd = Command::new(&nanovm_args[0])
             .args(&nanovm_args[1..])
             .stdin(Stdio::null())
-            .stdout(Stdio::null())
-            .stderr(Stdio::null())
+            // .stdout(Stdio::null())
+            // .stderr(Stdio::null())
             .current_dir(get_proj_root())
             .spawn()?;
```

and re-build and re-run:

```bash
make all-host-binaries RELEASE=yes LOG_LEVEL=panic BUILD_OPT=no
./bin/nanvix-bench.elf -benchmark boot-time -iterations 10
```

Annoyingly, if we fix the issue the boot time balloons to 30-40 ms. This, we know, is not the real boot time, and I think may just be a delay introduced by `std::process::Child::wait()`, which is the method we are currently using to wait for the VM to finish.

Thus, a proper fix for this benchmark remains to be discussed.

PR #698, opened 2025-07-21
[utils] Refactor Nanvixd To Act As Unified Control-Plane

csegarragonz:
This PR is a first step towards having a more feature-rich control-plane in Nanvix using `nanvixd`. As a first step, the first thing I do is I re-factor `nanvix-bench` to use `nanvixd` under the hood. This way future regressions will be caught.

The other notable addition in this PR is that I introduce an HTTP API to interact with a running `nanvixd` instance. For the moment I ony add API calls: `NEW` and `KILL` to create a new user VM, and kill it. The `NEW` API call returns a user VM id (can be used for killing the user VM afterwards) and a UNIX socket address that can be used to interact with the VMs stdin/stdout.

The separation of the stdin/stdout processing uncovered an issue with our existing benchmarks, that also made it possible to get rid of the existing wire-format to send/receive messages from the gateway. To allow for multi-threaded interaction, this logic in linuxd is moved to dedicated IO threads. In the future these threads will also multiplex over streams from different user VMs associated to a linuxd instance.

I also simplify the handling logic of the sandbox cache, removing unecessary levels of nested locking: nanvixd is a multi-threaded HTTP server, and each thread must operate on the sandbox cache in mutual exclusion. Thus, one lock for the whole cache is enough. If this causes throughout issues in the future we can address it separately. I, too, remove all the `-keep-alive` logic from nanvixd which is now obsolete.

In the process I uncovered #691, and I update the benchmarks to not error (but be slow), hence why the `boot-time` benchmark results are noticeably slower. Other than that the results should be somewhat comparable (a bit slower for warm-starts).


src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -762,50 +816,55 @@ impl LinuxDaemon {
...
+                profiler::timestamp_message!(&mut request.buffer, 0);
                 let count: usize = request.count as usize;
-                let buffer: &[u8] = &request.buffer[..count];
-                if request.fd == STDERR_FILENO {
-                    let _ = io::stderr().write_all(buffer);
-                    let _ = io::stderr().lock().flush();
-                } else {
-                    let _ = io::stdout().write_all(buffer);
-                    let _ = io::stdout().lock().flush();
+                if let Err(error) = gateway_stdout_tx.send((request, env.get_stdout_response_tx())) {
```
csegarragonz: @ppenna here we are being slightly conservative with writes to stdout.

We first send the write request (`request`) then wait for the IO thread to push an (empty) response confirming it has written the value to the output stream (`env.get_stdout_response_rx().recv()` in the `match` statement below).

Alternatively, we could not wait at all for the IO thread to send a response, this would also save us having to pass the `venv` to `handle_write_request`. Given that the write operation in the actual output stream (i.e. gateway socket) is also asynchronous, I think it is safe to not wait for the IO thread's response, but I wanted to know your take before implementing it.

Makefile
```diff
@@ -829,37 +823,45 @@ clippy-microvm:
 # Rules for Running System Level Tests Using Nanvix Daemon
 #===================================================================================================
 
+# List of supported functions in hyperlight.
+HYPERLIGHT_WHITELIST := echo-c echo-cpp echo-rust-nostd hello-c hello-cpp
```
csegarragonz, line 827: @ppenna it was not very clear to me what tests can actually run in `hyperlight`. I was checking a succesful CI run, and observed that: (1) none of the linuxd tests were running for hyperlight, and (2) from nanvixd only a handful: https://github.com/nanvix/nanvix/actions/runs/16472768859/job/46566063191#step:5:3059.

I have surfaced this with this explicit allow list. Please let me know if any other tests need to be added here.

scripts/test-nanvixd.sh
```diff
@@ -70,32 +46,42 @@ if ! ss -tln | grep -q ":${NANVIXD_PORT_NUMBER} "; then
 fi
 
 # Run a client.
-CURL_STDOUT_FILE_NAME="${LOGS_DIR}/curl-stdout_$(date "+%Y_%m_%d_%H_%M").log"
-curl \
+TENANT_ID="foo"
+APP_NAME="bar"
```
ppenna: Could you move these constants to the begin of the file so that they are more visible?

src/utils/nanvix-bench/Cargo.toml
```diff
@@ -17,13 +17,16 @@ flexi_logger = { workspace = true }
 indicatif = { workspace = true }
 log = { workspace = true }
 microvm = { workspace = true }
+nanvixd = { path = "../nanvixd" }
```
ppenna: Could you change `Cargo.toml` files so we can refer it from the workspace as well?

src/daemons/linuxd/src/venv.rs
```diff
@@ -40,7 +40,11 @@ pub struct VirtualEnvironment {
     /// Identifier.
     id: VirtualEnvironmentIdentifier,
     /// Standard input messages not yet consumed.
-    stdin_messages: VecDeque<Message>,
+    // stdin_messages: VecDeque<Message>,
```
ppenna: Could you remove this dead code and add documentation to the new member fields?

src/daemons/linuxd/src/venv.rs
```diff
@@ -84,31 +94,20 @@ impl VirtualEnvironment {
...
+    pub fn get_stdin_response_rx(&mut self) -> &mut Receiver<Message> {
+        &mut self.stdin_response_rx
+    }
+
+    pub fn get_stdout_response_tx(&self) -> Sender<Message> {
+        self.stdout_response_tx.clone()
+    }
+
+    pub fn get_stdout_response_rx(&mut self) -> &mut Receiver<Message> {
+        &mut self.stdout_response_rx
```
ppenna: Could you add documentation keep consistency with the file?

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -157,26 +159,101 @@ pub struct LinuxDaemon {
...
+                                .map_err(|e| anyhow::anyhow!("failed to acquire lock on GW stdin: {e:?}"))?
+                                .write_all(&write_request.buffer[..write_request.count as usize])?;
+
+                            // We send the same message to signal that we are done, but downstream
+                            // does not need it.
+                            response_tx.send(
+                                WriteResponse::build(0.into(), write_request.count as i32))?;
+                        }
+                        Err(RecvError) => {
+                            return Err(anyhow::anyhow!("gateway STDOUT channel disconnected"));
```
ppenna: Could you log an error before failing?

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -157,26 +159,101 @@ pub struct LinuxDaemon {
...
+
+            // Make sure that the input stream from the gateway is set to blocking, as otherwise we
+            // would not be able to differentiate between an EOF and a race between the application
+            // code and the gateway.
+            let gw_stdin_stream = gateway_stream.clone();
+            gw_stdin_stream
+                .lock()
+                .unwrap()
+                .set_nonblocking(false)
+                .unwrap();
```
ppenna: We should enforce this at creation time and enforce semantics using typestate: `BlockingSocketStream`. We can take it in a different PR. If so, could you just create an issue for that and link it here?

src/daemons/linuxd/src/linuxd/mod.rs
ppenna: It is safe to not wait.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -814,105 +873,60 @@ impl LinuxDaemon {
...
-                        #[cfg(not(feature = "timestamp-messages"))]
-                        env.push_stdin_message(ReadResponse::build(
-                            source,
-                            0,
-                            [0u8; ReadResponse::BUFFER_SIZE],
-                        ));
-
-                        ReadResponse::build(source, read_count as c_ssize_t, response_buf)
+            // Send ReadRequest to gateway IO thread.
+            if let Err(error) = gateway_stdin_tx.send((request, env.get_stdin_response_tx())) {
```
ppenna, line 886: We `send()` while holding the lock. If `send()` is blocked, do we have any unwanted side effects?

src/utils/nanvix-bench/src/env.rs
```diff
@@ -0,0 +1,7 @@
+// Copyright(c) The Maintainers of Nanvix.
+// Licensed under the MIT License.
+
+pub fn get_proj_root() -> String {
+    format!("{}/../../..", env!("CARGO_MANIFEST_DIR"))
+}
+
```
ppenna, line 7: I wish we could do it in a better way, but I am not sure if it is possible.

src/utils/nanvixd/src/logging.rs
```diff
@@ -25,7 +25,7 @@ use ::std::sync::Once;
 pub fn initialize() {
     static INIT_LOG: Once = Once::new();
     INIT_LOG.call_once(|| {
-        Logger::try_with_env()
+        Logger::try_with_env_or_str("error")
```
ppenna: Is this intentional?

src/utils/nanvixd/src/cache/mod.rs
```diff
@@ -113,149 +74,109 @@ impl SandboxCacheInner {
...
-    ///
-    pub async fn try_cleanup(&self) {
-        // Sleep for the keep alive timeout to avoid lock contention.
-        // NOTE: we sleep before locking the cache to avoid blocking other threads.
-        time::sleep(self.keep_alive_timeout).await;
-        // Lock the cache and try to cleanup expired sandboxes.
-        let mut locked_cache: LockedSandboxCache = self.inner.lock().await;
-        locked_cache.try_cleanup().await;
+        if let Some(user_vm) = self.user_vm_instances.remove(tag) {
+            if !Arc::strong_count(&user_vm) == 1 {
```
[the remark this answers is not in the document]
ppenna: @csegarragonz please check the correct semantic here.

src/utils/nanvixd/src/http.rs
```diff
@@ -117,104 +93,57 @@ impl HttpClient {
...
-                let reason: String = "failed to convert request length".to_string();
-                error!("serve(): {reason}");
-                anyhow::bail!(reason)
-            },
+        let control_plane_sockaddr: String =
+            config::control_plane_sockaddr_builder(&tmp_directory, tag.tenant_id());
+        let gateway_sockaddr: String =
+            config::gateway_sockaddr_builder(&tmp_directory, tag.tenant_id(), tag.app_name(), &tag.sandbox_id()[0..4]);
+        let user_vm_sockaddr: String =
+            config::user_vm_sockaddr_builder(&tmp_directory, tag.tenant_id(), tag.app_name(), &tag.sandbox_id()[0..4]);
```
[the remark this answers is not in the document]
ppenna: @csegarragonz may be worth checking.

src/utils/nanvixd/src/sandbox/microvm.rs
```diff
@@ -27,58 +31,59 @@ pub struct Microvm(Option<Child>);
...
     fn drop(&mut self) {
         if let Some(mut child) = self.0.take() {
-            tokio::spawn(async move {
-                debug!("killing microvm (child={:?})", child.id());
-                if let Err(e) = child.kill().await {
-                    error!("failed to kill microvm (child={:?}, Arror={:?})", child.id(), e);
-                }
+            match child.id() {
+                Some(pid) => {
+                    if let Err(e) = kill(Pid::from_raw(pid as i32), Signal::SIGINT) {
```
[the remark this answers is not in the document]
ppenna: I believe we have safe conversions for that.

src/utils/nanvixd/src/sandbox/linuxd.rs
```diff
@@ -17,40 +33,82 @@ use ::tokio::process::{
...
 impl Drop for LinuxDaemon {
     fn drop(&mut self) {
-        if let Some(mut child) = self.0.take() {
-            tokio::spawn(async move {
-                debug!("killing linux daemon");
-                if let Err(e) = child.kill().await {
-                    error!("failed to kill linux daemon ({e:?})");
+        match self.child.id() {
+            Some(pid) => {
+                if let Err(e) = kill(Pid::from_raw(pid as i32), Signal::SIGINT) {
```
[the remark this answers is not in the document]
ppenna: I believe we have safe conversions for that.

Makefile
ppenna: The way you handled it is right.

src/utils/nanvixd/src/logging.rs
csegarragonz: Yes, it got squashed with all the other changes, but doing this ensures that ERROR messages are printed if you don't run with any `RUST_LOG` env. var set.

Otherwise, even if the code fails you will see no error messages. If you want this to be the behaviour I can revert the change.

ppenna: Oh I see... I liked it, but we may want to fix that in the other projects (ie microvm and linuxd). Can you create a good first issue for that?

src/utils/nanvixd/src/http.rs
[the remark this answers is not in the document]
csegarragonz: The tests were failing because we were trying to create UNIX socket addresses with a name that was too long (combination of the new working directory, `$(pwd)` instead of `/tmp`, and the use of full UUIDs).

I instead choose to use only the first four bytes as this should give us enough variability (`59^4`, [0-9a-zA-Z] * 4), but we can always extend it.

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: Good point. However, once we move to mio-based asynchrnous sockets (my next PR) all sockets will become asynchronous, so maybe no need for the issue.

I will work on a standalone PR right after this one to switch the underlying sockets in `syscomm` to use `mio`.

src/utils/nanvixd/src/logging.rs
csegarragonz: Done: https://github.com/nanvix/nanvix/issues/704

Do you want me to undo these changes and apply them all in a separate PR?

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: Good catch. Seems unnecesarily risky. I have moved the send to after the match statement, once we have release the lock.

src/utils/nanvixd/src/logging.rs
ppenna: I can take that

scripts/test-nanvixd.sh
```diff
@@ -1,50 +1,30 @@
...
-    sudo /usr/bin/pkill -e -INT -P $parent_pid
-}
+mkdir -p ${LOGS_DIR}
 
 # Run nanvixd.
-NANVIXD_STDOUT_FILE_NAME="${LOGS_DIR}/nanvixd-stdout_$(date "+%Y_%m_%d_%H_%M").log"
-NANVIXD_STDERR_FILE_NAME="${LOGS_DIR}/nanvixd-stderr_$(date "+%Y_%m_%d_%H_%M").log"
 CONSOLE_FILE_NAME="${LOGS_DIR}/kernel_$(date "+%Y_%m_%d_%H_%M").log"
-RUST_LOG=trace timeout -s SIGINT --preserve-status --foreground ${TIMEOUT} \
+timeout -s SIGINT --preserve-status --foreground ${TIMEOUT} \
```
ppenna: We are no longer running with trace enabled, which impacts debuggability in the test pipeline. Is this intentional?

csegarragonz: It was not intentional. Thanks for pointing it out.

src/benchmarks/echo-c/main.c
```diff
@@ -28,19 +28,19 @@ int main(void)
...
-    if (n > 0) {
-        fwrite(buffer, 1, n, stdout);
-        fflush(stdout);
+        if (n > 0) {
+            fwrite(buffer, 1, n, stdout);
+            fflush(stdout);
+        }
     }
 
     return 0;
```
ppenna: We dont need two variables `n` and `nread` anymore. Could you change the benchmark to use just one?

src/benchmarks/echo-cpp/main.cpp
ppenna, line 1: See my previous comment.

ppenna, line 1: See my previous comment.

src/benchmarks/echo-wasm-rust/src/main.rs
ppenna, line 1: See my previous comment.

ppenna (dismissed):
Thanks for working on this change :)

scripts/test-nanvixd.sh
```diff
@@ -70,32 +50,40 @@ if ! ss -tln | grep -q ":${NANVIXD_PORT_NUMBER} "; then
 fi
 
 # Run a client.
-CURL_STDOUT_FILE_NAME="${LOGS_DIR}/curl-stdout_$(date "+%Y_%m_%d_%H_%M").log"
-curl \
+NEW_JSON=$(jq -n \
```
ppenna, line 53: `jq` is not installed in every Ubuntu system. Please add it to `ubutu.sh`

ppenna (changes requested):
I found the root cause why `pipeline.sh` fails locally.

scripts/test-nanvixd.sh
csegarragonz: Done! Thanks for catching this.

PR #701, opened 2025-07-23
[microvm] Add states and transitions to MicroVM snapshotting protocol

LucasGdosR:
# What the PR aims to do
This sets up the protocol for snapshots. All the states are traversed, and all channels send and receive messages. It still needs to build a connection to a client that sends the snapshot commands, and it doesn't actually pause and resume the MicroVM. It's just setting up the state transitions with a mocked up implementation.

# How it's done
`IoThread::run()` now has a different behavior depending on the state in the protocol. This method has a few TODOs related to the control connection to the Linux daemon, which isn't set up yet.

`IoThread::*_receive_from_outside_control()` are mocked methods. They simulate how a client might send commands to advance the protocol. The `blocking` version does communicate with the VMM, so the channels between the `IoThread` and the `Vmm` are tested by this method. `Vmm::handle_command()` is mocked on the other end to make this test possible.

The rest of the methods are self explanatory by their names.

# What I'd like feedback on
1. As is, this PR does not break anything. However, it changes the tail-latency in `IoThread::run()`. Instead of just polling all channels, it keeps transitioning between states, and only polls them in the `Running` state. Is this ok?

    When the protocol is fully implemented with the missing connections, then the `IoThread` will poll most of the time, since it will be in the `Running` state most of the time. This performance degradation wouldn't be permanent.

    The alternative I see is having dead code for the transitions. I believe the temporary slowdown is preferrable.

2. Are logging messages necessary during state transitions or when sending / receiving messages through channels?

3. I split the changes over three semantic commits. I hope they're clear. Should I squash them when it's time to merge?

NOTE: all TODOs are a staying part of this PR. I plan on working on them on the following PRs.

### DIFF
--- a/src/microvm/src/vmm/microvm/io.rs
+++ b/src/microvm/src/vmm/microvm/io.rs
@@ -195,17 +195,18 @@ impl IoThread {
     ///
     /// # Returns
     ///
-    /// Upon success, the received message is returned. Otherwise, an error is returned.
+    /// Upon success, the received message is pushed into the `incoming` queue, and `true` is returned.
+    /// Otherwise, if it would block, `false` is returned. Otherwise, an error is returned.
     ///
-    fn try_receive_from_gateway(&mut self) -> Result<()> {
+    fn try_receive_from_gateway(&mut self) -> Result<bool> {
         match self.gateway.try_receive() {
             Ok(message) => {
                 self.incoming.push_back(message);
-                Ok(())
+                Ok(true)
             },
             Err(e) => {
                 if e.kind() == ErrorKind::WouldBlock {
-                    Ok(())
+                    Ok(false)
                 } else {
                     let reason: String =
                         format!("failed to receive message from the gateway (error={e:?})");
@@ -223,9 +224,10 @@ impl IoThread {
     ///
     /// # Returns
     ///
-    /// Upon success, the received message is returned. Otherwise, an error is returned.
+    /// Upon success, the received message is pushed into the `outgoing` queue, and `true`is returned.
+    /// Otherwise, if the channel is empty, `false` is returned. Otherwise, an error is returned.
     ///
-    fn try_receive_from_microvm(&mut self) -> Result<()> {
+    fn try_receive_from_microvm(&mut self) -> Result<bool> {
         match self.microvm_rx.try_recv() {
             Ok(mut message) => {
                 profiler::timestamp_message!(
@@ -234,9 +236,9 @@ impl IoThread {
                         + std::mem::offset_of!(syscall::unistd::message::WriteRequest, buffer)
                 );
                 self.outgoing.push_back(message);
-                Ok(())
+                Ok(true)
             },
-            Err(TryRecvError::Empty) => Ok(()),
+            Err(TryRecvError::Empty) => Ok(false),
             Err(TryRecvError::Disconnected) => {
                 let reason: String = "the microvm has disconnected".to_string();
                 // When the guest finishes , the vCPU thread will disconnect from this thread. This
--- a/src/microvm/src/vmm/microvm/io.rs
+++ b/src/microvm/src/vmm/microvm/io.rs
@@ -5,7 +5,10 @@
 // Imports
 //==================================================================================================
 
-use crate::Gateway;
+use crate::{
+    Gateway,
+    vmm::ControlCommand,
+};
 use ::anyhow::Result;
 use ::std::{
     collections::VecDeque,
@@ -42,12 +45,10 @@ pub struct IoThread {
     incoming: VecDeque<Message>,
     /// Queue of outgoing messages.
     outgoing: VecDeque<Message>,
-    /// State in the snapshotting protocol.
-    _state: OrchestratorState,
     /// Command sender to the VMM.
-    _control_input_tx: Sender<ControlCommand>,
+    _vmm_control_tx: Sender<ControlCommand>,
     /// Response receiver from the VMM.
-    _control_output_rx: Receiver<ControlCommandResponse>,
+    vmm_control_rx: Receiver<ControlCommandResponse>,
     // TODO: channels to an outside issuer of snapshot commands and to linuxd.
 }
 
@@ -58,30 +59,15 @@ pub struct IoThread {
 ///
 /// # Description
 ///
-/// States relating to snapshots functionality. Snapshots may be loaded at PreBoot, and created at Paused.
-/// TODO: add `Running`, `Pausing`, `PausingAndOutputFlushed`, and `Paused` states.
-///
-enum OrchestratorState {
-    PreBoot,
-}
-
-///
-/// # Description
-///
-/// Control plane commands.
-/// TODO:
-/// Add commands relating to snapshots: `StartMicroVM`, `LoadAndRun`, `ResumeMicroVM`, `PauseMicroVM`, `PauseAndCreateSnapshot`, `LinuxDaemonFlushed`, `CreateSnapshot`, `LoadSnapshot`.
-///
-pub enum ControlCommand {}
-
-///
-/// # Description
-///
 /// Control plane command responses.
-/// TODO:
-/// Add `MicroVmPaused` response.
 ///
-pub enum ControlCommandResponse {}
+#[derive(PartialEq)]
+pub enum ControlCommandResponse {
+    MicroVmPaused,
+    SnapshotCreated,
+    FlushOutput,
+    FlushInput,
+}
 
 //==================================================================================================
 // Implementations
@@ -98,8 +84,8 @@ impl IoThread {
     /// - `gateway`: Connection to gateway.
     /// - `microvm_rx`: MicroVM receiver.
     /// - `microvm_tx`: MicroVM sender.
-    /// - `control_input_tx`: Command sender.
-    /// - `control_output_rx`: Response receiver.
+    /// - `vmm_control_tx`: Command sender.
+    /// - `vmm_control_rx`: Response receiver.
     ///
     /// # Returns
     ///
@@ -109,17 +95,12 @@ impl IoThread {
         gateway: Gateway,
         microvm_rx: Receiver<Message>,
         microvm_tx: Sender<Message>,
-        control_input_tx: Sender<ControlCommand>,
-        control_output_rx: Receiver<ControlCommandResponse>,
+        vmm_control_tx: Sender<ControlCommand>,
+        vmm_control_rx: Receiver<ControlCommandResponse>,
     ) -> JoinHandle<Result<()>> {
         thread::spawn(move || {
-            let mut io_thread: IoThread = IoThread::new(
-                gateway,
-                microvm_rx,
-                microvm_tx,
-                control_input_tx,
-                control_output_rx,
-            )?;
+            let mut io_thread: IoThread =
+                IoThread::new(gateway, microvm_rx, microvm_tx, vmm_control_tx, vmm_control_rx)?;
             io_thread.run()?;
             Ok(())
         })
@@ -135,8 +116,8 @@ impl IoThread {
     /// - `gateway`: Connection to gateway.
     /// - `microvm_rx`: MicroVM receiver.
     /// - `microvm_tx`: MicroVM sender.
-    /// - `control_input_tx`: Command sender.
-    /// - `control_output_rx`: Response receiver.
+    /// - `vmm_control_tx`: Command sender.
+    /// - `vmm_control_rx`: Response receiver.
     ///
     /// # Returns
     ///
@@ -146,8 +127,8 @@ impl IoThread {
         gateway: Gateway,
         microvm_rx: Receiver<Message>,
         microvm_tx: Sender<Message>,
-        control_input_tx: Sender<ControlCommand>,
-        control_output_rx: Receiver<ControlCommandResponse>,
+        vmm_control_tx: Sender<ControlCommand>,
+        vmm_control_rx: Receiver<ControlCommandResponse>,
     ) -> Result<Self> {
         Ok(Self {
             gateway,
@@ -155,36 +136,27 @@ impl IoThread {
             microvm_tx,
             incoming: VecDeque::new(),
             outgoing: VecDeque::new(),
-            _state: OrchestratorState::PreBoot,
-            _control_input_tx: control_input_tx,
-            _control_output_rx: control_output_rx,
+            _vmm_control_tx: vmm_control_tx,
+            vmm_control_rx,
         })
     }
 
     ///
     /// # Description
     ///
-    /// Runs the I/O thread.
+    /// Runs the I/O thread according to the state in the snapshotting protocol state machine.
     ///
     /// # Returns
     ///
     /// Upon success, empty is returned. Otherwise, an error is returned instead.
     ///
     fn run(&mut self) -> Result<()> {
-        let mut round: usize = 0;
-
-        // Cycle through actions to avoid starvation.
         loop {
-            if round % 4 == 0 {
-                self.try_receive_from_microvm()?;
-            } else if round % 4 == 1 {
-                self.try_send_to_gateway()?;
-            } else if round % 4 == 2 {
-                self.try_receive_from_gateway()?;
-            } else {
-                self.try_send_to_microvm()?;
-            }
-            round += 1;
+            self.try_receive_from_microvm()?;
+            self.try_send_to_gateway()?;
+            self.try_receive_from_gateway()?;
+            self.try_send_to_microvm()?;
+            self.try_receive_from_vmm_control()?;
         }
     }
 
@@ -309,5 +281,64 @@ impl IoThread {
             },
             None => Ok(()),
         }
+    }
+
+    ///
+    /// # Description
+    ///
+    /// Attempts to receive a response from the VMM.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, empty is returned. Otherwise, an error is returned.
+    ///
+    fn try_receive_from_vmm_control(&mut self) -> Result<()> {
+        match self.vmm_control_rx.try_recv() {
+            Ok(response) => match response {
+                ControlCommandResponse::FlushOutput => self.flush_microvm_output(),
+                ControlCommandResponse::FlushInput => self.flush_linuxd_input(),
+                _ => Ok(()), // TODO: forward to whoever is interested. This requires having control channels.
+            },
+            Err(TryRecvError::Empty) => Ok(()),
+            Err(TryRecvError::Disconnected) => {
+                let reason: String = "the vmm has disconnected".to_string();
+                // When the guest finishes , the vCPU thread will disconnect from this thread. This
+                // situation is normal and should not create an error log.
+                anyhow::bail!(reason)
+            },
+        }
+    }
+
+    /// # Description
+    ///
+    /// Attempts to flush all outstanding output from the MicroVM to the Linux daemon.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, empty is returned. Otherwise, an error is returned.
+    ///
+    fn flush_microvm_output(&mut self) -> Result<()> {
+        while self.try_receive_from_microvm()? {
+            // Keep looping until `microvm_rx` is empty, which breaks the loop.
+        }
+        while !self.outgoing.is_empty() {
+            self.try_send_to_gateway()?;
+        }
+        Ok(())
+    }
+
+    /// # Description
+    ///
+    /// Attempts to flush all outstanding input from the the Linux daemon to the MicroVM.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, empty is returned. Otherwise, an error is returned.
+    ///
+    fn flush_linuxd_input(&mut self) -> Result<()> {
+        while self.try_receive_from_gateway()? {
+            // Keep looping until receiving from the gateway would block, which breaks the loop.
+        }
+        Ok(())
     }
 }
--- a/src/microvm/src/vmm/microvm/mod.rs
+++ b/src/microvm/src/vmm/microvm/mod.rs
@@ -29,7 +29,6 @@ use crate::{
     Gateway,
     vmm::microvm::{
         io::{
-            ControlCommand,
             ControlCommandResponse,
             IoThread,
         },
@@ -55,11 +54,19 @@ use ::std::{
         },
     },
     thread::JoinHandle,
+    time::Instant,
 };
 use ::sys::ipc::{
     Message,
     MessageType,
 };
+
+//==================================================================================================
+// Constants
+//==================================================================================================
+
+// This value was chosen so it catches issues without polluting the logs with too many warnings.
+const TIMEOUT_WARNING_INTERVAL_IN_MS: usize = 10;
 
 //==================================================================================================
 // Structure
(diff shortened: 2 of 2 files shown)


ppenna:
> 1. As is, this PR does not break anything. However, it changes the tail-latency in `IoThread::run()`. Instead of just polling all channels, it keeps transitioning between states, and only polls them in the `Running` state. Is this ok?
>    When the protocol is fully implemented with the missing connections, then the `IoThread` will poll most of the time, since it will be in the `Running` state most of the time. This performance degradation wouldn't be permanent.
>    The alternative I see is having dead code for the transitions. I believe the temporary slowdown is preferrable.

I dont get why the performance regression is something transient. I would expect that the final solution would have at least the same performance that you are now observing. Having that said, how much is the regression that we are observing? I would say to open this PR for review to have official numbers on our CI infrastructure.

ppenna:
> Are logging messages necessary during state transitions or when sending / receiving messages through channels?

Logging is to help on debugging. You should place trace messages to help navigate execution, warning messages when you encounter an error but can move on and error messages when a fatal error occurs.

ppenna:
> Split the changes over three semantic commits. I hope they're clear. Should I squash them when it's time to merge?

Please add to each commit a full description on what it does.

When merging I will squash and combine all commit messages into one.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -155,36 +167,71 @@ impl IoThread {
...
+                    self.try_send_to_microvm()?;
+                    self.try_receive_from_outside_control()?;
+                },
+                // In `Pausing` state, pause the MicroVM and flush all outstanding output.
+                // Flushing the output means it doesn't need to be saved in snapshots.
+                OrchestratorState::Pausing => {
+                    self.control_input_tx.send(ControlCommand::PauseMicroVm)?;
+                    while self.try_receive_from_control_output()?
+                        != ControlCommandResponse::MicroVmPaused
+                    {
```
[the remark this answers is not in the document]
LucasGdosR: It is supposed to keep running while receiving Empty. It will eventually receive the expected `MicroVmPaused` response.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -155,36 +167,71 @@ impl IoThread {
...
+                // ensuring any outstanding messages are either buffered in `incoming`, or are in `linuxd`.
+                OrchestratorState::PausingAndOutputFlushed => {
+                    // TODO:
+                    // Try to receive `LinuxDaemonFlushed` from linuxd to advance the snapshotting protocol.
+                    // Also try to receive data from linuxd while the `LinuxDaemonFlushed` response doesn't arrive.
+                    // Should be something like the while-loop from the `Pausing` state.
+                    // This is a TODO because there's no control communication between linuxd and the `IoThread` yet.
+
+                    // Flush the incoming messages into the buffer.
+                    while self.try_receive_from_gateway()? {}
```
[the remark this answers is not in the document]
LucasGdosR: The current implementation already breaks when receiving `Ok(false)`. The `?` operator extracts the bool.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -308,4 +357,80 @@ impl IoThread {
...
+    /// # Description
+    ///
+    /// Attempts to flush all outstanding output from the MicroVM to the Linux daemon.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, empty is returned. Otherwise, an error is returned.
+    ///
+    fn flush_microvm_output(&mut self) -> Result<()> {
+        while self.try_receive_from_microvm()? {}
```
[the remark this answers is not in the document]
LucasGdosR: The current implementation already breaks when receiving `Ok(false)`. The `?` operator extracts the bool.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -308,4 +357,80 @@ impl IoThread {
...
+        }
+        Ok(())
+    }
+
+    // Placeholder function simulating the protocol for the `Paused` state.
+    fn blocking_receive_from_outside_control(&mut self) -> Result<()> {
+        // TODO:
+        // Create snapshot OR transition to `Running`. Currently does both to check the codepaths.
+        self.control_input_tx.send(ControlCommand::CreateSnapshot)?;
+        while self.try_receive_from_control_output()? != ControlCommandResponse::SnapshotCreated {}
```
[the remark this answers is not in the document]
LucasGdosR: It should keep running when it receives an `Empty` response. Eventually the `SnapshotCreated` response will come.

src/microvm/src/vmm/microvm/io.rs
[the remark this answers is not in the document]
ppenna: Yes it does, but I would slightly agree that we should rewrite this to be more explicit about what we are doing. You can alternatively have a comment in the loop body.

src/microvm/src/vmm/microvm/io.rs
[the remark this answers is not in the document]
ppenna: Same comment as the previous one.

src/microvm/src/vmm/microvm/io.rs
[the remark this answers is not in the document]
ppenna: If everything goes well, eventually the response will come. Anyways, the timeout technique is important to have. Set it to X ms and start logging warning messages every X ms have passed and you have not received the expected the response. The orchestrator will decide what to do.

src/microvm/src/vmm/microvm/io.rs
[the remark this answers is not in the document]
ppenna: See my previous response.

LucasGdosR:
> I dont get why the performance regression is something transient. I would expect that the final solution would have at least the same performance that you are now observing. Having that said, how much is the regression that we are observing? I would say to open this PR for review to have official numbers on our CI infrastructure.

This is why I expect the performance regression to be transient:
1. (Baseline) Before this PR, `IoThread::run()` just polled for messages and sent them;
2. (Unneeded state transitions) With this PR, there's a bunch of state transitions between polling for messages. The states transition automatically, even if no one requested the VMM to pause;
3. (State transitions on demand) After the feature is fully implemented, the states will only transition when a request for pausing arrives, so we'll be back at just polling most of the time.

I have opened the PR for review. Let's see what comes up :)

csegarragonz:
@LucasGdosR You can use the `warm-start-vmm` benchmark to measure the raw performance of traversing the VMM. You can use it with:

```bash
./bin/nanvix-bench.elf -benchmark warm-start-vmm -hwloc <path/to/hwloc.json>
```

you may need to first configre a `hwloc.json` file appropriate to your machine topology to be able to get comparable results between `dev` and your branch. When #703 lands this benchmark will also be run as part of the CI.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -155,36 +168,87 @@ impl IoThread {
...
     /// # Returns
     ///
     /// Upon success, empty is returned. Otherwise, an error is returned instead.
     ///
     fn run(&mut self) -> Result<()> {
-        let mut round: usize = 0;
-
-        // Cycle through actions to avoid starvation.
         loop {
-            if round % 4 == 0 {
```
csegarragonz, line 178: Like in the previous PR, I am not sure that the IO thread is the right place to implement this state machine logic. To me, the state machine you describe is for the VM (monitored by the VMM) not the IO thread.

The IO thread should be a relatively dumb component that just manages the IO of the VMM, nothing else. What you call `OrchestratorState` should be the `VmmState` and should live in the VMM, not in the IO thread.

More precisely, the IO thread should always try and communicate between linuxd and the VMM. If the VMM needs to snapshot, then it can let linuxd know by sending the corresponding state transition (which will flow through the IO thread).

Apologies if I should have brought this in the previous PR, but I don't like how the IO thread is accumulating more and more state and complexity, when it is not the right component to manage it.

LucasGdosR: Thanks for the review. There's no problem whatsoever in bringing up an issue with a previous implementation. In general, I'm glad to improve previous decisions if they later turn out to be subpar. I appreciate it.

I think in this case, unlike in the previous PR, there're good reasons to keep state and logic in the IO thread. At this time, there's only state in the IO thread, but later I'd distribute more state. Each vCPU (at this time, just the one) would have its state, and linuxd would have state for each connection with a VMM. vCPUs would be either in `Running` or `Paused`, and linuxd connections would be in `Running`, `Flushing`, or `Paused`.

My idea for consistency in snapshots is this protocol:
1. IO `Running`: receive pause message (I think nanvixd would send this message, is that it?) -> IO `Pausing`
2. Broadcast the message to linuxd and the VMM (linuxd connection `Running` -> `Flushing`)
3. All vCPUs must be paused (vCPUs `Running` -> `Paused`)
4. When all vCPUs are paused, all output from the VMM should be flushed to linuxd. This ensures that the `outgoing` queue of messages is empty, and therefore does not need to be saved in snapshots (IO `Pausing` -> `Pausing (output flushed)`
5. When all output has been flushed, the VMM shouldn't try to send anything to linuxd any longer, since all vCPUs are paused, and nothing will get produced. It should tell linuxd that it is done flushing
6. When linuxd acknowledges that no more output is coming its way, it should get in a consistent state for the snapshot. This consistent state is achieved by avoiding being in the middle of sending a message. It should finish sending a whole message, if it is in the middle of sending one, and then stop sending more input messages, instead sending a control message that it is done sending messages (linuxd connection `Flushing` -> `Paused`)
7. When the VMM acknowledges that no more input is coming its way, it shouldn't try to receive anything from linuxd any longer (IO `Pausing (output flushed)` -> `Paused`)
8. Now the VMM is in a consistent state for creating snapshots. It should save the MicroVM's state (memory and vCPU registers), as well as the `incoming` queue, which is input linuxd has already sent, but wasn't processed by the MicroVM yet. linuxd is responsible for persisting whatever input messages it has buffered for the VMM, as well as other data that lives in it (file system, for example, I think)
9. When a resume message arrives (again, from nanvixd, right?), the IO thread broadcasts it to the VMM and linuxd (IO `Paused` -> `Running`; vCPUs `Paused` -> `Running`; linuxd connection `Paused` -> `Running`)
10. At this point, it is safe to receive a new pause message arbitrarily close to the previous resume message, so there are no intermediate states between `Paused` and `Running`.

Since there are steps in the protocol when input / output is flushed, and there are states where the data communication between the VMM and linuxd is paused, the IO thread's behavior should change based on the state in the protocol.

Since linuxd and the VMM must coordinate to get into a consistent state, the best place for them to coordinate is in their interface, the IO thread. Having a different thread coordinate them adds another layer of asynchronous communication between the IO thread and the VMM, adding complexity to the protocol. It is simpler to coordinate the thread that's doing the communication with linuxd directly.

The IO thread needs to broadcast messages to the VMM and linuxd. I believe this is simpler than sending a message to the VMM, and later having the VMM send another message to the IO thread to forward it to linuxd.

I'm sorry for the long comment, but I think it was time for us to sync on the whole protocol draft. If you find any issues with it, I'd be happy to improve it.

ppenna: > At this time, there's only state in the IO thread, but later I'd distribute more state.

@LucasGdosR could you open an issue about the snapshot features, and share there all details and design decisions that you are considering? I think that this will help @csegarragonz and I to better understand the big picture. Even for me, the information is spread.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -155,36 +168,87 @@ impl IoThread {
...
+                },
+                // In `Running` state,
+                // execute the application and check for control commands outside the VMM.
+                OrchestratorState::Running => {
+                    self.try_receive_from_microvm()?;
+                    self.try_send_to_gateway()?;
+                    self.try_receive_from_gateway()?;
+                    self.try_send_to_microvm()?;
+                    self.try_receive_from_outside_control()?;
+                },
```
ppenna: We are having a major regression on warm-start microbenchmark. I wonder if the additional poll for the control channel is introducing this overhead. Have you evaluated this? Could you check if we multiplex control and data messages over the same channel this behavior is eliminated?

src/microvm/src/vmm/microvm/io.rs
csegarragonz: @LucasGdosR thanks for the detailed write-up!

Agree with @ppenna that would be very helpful!

src/microvm/src/vmm/microvm/io.rs
LucasGdosR: I'm pretty sure I know what's going on. I'll make `try_receive_from_outside_control()` transition less often while in the `Running` state. In a couple hours I'll get off work and test this hypothesis.

LucasGdosR:
I couldn't test this before pushing, as there's some WIP in linuxd tests and nanvixd tests, from what I could gather. However, I was able to test most of it before rebasing. It was working then. It should be fine. Still, I'd rather wait for a definite result once those tests are reestablished.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -308,4 +283,85 @@ impl IoThread {
...
+            3 => ControlCommand::LinuxDaemonFlushed,
+            4 => ControlCommand::CreateSnapshot,
+            5 => ControlCommand::ResumeMicroVm,
+            6 => ControlCommand::PauseAndCreateSnapshot,
+            7 => ControlCommand::LinuxDaemonFlushed,
+            8 => ControlCommand::LoadSnapshotAndRun,
+            _ => ControlCommand::ResumeMicroVm,
+        })?;
+        Ok(())
+    }
```
ppenna: Why is this required?

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -371,9 +401,75 @@ impl Vmm {
...
+                    }
+                    Ok(())
+                },
+                ControlCommand::LinuxDaemonFlushed => {
+                    // NOTE: this will be unreachable once the communication is fully implemented
+                    // `LinuxDaemonFlushed` should only be sent in the middle of `pause_protocol`.
+                    // In fact, it should already be unreachable, but it cannot be tested ATM.
+                    Ok(())
+                },
+            },
```
ppenna: If either `pause_protocol()`, `create_snapshot()` or `resume_microvm()`, how is gracefully termination handled?

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -383,4 +479,82 @@ impl Vmm {
...
+    ///
+    /// # Returns
+    ///
+    /// Upon success, empty is returned. Otherwise, an error is returned instead.
+    ///
+    fn receive_linux_daemon_flushed(&mut self) -> Result<()> {
+        // Check how long it takes to receive a response
+        let start: Instant = Instant::now();
+        const INTERVAL: u128 = 10;
+        let mut counter: u128 = 1;
```
ppenna: Why do we need a `u128`for these variables? Could you improve the name for `INTERVAL` variable? it is very generic and people dont read code :) Also, could you add a comment explaining how its value was chosen and also move it to file scope?

ppenna (commented):
I believe that after my minor comments are addressed, we will be good to go :)

src/microvm/src/vmm/microvm/io.rs
LucasGdosR: There are two reasons for this:
1. So all new codepaths are executed;
2. So there's no performance regression.

If there was no mocking code, there might be bugs that weren't exposed, because they weren't executed.

`ResumeMicroVm` was chosen as the default command for `c > 8` because it has the least overhead, and is closer to the expected behavior once the nanvixd control channel is implemented.

src/microvm/src/vmm/microvm/mod.rs
LucasGdosR: Graceful termination isn't handled, but I think I see your point. Those leaf functions might return errors, and the place to handle them is here, not upwards the call stack, so instead of the `?` operator, I should use a match-statement with specific handling.

For now, all those functions are placeholders for future implementation. I think it's best to write error handling code once we know what might arise. For now, I think adding an error message might be best, so errors are easier to trace to the source code

src/microvm/src/vmm/microvm/mod.rs
LucasGdosR: I used a u128 so it's the same type as returned by `Instant::elapsed().as_millis()`, so I can directly compare it with `INTERVAL * counter`. I could cast it to another type such as `usize` and use both variables as `usize` as well.

I agree INTERVAL is generic. I believe that was an appropriate name for a function scope. If you believe it should be in file scope to potentially be reused for other timeout functionality, then I agree a more descriptive name is warranted. I'll move it, rename it, and add a comment about its value.

src/microvm/src/vmm/microvm/io.rs
ppenna: > So there's no performance regression.

Why is this fixing performance regression?

> So all new codepaths are executed;

Please add a comment to this function saying that and then mentioning that it will be removed.

src/microvm/src/vmm/microvm/mod.rs
ppenna: >For now, all those functions are placeholders for future implementation. I think it's best to write error handling code once we know what might arise. For now, I think adding an error message might be best, so errors are easier to trace to the source code

Please catch all errors, log error messages and rethrow the error.

src/microvm/src/vmm/microvm/mod.rs
ppenna: > so I can directly compare it with INTERVAL * counter

Why counter needs to be a `u128`? Isn't `usize` (64-bit) enough? Why not? Please document.

ppenna: > I believe that was an appropriate name for a function scope

It should be file scope. We use this pattern across the project source tree.

src/microvm/src/vmm/microvm/io.rs
LucasGdosR: > Why is this fixing performance regression?

Messages sent through `vmm_control_tx` are then handled by `Vmm::handle_command()`. Handling `ResumeMicroVm` while in the `Running` state is just skipping a predictable branch. Other commands lead to more significant overhead, such as `PauseAndCreateSnapshot`, which calls `Vmm::pause_protocol()` and `Vmm::create_snapshot()`. The pausing protocol involves flushing output to linuxd, and flushing inputs from linuxd, as well as communication / coordination between the main thread and the IO thread.

I expect that looping through handling different messages in every iteration would lead to performance regression, as it did in the previous implementation that used that strategy. Handling each kind of message just once should not add significant overhead.

> Please add a comment to this function saying that and then mentioning that it will be removed.

I will add a more verbose comment. My intention with:
`// Placeholder to simulate a client.`
was precisely to convey that idea.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -146,45 +127,37 @@ impl IoThread {
...
-            } else {
-                self.try_send_to_microvm()?;
-            }
-            round += 1;
+            self.try_receive_from_microvm()?;
+            self.try_send_to_gateway()?;
+            self.try_receive_from_vmm_control()?;
+            self.try_send_to_microvm()?;
+            self.try_receive_from_gateway()?;
+            self.mock_receive_from_outside_control_and_send_to_vmm_control()?;
```
ppenna: I am still not convinced why this is required. If no control commands are sent at all, there should be no impact in performance. If no commands are sent at all, this statement is not required.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -308,4 +283,87 @@ impl IoThread {
...
+
+    // Placeholder to simulate a client.
+    // This function will be removed.
+    // It only exists so all codepaths related to the snapshotting protocol get executed for debugging purposes.
+    fn mock_receive_from_outside_control_and_send_to_vmm_control(&mut self) -> Result<()> {
+        let count: usize = unsafe {
+            static mut COUNT: usize = 0;
+            COUNT += 1;
+            COUNT
+        };
```
ppenna: There is no reason for having unsafe code here. Please use atomic variables.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -308,4 +283,87 @@ impl IoThread {
...
+        self.vmm_control_tx.send(match count {
+            c if c > 8 => ControlCommand::ResumeMicroVm, // Keep running
```
ppenna: What I understand is that you require that this function executes at least 8 times in the loop, but it is not clear at all why. Why 8? Why To start the protocol we need to send `PauseAndCreateSnapshot`?

Please improve comments

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -308,4 +283,87 @@ impl IoThread {
...
+                // When the guest finishes , the vCPU thread will disconnect from this thread. This
+                // situation is normal and should not create an error log.
+                anyhow::bail!(reason)
+            },
+        }
+    }
+
+    // Placeholder to simulate a client.
+    // This function will be removed.
+    // It only exists so all codepaths related to the snapshotting protocol get executed for debugging purposes.
```
ppenna: > `... get executed for debugging purposes.`

Is it debugging or performance? I am lost.

ppenna (changes requested):

src/microvm/src/vmm/microvm/io.rs
ppenna: https://github.com/nanvix/nanvix/issues/710

src/microvm/src/vmm/microvm/io.rs
ppenna: I am not convinced. I continued this discussion in the next review.

LucasGdosR:
@ppenna I think it's best to take all current comments in a single message, as they're all related.

I think that what I intended to do would perhaps be best done via unit / integration tests, rather than the way I did it. That's probably the root of the confusion. `mock_receive_from_outside_control_and_send_to_vmm_control` is essentially a testing function. Mixing its execution with actual program execution doesn't make sense.

This function is going to be thrown away soon enough, once there's a control channel that issues commands. I do not believe it is worth putting too much effort into it. In fact, as you point out, it is not required, and could be erased right now with only dead code warnings.

I will try to make the comments clear enough so it's possible to fully understand everything.

ppenna:
> I think that what I intended to do would perhaps be best done via unit / integration tests, rather than the way I did it. That's probably the root of the confusion. mock_receive_from_outside_control_and_send_to_vmm_control is essentially a testing function. Mixing its execution with actual program execution doesn't make sense.

> This function is going to be thrown away soon enough, once there's a control channel that issues commands. I do not believe it is worth putting too much effort into it. In fact, as you point out, it is not required, and could be erased right now with only dead code warnings.

> I will try to make the comments clear enough so it's possible to fully understand everything.

What is confusing me:
1. The `mock_receive_from_outside_control_and_send_to_vmm_control` function does not need to be there, as you just stated it is for debugging purposes.
2. I said: "let's remove it"
3. You argued that it was introduced to improve performance.

Maybe I miss interpreted (3)? If so, could you explain what was causing the early performance regressions that we observed? If not, I insist "let's remove `mock_receive_from_outside_control_and_send_to_vmm_control` from the mainline code".

LucasGdosR:
> What is confusing me:
> 
> 1. The `mock_receive_from_outside_control_and_send_to_vmm_control` function does not need to be there, as you just stated it is for debugging purposes.
> 2. I said: "let's remove it"
> 3. You argued that it was introduced to improve performance.
> 
> Maybe I miss interpreted (3)? If so, could you explain what was causing the early performance regressions that we observed? If not, I insist "let's remove `mock_receive_from_outside_control_and_send_to_vmm_control` from the mainline code".

I understand now. I have removed it.

Early performance regressions happened because of tests running in the mainline code. Using the count and the weird loop would reduce the overhead of running tests in the mainline code, but removing the tests altogether removes that issue entirely.

src/microvm/src/vmm/microvm/io.rs
```diff
@@ -146,45 +127,36 @@ impl IoThread {
...
-                self.try_receive_from_gateway()?;
-            } else {
-                self.try_send_to_microvm()?;
-            }
-            round += 1;
+            self.try_receive_from_microvm()?;
+            self.try_send_to_gateway()?;
+            self.try_receive_from_vmm_control()?;
+            self.try_send_to_microvm()?;
+            self.try_receive_from_gateway()?;
```
ppenna: Why changing this order? Can `self.try_receive_from_vmm_control()?` be the last?

LucasGdosR: It can be last, no problem. With 3 channels (let's call them A, B, C), I thought it was aesthetic and perhaps balanced to order them:

RA

SB

RC

SA

RB

SC

Or, in the general case, go A..N twice alternating between receiving and sending.

ppenna: Please move it to the last. I understand your motivation, but you should not touch code that is unrelated to your change.

ppenna (approved):

PR #705, opened 2025-07-24
[libs] Move Sycomm To Use Non-Blocking Metal-IO Sockets

csegarragonz:
This switch will allow us to comfortably poll different SocketStreams in a non-blocking, asynchronous, fashion.


ppenna:
Superseded by #735

PR #735, opened 2025-08-05
[daemons] Make Linuxd Support Multiple User VM Connections

### LINKED ISSUE #674: [daemons] `linuxd` exits after running a single `microvm` application
## Description

`linuxd` is supposed to persist after a `microvm` connects to it and exits. It is not persisting, but exiting.

## How to Reproduce

1. In one terminal, run `./bin/linuxd.elf -user-vm-bind-addr 127.0.0.1:1234`
2. In another terminal, run `sudo -E ./bin/microvm.elf -kernel bin/kernel.elf -initrd bin/hello-c.elf -gateway 127.0.0.1:1234
3. See that `linuxd.elf` has exited

## Screenshots

Image

## Expected Behavior

`linuxd` should keep accepting new connections.

csegarragonz:
In this PR I implement support in linuxd to manage multiple user VM connections.  There are two places where we poll for incoming connections:
1. The main linuxd thread, that polls for incoming user VM connections.
2. A new `GatewayPollThread` that polls incoming client connections to connect to a user VM's stdin/stdout.

In order to poll over a set of SocketStreams, we need to use non-blocking Metal IO sockets, so in the first commit I move the `syscomm` lib to use those. Unfortunately, in some scenarios we need to perform a blocking read. For example, when we are reading data from the gateway we want to block the caller thread until data is available, but we also want to block in a way that we can be interrupted if needed. I therefore introduce support to perform blocking reads on non-blocking sockets by keeping a per-socket poll structure. This is incompatible with adding the socket to an external poll structure.

Another hairy aspect is that we require the user VM and the gateway to connect roughly at the same time, and no other user VMs or gateways to connect before the match has been made. I have ideas on how to refine this protocol, but they belong in a follow-up PR.

Closes #674 (fixes)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,6 +80,7 @@ kvm-ioctls = "0.23.0"
 libc = "0.2.172"
 log = "0.4.27"
 microvm = { path = "src/microvm", default-features = false }
+mio = { version = "1", default-features = false }
 nanvixd = { path = "src/utils/nanvixd", default-features = false }
 num_enum = { version = "0.7.3", default-features = false }
 nvx = { path = "src/libs/nvx", default-features = false }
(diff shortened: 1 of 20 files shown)


csegarragonz:
Closed in favour of #814

PR #741, opened 2025-08-06
[daemons] Support Graceful Interruption of Linuxd Worker Threads

csegarragonz:
In this PR I implement support to gracefully shutdown the worker threads in linuxd associated to a user VM execution.

A worker thread can be in three different states:
1. Running
2. Blocked on the main `recv` loop in `handle_message` (trying to receive from a `mpsc::Channel`
3. Blocked on a system call

To address 1 and 2, I introduce a shutdown message that can be queued to the main work queue. To address 3 I install a SIGUSR1 handler in worker threads, and send them a signal when it is time to exit. Note that a signal interrupt will not wake a thread blocked on a `mpsc::Channel`.

To handle the signal interrupt and propagate it all the way up the stack such that the thread knows it needs to exit, I need to modify every system call to handle an `EINTR` return value, and propagate it encapsulated in an `Error`.

To encapsulate the logic of starting/stopping an interruptible worker thread, including the handler installation, I abstract all the logic corresponding to each worker thread from `src/daemons/linuxd/src/linuxd/mod.rs` to `src/daemons/linuxd/src/worker_thread.rs`.

### DIFF
--- a/src/daemons/linuxd/Cargo.toml
+++ b/src/daemons/linuxd/Cargo.toml
@@ -16,6 +16,7 @@ config = { workspace = true }
 flexi_logger = { workspace = true }
 libc = { workspace = true }
 log = { workspace = true }
+nix = { workspace = true, features = ["pthread", "signal"] }
 profiler = { workspace = true }
 signal-hook = { workspace = true }
 sys = { workspace = true, features = ["std"] }
--- a/src/daemons/linuxd/Cargo.toml
+++ b/src/daemons/linuxd/Cargo.toml
@@ -16,7 +16,6 @@ config = { workspace = true }
 flexi_logger = { workspace = true }
 libc = { workspace = true }
 log = { workspace = true }
-nix = { workspace = true, features = ["pthread", "signal"] }
 profiler = { workspace = true }
 signal-hook = { workspace = true }
 sys = { workspace = true, features = ["std"] }
--- a/src/daemons/linuxd/src/error.rs
+++ b/src/daemons/linuxd/src/error.rs
@@ -1,0 +1,42 @@
+// Copyright(c) The Maintainers of Nanvix.
+// Licensed under the MIT License.
+
+//==================================================================================================
+// Configuration
+//==================================================================================================
+
+#![deny(clippy::all)]
+
+//==================================================================================================
+// Imports
+//==================================================================================================
+
+use ::nix::libc;
+use ::std::io;
+use ::syscomm::SocketError;
+
+#[derive(Debug)]
+pub enum WorkerThreadError {
+    Error(io::Error),
+    Interrupted,
+}
+
+impl From<io::Error> for WorkerThreadError {
+    fn from(e: io::Error) -> Self {
+        if e.raw_os_error() == Some(libc::EINTR) {
+            WorkerThreadError::Interrupted
+        } else {
+            WorkerThreadError::Error(e)
+        }
+    }
+}
+
+impl From<SocketError> for WorkerThreadError {
+    fn from(e: SocketError) -> Self {
+        if e.raw_os_error() == Some(libc::EINTR) {
+            WorkerThreadError::Interrupted
+        } else {
+            WorkerThreadError::Error(e.into())
+        }
+    }
+}
--- a/src/daemons/linuxd/src/error.rs
+++ b/src/daemons/linuxd/src/error.rs
@@ -11,32 +11,14 @@
 // Imports
 //==================================================================================================
 
-use ::nix::libc;
-use ::std::io;
-use ::syscomm::SocketError;
-
 #[derive(Debug)]
 pub enum WorkerThreadError {
-    Error(io::Error),
+    Error(sys::error::Error),
     Interrupted,
 }
 
-impl From<io::Error> for WorkerThreadError {
-    fn from(e: io::Error) -> Self {
-        if e.raw_os_error() == Some(libc::EINTR) {
-            WorkerThreadError::Interrupted
-        } else {
-            WorkerThreadError::Error(e)
-        }
-    }
-}
-
-impl From<SocketError> for WorkerThreadError {
-    fn from(e: SocketError) -> Self {
-        if e.raw_os_error() == Some(libc::EINTR) {
-            WorkerThreadError::Interrupted
-        } else {
-            WorkerThreadError::Error(e.into())
-        }
+impl From<sys::error::Error> for WorkerThreadError {
+    fn from(err: sys::error::Error) -> Self {
+        WorkerThreadError::Error(err)
     }
 }
--- a/src/daemons/linuxd/src/linuxd/mod.rs
+++ b/src/daemons/linuxd/src/linuxd/mod.rs
@@ -13,7 +13,10 @@ mod assemble;
 
 use crate::{
     message::RequestAssembler,
-    venv::VirtualEnviromentDirectory,
+    venv::{
+        VenvCommand,
+        VirtualEnviromentDirectory,
+    },
     worker_thread::WorkerThreadHandle,
 };
 use ::anyhow::Result;
@@ -132,7 +135,8 @@ impl LinuxDaemon {
                                 response_buf))?;
                         }
                         Err(RecvError) => {
-                            info!("gateway STDIN channel disconnected");
+                            // This may happen during shutdown.
+                            debug!("gateway STDIN channel disconnected");
                             break Ok(());
                         }
                     }
@@ -154,9 +158,10 @@ impl LinuxDaemon {
                             // writting thread has already moved on.
                         }
                         Err(RecvError) => {
+                            // This may happen during shutdown.
                             let reason: String = "gateway STDOUT channel disconnected".to_string();
-                            error!("{}", reason);
-                            return Err(anyhow::anyhow!(reason));
+                            debug!("{}", reason);
+                            break Ok(());
                         }
                     }
                 }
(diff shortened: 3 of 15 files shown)


src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -238,9 +158,10 @@ impl LinuxDaemon {
                             // writting thread has already moved on.
                         }
                         Err(RecvError) => {
+                            // This may happen during shutdown.
                             let reason: String = "gateway STDOUT channel disconnected".to_string();
-                            error!("{}", reason);
-                            return Err(anyhow::anyhow!(reason));
+                            debug!("{}", reason);
```
ppenna: We can collapse these.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -263,7 +187,25 @@ impl LinuxDaemon {
...
+                        // To gracefully shutdown the thread, we enqueue a shutdown message to the
+                        // message channel. In case the thread is blocked on a system call, we also
+                        // send it an interrupt signal and handle EINTR accordingly.
+                        for worker_thread in worker_threads.drain(..) {
+                            trace!("sending interrupt to worker thread (thread_id={:?})", worker_thread.id);
+                            worker_thread.cmd_tx.send(VenvCommand::Shutdown)
+                                .map_err(|_| Error::new(ErrorCode::IoErr, "failed to send VenvCommand to queue"))?;
+                            worker_thread.stop()?;
+                            worker_thread.handle.join()
+                                .map_err(|_| Error::new(ErrorCode::IoErr, "failed to join worker thread"))?;
```
ppenna: Could we log errors before failing? This way we don't fail silently.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -263,7 +187,25 @@ impl LinuxDaemon {
...
+                        // 3. Blocked waiting for a new message from the channel
+                        //
+                        // To gracefully shutdown the thread, we enqueue a shutdown message to the
+                        // message channel. In case the thread is blocked on a system call, we also
+                        // send it an interrupt signal and handle EINTR accordingly.
+                        for worker_thread in worker_threads.drain(..) {
+                            trace!("sending interrupt to worker thread (thread_id={:?})", worker_thread.id);
+                            worker_thread.cmd_tx.send(VenvCommand::Shutdown)
+                                .map_err(|_| Error::new(ErrorCode::IoErr, "failed to send VenvCommand to queue"))?;
+                            worker_thread.stop()?;
```
ppenna: If we fail to send the stop signal, we want to log a warning message and keep trying to harvest the other worker threads.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -301,7 +243,12 @@ impl LinuxDaemon {
                         Err(error) => {
                             warn!("failed to join new virtual environment (error={error:?})");
                             let message: Message = crate::build_error(source, error.code);
-                            Self::send(uvm_stream.clone(), message).unwrap();
+                            uvm_stream
+                                .lock()
+                                .unwrap()
+                                .write_all(&message.to_bytes())
+                                .map_err(|_| Error::new(ErrorCode::IoErr, "failed to write to user VM stream"))?;
```
ppenna: Could you log an error before failing? This way we don't fail silently.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -314,13 +261,16 @@ impl LinuxDaemon {
                 let venv: Arc<Mutex<VirtualEnviromentDirectory>> = venv.clone();
                 let gw_stdin_tx = gw_stdin_tx.clone();
                 let gw_stdout_tx = gw_stdout_tx.clone();
-                let _ = std::thread::spawn(move || {
-                    Self::handle_message(channel_rx, uvm_stream, gw_stdin_tx, gw_stdout_tx, venv, assembler);
-                });
+                let assembler = assembler.clone();
+
+                // Spawn an interruptible thread to handle the message.
+                let worker_thread_handle =
```
ppenna: Could you add a type annotation here?

src/daemons/linuxd/src/error.rs
```diff
@@ -0,0 +1,49 @@
...
+        if e.raw_os_error() == Some(libc::EINTR) {
+            WorkerThreadError::Interrupted
+        } else {
+            WorkerThreadError::Error(e.into())
+        }
+    }
+}
+
+impl From<sys::error::Error> for WorkerThreadError {
+    // TODO: proper error-code propagation.
```
ppenna: Can you create an issue for this?

src/daemons/linuxd/src/unistd.rs
```diff
@@ -110,30 +119,38 @@ pub fn do_chdir(tid: ThreadIdentifier, request: ChangeDirectoryRequest) -> Vec<M
...
     match unsafe { libc::close(fd) } {
-        ret if ret == 0 => CloseResponse::build(tid, ret),
-        _ => crate::build_error(tid, ErrorCode::InvalidArgument),
+        ret if ret == 0 => Ok(CloseResponse::build(tid, ret)),
+        _ => {
+            // Check if the thread has been interrupted.
+            let err = std::io::Error::last_os_error();
+            if err.raw_os_error() == Some(libc::EINTR) {
+                return Err(WorkerThreadError::Interrupted);
+            }
```
ppenna: Why is this special (ie different from the others)?

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -340,408 +290,7 @@ impl LinuxDaemon {
         Ok(())
     }
 
-    fn handle_message(
```
ppenna, line 343: Could we have done that in a different PR ?

src/daemons/linuxd/Cargo.toml
```diff
@@ -16,6 +16,7 @@ config = { workspace = true }
 flexi_logger = { workspace = true }
 libc = { workspace = true }
 log = { workspace = true }
+nix = { workspace = true, features = ["pthread", "signal"] }
```
ppenna: Couldn't we achieve what we want with just libc? I would rather have either libc or nix.

ppenna (commented):
Thanks for working on this. I left some coments.

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: You are right, we should not be failing when we are cleaning-up.

src/daemons/linuxd/src/unistd.rs
csegarragonz: No real reason. Making them the same.

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: Yes, the only changes to the code here are catching the `WorkerThreadError::Interrupted` errors and the `VenvCommand::Shutdown` message.

Should I split them?

src/daemons/linuxd/Cargo.toml
csegarragonz: Yes we can! Done.

src/daemons/linuxd/src/linuxd/mod.rs
ppenna: I leave that to you... I basically assume that you just moved the code and have not introduced any changes?

csegarragonz: Code is unchanged with two exceptions:
1. The return value of `handle_{special,short,long}_message` has changed from `Message` to `Result<Message, WorkerThreadError>`.
2. The main match clause in `handle_message` now matches on the return value of the methods above, and if they are `WorkerThreadError::Interrupted` breaks from the loop.

~10-20 lines of code changed.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -263,7 +186,44 @@ impl LinuxDaemon {
...
+                                }
+                            }
+                            match worker_thread.handle.join() {
+                                Ok(_) => {},
+                                Err(e) => {
+                                    error!("error joining worker thread (thread_id={e:?})");
+                                    continue;
+                                }
+                            }
+                        }
```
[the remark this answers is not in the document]
csegarragonz: Stop and join are called on a different thread.

src/daemons/linuxd/src/worker_thread.rs
```diff
@@ -0,0 +1,920 @@
...
+    /// Stop a worker-thread by setting the shutdown flag and sending an interrupt.
+    pub fn stop(&self) -> Result<(), Error> {
+        let raw_tid = self.pthread_id.load(Ordering::Relaxed);
+
+        if raw_tid == 0 {
+            let reason = "trying to stop thread with tid 0";
+            error!("{reason}");
+            return Err(Error::new(ErrorCode::InvalidArgument, reason));
+        }
+
```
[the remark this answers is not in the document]
csegarragonz: The function description is outdated, I update it.

src/daemons/linuxd/src/worker_thread.rs
```diff
@@ -0,0 +1,920 @@
...
+                    sigemptyset(&mut set);
+                    set
+                },
+                // No SA_RESTART so that syscall will return EINTR.
+                sa_flags: 0,
+                sa_restorer: None,
+            };
+
+            sigaction(INTERRUPT_SIGNAL, &sig_action, ptr::null_mut());
+        }
```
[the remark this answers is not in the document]
csegarragonz: Good catch, I will check the return value.

ppenna (approved):

PR #743, opened 2025-08-07
[utils] Make Nanvixd Send Shutdown Messages To Linuxd

csegarragonz:
In this PRI implement the Nanvixd support for graceful linuxd shutdown. When Nanvixd receives a SIGINT, it will send a `Shutdown` message over the control-plane socket that it shares with linuxd.

This PR misses the linuxd handling to be complete, which is blocked until we can poll over different sockets in linuxd. I have left FIXMEs in the bits of code that will change once this is implemented.

### DIFF
--- a/src/libs/syscomm/src/lib.rs
+++ b/src/libs/syscomm/src/lib.rs
@@ -78,7 +78,7 @@ impl ::std::str::FromStr for SocketType {
 
 impl SocketListener {
     /// Accepts a connection on a socket.
-    pub fn accept(&self) -> Result<SocketStream> {
+    pub fn accept(&self) -> Result<SocketStream, SocketError> {
         match self {
             SocketListener::Tcp(listener) => {
                 let (stream, _sockaddr): (TcpStream, std::net::SocketAddr) = listener.accept()?;
@@ -245,6 +245,12 @@ impl fmt::Display for SocketError {
 impl From<SocketError> for io::Error {
     fn from(err: SocketError) -> Self {
         err.error
+    }
+}
+
+impl From<std::io::Error> for SocketError {
+    fn from(error: std::io::Error) -> Self {
+        SocketError::new(error)
     }
 }
 
--- a/src/utils/nanvix-bench/src/main.rs
+++ b/src/utils/nanvix-bench/src/main.rs
@@ -269,6 +269,17 @@ impl Benchmark {
             if ret_code < 0 {
                 error!("error sending SIGINT to nano VM: {}", std::io::Error::last_os_error());
             }
+
+            if let Some(nanvixd) = self.nanvixd.as_mut() {
+                match nanvixd.wait() {
+                    Ok(exit_status) => {
+                        if !exit_status.success() {
+                            error!("nanvixd returned with non-zero exit status: {:?}", exit_status.code());
+                        }
+                    },
+                    Err(e) => error!("error waiting for nanvixd: {e:?}"),
+                }
+            }
         }
     }
 
@@ -378,7 +389,6 @@ impl Benchmark {
         println!("p99: {} us", latencies[(self.iterations as f32 * 0.99) as usize]);
 
         print!("Cleaning up...");
-        self.cleanup();
         println!("done!");
 
         Ok(())
@@ -744,9 +754,6 @@ async fn main() -> Result<()> {
         Ok(_) => {},
         Err(e) => error!("error running benchmark: {e:?}"),
     }
-
-    // Additional clean-up in case of errors in the benchmarks.
-    benchmark.cleanup();
 
     Ok(())
 }
--- a/src/utils/nanvixd/src/cache/mod.rs
+++ b/src/utils/nanvixd/src/cache/mod.rs
@@ -6,9 +6,9 @@
 //==================================================================================================
 
 use crate::sandbox::{
+    config::SandboxConfig,
     linuxd::LinuxDaemon,
     microvm::Microvm,
-    config::SandboxConfig,
     tag::SandboxTag,
 };
 use ::anyhow::Result;
@@ -180,5 +180,31 @@ impl SandboxCache{
         self.sandbox_index.remove(user_vm_id);
 
         Ok(())
+    }
+
+    ///
+    /// # Description
+    ///
+    /// Stop all instances in the cache.
+    ///
+    /// # Returns
+    ///
+    /// On success empty is returned. On failure an error is returned instead.
+    ///
+    pub async fn cleanup(&mut self) {
+        // TODO: for the time being only linuxd instances support being gracefully shutdown. User
+        // VMs are missing a control-plane socket.
+        for (tenant_id, linuxd_instance) in self.linuxd_instances.iter_mut() {
+            if let Some(linuxd_instance_mut) = Arc::get_mut(linuxd_instance) {
+                if let Err(e) = linuxd_instance_mut.shutdown().await {
+                    // FIXME: convert to error once linuxd supports a graceful shutdown.
+                    debug!("error cleaning-up linuxd instance for tenant {tenant_id}: {e:?}");
+                } else {
+                    debug!("cleaned-up linuxd instance for tenant {tenant_id}");
+                }
+            } else {
+                error!("error cleaning-up linuxd instance for tenant {tenant_id}: instance not found");
+            }
+        }
     }
 }
--- a/src/utils/nanvixd/src/control_plane.rs
+++ b/src/utils/nanvixd/src/control_plane.rs
@@ -1,0 +1,43 @@
+// Copyright(c) The Maintainers of Nanvix.
+// Licensed under the MIT License.
+
+// This file contains the messages used in the control-plane API between
+// nanvixd and linuxd/user VM.
+
+use ::anyhow::Result;
+use ::num_enum::{
+    IntoPrimitive,
+    TryFromPrimitive,
+};
+use ::std::io::{
+    Error,
+    ErrorKind,
+};
+use ::syscomm::{
+    SocketError,
+    SocketStream,
+};
+
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, IntoPrimitive, TryFromPrimitive)]
+pub enum Command {
+    Shutdown,
+}
+
+// This function is actually used by linuxd, consuming nanvixd as a library. It is never used from
+// the main nanvixd binary, hence why we need to add this annotation to silent clippy.
+#[allow(dead_code)]
+pub fn read_command(stream: &mut SocketStream) -> Result<Command, SocketError> {
+    let mut buf: [u8; 1] = [0u8; 1];
+    stream.read_exact(&mut buf)?;
+
+    Command::try_from(buf[0]).map_err(|_| {
+        Error::new(ErrorKind::InvalidData, "error parsing control-plane command".to_string()).into()
+    })
+}
+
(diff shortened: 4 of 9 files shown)


src/utils/nanvixd/src/sandbox/linuxd.rs
```diff
@@ -101,32 +92,24 @@ impl LinuxDaemon {
...
+        match control_plane::send_command(&mut self.control_plane_stream, control_plane::Command::Shutdown) {
             Ok(_) => {},
-            Err(ref e) if e.kind() == ErrorKind::NotFound => {},
-            Err(e) => error!("error removing UNIX socket (path={}, error={e:?})", self.linuxd_sockaddr.clone()),
-        }
+            Err(_) => {
+                // FIXME: this send_command is expected to fail until support is implemented in
+                // linuxd.
+            }
+        };
```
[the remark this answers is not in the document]
csegarragonz: This is deliberate to indicate that it is now error-ing, but it should not.

src/libs/syscomm/src/lib.rs
```diff
@@ -78,7 +78,7 @@ impl ::std::str::FromStr for SocketType {
 
 impl SocketListener {
     /// Accepts a connection on a socket.
-    pub fn accept(&self) -> Result<SocketStream> {
+    pub fn accept(&self) -> Result<SocketStream, SocketError> {
```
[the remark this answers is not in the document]
csegarragonz: Thanks. By making it return a SockerError we can preserve the underlying error kind, in case we have an `EWOULDBLOCK`.

src/utils/nanvix-bench/src/main.rs
```diff
@@ -273,7 +273,12 @@ impl Benchmark {
             debug!("Sending SIGINT to nanvixd");
             match kill(Pid::from_raw(self.nanvixd.as_mut().unwrap().id() as i32), Signal::SIGINT) {
                 Ok(_) => {},
-                Err(e) => error!("error sending SIGINT to nano VM: {e:?}"),
+                Err(e) => error!("error sending SIGINT to nanvixd {e:?}"),
+            }
+
+            match self.nanvixd.as_mut().unwrap().wait() {
+                Ok(_) => {},
```
ppenna: Could you explicitly match with `()`?

src/utils/nanvixd/src/control_plane.rs
```diff
@@ -0,0 +1,40 @@
...
+};
+use ::syscomm::{
+    SocketError,
+    SocketStream,
+};
+
+#[repr(u8)]
+#[derive(Clone, Copy, Debug)]
+pub enum Command {
+    Shutdown = 1,
```
ppenna: Can we have a global constant for this and use it throughout the file?

src/utils/nanvixd/src/message.rs
```diff
@@ -1,5 +1,7 @@
 // Copyright(c) The Maintainers of Nanvix.
 // Licensed under the MIT License.
+//
+// This file contains the messages used in the HTTP API between end-clients and nanvixd.
```
ppenna: Could you detach this from the license header?

```
// Copyright(c) The Maintainers of Nanvix.
// Licensed under the MIT License.

/// This file contains the messages used in the HTTP API between end-clients and nanvixd.
```

src/utils/nanvixd/src/sandbox/linuxd.rs
```diff
@@ -35,13 +27,11 @@ use ::tokio::process::{
 //==================================================================================================
 
 pub struct LinuxDaemon {
+    // TODO: once linuxd reacts to the shutdown command, we will use this member to wait for the
+    // underlying process to finish.
```
ppenna: Ok.

src/utils/nanvixd/src/sandbox/linuxd.rs
```diff
@@ -101,32 +92,24 @@ impl LinuxDaemon {
...
+    /// Send a shutdown message to linuxd so that it can clean-up its internal resources.
+    pub async fn shutdown(&mut self) -> Result<()> {
+        match control_plane::send_command(&mut self.control_plane_stream, control_plane::Command::Shutdown) {
             Ok(_) => {},
-            Err(ref e) if e.kind() == ErrorKind::NotFound => {},
-            Err(e) => error!("error removing UNIX socket (path={}, error={e:?})", self.linuxd_sockaddr.clone()),
-        }
+            Err(_) => {
+                // FIXME: this send_command is expected to fail until support is implemented in
+                // linuxd.
```
ppenna, line 102: Could you log an error message here for now and keep this comment?

src/utils/nanvix-bench/src/main.rs
csegarragonz: Sorry what do you mean by this?

src/utils/nanvixd/src/control_plane.rs
csegarragonz: I am not sure I understand. `Shutdown` is one of the possible commands we will have. I am giving it a number because to avoid implementing an expensive protocul, I used `[repr(u8)]` to have rust automatically (de-)serialize it for us.

src/utils/nanvix-bench/src/main.rs
ppenna: Could you explicitly match the value returned by `wait`? Even if it is `()`.

src/utils/nanvixd/src/control_plane.rs
ppenna: Yes, still you are hardcoding the constant value when handling the conversion. You could alternatively use num_enum crate.

csegarragonz: Makes sense. Done.

src/utils/nanvix-bench/src/main.rs
```diff
@@ -269,6 +269,15 @@ impl Benchmark {
             if ret_code < 0 {
                 error!("error sending SIGINT to nano VM: {}", std::io::Error::last_os_error());
             }
+
```
ppenna, line 272: (nitpick) Could you avoid `unwrap`s in this function by using `if let Some( .. ) = ` pattern?

src/utils/nanvixd/src/cache/mod.rs
```diff
@@ -181,4 +181,24 @@ impl SandboxCache{
 
         Ok(())
     }
+
+    ///
+    /// # Description
+    ///
+    /// Stop all instances in the cache.
+    ///
+    pub async fn cleanup(&mut self) -> Result<()> {
```
ppenna: Please document return code (we need to start making the source tree AI friendly):

```text
On success empty is returned. Or failure an error is returned instead.
```

src/utils/nanvixd/src/cache/mod.rs
```diff
@@ -181,4 +181,24 @@ impl SandboxCache{
...
+        // TODO: for the time being only linuxd instances support being gracefully shutdown. User
+        // VMs are missing a control-plane socket.
+        for (tenant_id, linuxd_instance) in self.linuxd_instances.iter_mut() {
+            if let Some(linuxd_instance_mut) = Arc::get_mut(linuxd_instance) {
+                linuxd_instance_mut.shutdown().await?;
```
ppenna: We probably want to have thit in the drop method. We can address that in a later PR. If you agree could you create an issue for that?

src/utils/nanvixd/src/sandbox/linuxd.rs
```diff
@@ -94,14 +89,23 @@ impl LinuxDaemon {
...
-    fn drop(&mut self) {
+    /// Send a shutdown message to linuxd so that it can clean-up its internal resources.
+    pub async fn shutdown(&mut self) -> Result<()> {
+        match control_plane::send_command(&mut self.control_plane_stream, control_plane::Command::Shutdown) {
+            Ok(_) => {},
+            Err(_) => {
+                // FIXME: this send_command is expected to fail until support is implemented in
+                // linuxd.
+                error!("failed to send shutdown command to linuxd");
+            }
```
ppenna: Can we catch the error and print it here?

src/utils/nanvixd/src/sandbox/linuxd.rs
```diff
@@ -94,14 +89,23 @@ impl LinuxDaemon {
...
         })
     }
-}
 
-impl Drop for LinuxDaemon {
-    fn drop(&mut self) {
+    /// Send a shutdown message to linuxd so that it can clean-up its internal resources.
+    pub async fn shutdown(&mut self) -> Result<()> {
+        match control_plane::send_command(&mut self.control_plane_stream, control_plane::Command::Shutdown) {
+            Ok(_) => {},
```
ppenna: Can we capture the return value here to enforce strong typing?

src/utils/nanvixd/src/sandbox/linuxd.rs
```diff
@@ -113,11 +117,9 @@ impl Drop for LinuxDaemon {
...
-        // Clean-up the socket files left behind by linuxd.
-        match fs::remove_file(self.linuxd_sockaddr.clone()) {
-            Ok(_) => {},
-            Err(ref e) if e.kind() == ErrorKind::NotFound => {},
-            Err(e) => error!("error removing UNIX socket (path={}, error={e:?})", self.linuxd_sockaddr.clone()),
-        }
+        // Wait for linuxd instance to finish.
+        self.child.wait().await?;
+
+        Ok(())
```
ppenna: Couldn't it be simply:

```
  ...
  // Wait for linuxd instance to finish.
  self.child.wait().await
}

```

src/utils/nanvixd/src/control_plane.rs
```diff
@@ -0,0 +1,43 @@
...
+#[derive(Clone, Copy, Debug, IntoPrimitive, TryFromPrimitive)]
+pub enum Command {
+    Shutdown,
+}
+
+// This function is actually used by linuxd, consuming nanvixd as a library. It is never used from
+// the main nanvixd binary, hence why we need to add this annotation to silent clippy.
+#[allow(dead_code)]
+pub fn read_command(stream: &mut SocketStream) -> Result<Command, SocketError> {
+    let mut buf = [0u8; 1];
```
ppenna: Please add a type annotation here.

src/utils/nanvixd/src/message.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright(c) The Maintainers of Nanvix.
 // Licensed under the MIT License.
 
+// This file contains the messages used in the HTTP API between end-clients and nanvixd.
```
ppenna: Is this meant to be a module-level documentation?

ppenna (commented):
Left some minor comments.

src/utils/nanvixd/src/cache/mod.rs
csegarragonz: You suggest triggering the `cleanup` method inside the cache's drop?

I thought about this but the issue is that, given that cleanup involves sending network requests to other components, we need to `await` on them, and we cannot do that easily when implementing the `drop` trait (which is not async).

src/utils/nanvixd/src/sandbox/linuxd.rs
csegarragonz: Good catch. I realized that we actually don't want to return a `Result` because we want to catch, and tolerate, errors during clean-up.

PR #746, opened 2025-08-07
[daemons] Fix Worker Thread Panic On VM Shutdown

csegarragonz:
There is a race condition when a user VM is shut down, but the linuxd worker threads are doing work but are _not_ blocked on a system call.

In this case, when threads are done doing work and try to write to the user VM stream, they will find the stream is closed.

We should handle this case gracefully.

This was causing linuxd to panic in some of the benchmarks, but we were not catching that because, until we have a graceful shutdown, we don't really have a way to know linuxd's exit status.


ppenna (approved):
Makes sense.

PR #774, opened 2025-08-12
[libs] Move Syscomm To Metal IO Sockets

csegarragonz:
In this PR I move the syscomm library to use metal IO sockets. 

These sockets are non-blocking by default, and the only way to block on them is via a `poll` structure. Given that in some places we rely on blocking reads, I implement a `BlockingSocketStream` object that wraps around the blocking logic. Thus, `SocketStream` only implements non-blocking `try_read/try_read_exact`, whereas `BlockingSocketStream` only implements blocking calls `read/read_exact`. We can get a `BlockingSocketStream` from a `SocketStream` by calling `SocketStream::set_blocking()`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,6 +80,7 @@ kvm-ioctls = "0.23.0"
 libc = "0.2.172"
 log = "0.4.27"
 microvm = { path = "src/microvm", default-features = false }
+mio = { version = "1.0.4", default-features = false }
 nanvixd = { path = "src/utils/nanvixd", default-features = false }
 num_enum = { version = "0.7.3", default-features = false }
 nvx = { path = "src/libs/nvx", default-features = false }
--- a/src/daemons/linuxd/src/linuxd/mod.rs
+++ b/src/daemons/linuxd/src/linuxd/mod.rs
@@ -183,10 +183,23 @@ impl LinuxDaemon {
                         Err(error) => {
                             warn!("failed to join new virtual environment (error={error:?})");
                             let message: Message = crate::build_error(source, error.code);
-                            uvm_handle
-                                .get_user_vm_stream()
-                                .lock()
-                                .unwrap()
+
+                            let uvm_stream: Arc<Mutex<(SocketStream, VecDeque<u8>)>> =
+                                uvm_handle.get_user_vm_stream();
+                            let mut guard: MutexGuard<'_, (SocketStream, VecDeque<u8>)> =
+                                match uvm_stream.lock() {
+                                    Ok(guard) => guard,
+                                    Err(e) => {
+                                        error!(
+                                            "error acquiring lock on user VM stream (error={e:?})"
+                                        );
+                                        continue;
+                                    },
+                                };
+                            let (locked_uvm_stream, _): &mut (SocketStream, VecDeque<u8>) =
+                                &mut guard;
+
+                            locked_uvm_stream
                                 .write_all(&message.to_bytes())
                                 .map_err(|_| {
                                     let reason = "failed to write to user VM stream";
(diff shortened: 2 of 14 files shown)


Cargo.toml
```diff
@@ -80,6 +80,7 @@ kvm-ioctls = "0.23.0"
 libc = "0.2.172"
 log = "0.4.27"
 microvm = { path = "src/microvm", default-features = false }
+mio = { version = "1", default-features = false }
```
ppenna: Is it possible to specify the full version number so we can have more reproducible builds?

src/libs/config/src/lib.rs
```diff
@@ -65,6 +65,13 @@ pub mod syscomm {
     /// Provides the length of the temporary bufer we use to read messages from the gateway.
     ///
     pub const GW_READ_BUFFER_LEN: usize = 4 * 1024;
+
+    ///
+    /// # Description
+    ///
+    /// Provides the maximum number of poll events we can buffer.
```
ppenna: Could you add the rationale for this value?

src/utils/nanvix-bench/src/main.rs
```diff
@@ -538,7 +550,20 @@ impl Benchmark {
...
             input_stream.write_all(&read_response.to_bytes())?;
-            input_stream.read_exact(&mut buf)?;
+
+            // Explicitly spin-loop when receiving to isolate the overheads of the VMM.
+            loop {
+                match input_stream.try_read_exact(&mut buf) {
+                    Ok(()) => break,
+                    Err(ref e) if e.kind() == std::io::ErrorKind::WouldBlock => continue,
+                    Err(e) => {
+                        return Err(anyhow::anyhow!(
```
ppenna: Please log an error before failing.

src/libs/syscomm/src/lib.rs
```diff
@@ -195,12 +324,132 @@ impl SocketStream {
...
+    ///
+    /// The number of bytes read into the buffer.
+    ///
+    pub fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        match self {
+            BlockingSocketStream::Tcp(stream, poll) => {
+                let mut events = Events::with_capacity(config::syscomm::MAX_NUM_POLL_EVENTS);
+
+                // Even after a poll wake-up the socket may still return WouldBlock.
+                loop {
```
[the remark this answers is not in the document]
csegarragonz: On a spurious wake-up we immediately poll again (and sleep).

src/libs/syscomm/src/lib.rs
```diff
@@ -167,16 +250,62 @@ impl SocketStream {
...
+    pub fn try_read_exact(&mut self, buf: &mut [u8]) -> Result<(), SocketError> {
+        let mut total_read = 0;
+
+        while total_read < buf.len() {
+            match self.try_read(&mut buf[total_read..]) {
+                Ok(0) => {
+                    return Err(io::Error::new(ErrorKind::UnexpectedEof, "connection closed").into())
+                },
+                Ok(n) => total_read += n,
+                Err(ref e) if e.kind() == ErrorKind::WouldBlock => {
```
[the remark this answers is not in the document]
ppenna: @csegarragonz I agree. this can be a problem, unless you carefully handle the `WOuldBlock` case.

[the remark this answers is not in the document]
csegarragonz: Makes sense. I have addressed this and made `try_read_exact` return the number of bytes it managed to read.

src/libs/syscomm/src/lib.rs
```diff
@@ -133,24 +181,59 @@ impl SocketStream {
...
+                        &mut stream,
+                        BLOCKING_THREAD_TOKEN,
+                        Interest::READABLE | Interest::WRITABLE,
+                    )
+                    .map_err(|e| {
+                        io::Error::other(format!("failed to register thread to poll (error={e:?})"))
+                    })?;
+
+                Ok(BlockingSocketStream::Unix(stream, poll_instance))
+            },
```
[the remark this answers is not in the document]
csegarragonz: This is a deliberate design choice, and is only used for sockets that need to be shared across worker threads where we cannot have a centralized polling structure.

src/utils/nanvix-bench/src/main.rs
```diff
@@ -513,7 +512,26 @@ impl Benchmark {
...
+                            break;
+                        }
+                    },
+                    Err(ref e) if e.kind() == std::io::ErrorKind::WouldBlock => continue,
+                    Err(e) => {
+                        return Err(anyhow::anyhow!(
+                            "error reading from input VMM stream (error={e:?})"
+                        ));
+                    },
+                }
```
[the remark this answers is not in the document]
csegarragonz: This is only used in a special benchmark that needs access to the underlying sockets.

src/microvm/src/gateway.rs
```diff
@@ -84,23 +89,34 @@ impl Gateway {
...
+
+            buf[..num_partial_read].copy_from_slice(&partial_bytes[..num_partial_read]);
+
+            // Clear partial read buffer.
+            self.partial_read_buffer.clear();
+            num_filled += num_partial_read;
+        }
+        // Post-condition: partial_read_buffer is empty.
+
+        match self.stream.try_read_exact(&mut buf) {
```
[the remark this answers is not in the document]
csegarragonz: damn, good catch copilot

src/libs/syscomm/src/lib.rs
```diff
@@ -167,17 +250,59 @@ impl SocketStream {
...
+    ///
+    /// The number of bytes read into the buffer.
+    ///
+    pub fn try_read_exact(&mut self, buf: &mut [u8]) -> Result<usize, SocketError> {
+        let mut total_read = 0;
+
+        while total_read < buf.len() {
+            match self.try_read(&mut buf[total_read..]) {
+                Ok(0) => {
+                    return Err(io::Error::new(ErrorKind::UnexpectedEof, "connection closed").into())
```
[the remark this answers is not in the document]
csegarragonz: I don't think this is true? It should return WouldBlock.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -189,10 +183,23 @@ impl LinuxDaemon {
...
+                            let uvm_stream: Arc<Mutex<(SocketStream, VecDeque<u8>)>> =
+                                uvm_handle.get_user_vm_stream();
+                            let mut guard: MutexGuard<'_, (SocketStream, VecDeque<u8>)> =
+                                match uvm_stream.lock() {
+                                    Ok(guard) => guard,
+                                    Err(e) => {
+                                        error!(
+                                            "error acquiring lock on user VM stream (error={e:?})"
+                                        );
+                                        continue;
```
ppenna, line 196: Failing to notify the user VM about this may have undefined behavior. If we fail to obtain the lock, it is likely that we had a contention, we could sleep and retry. If we retried for more than X times, than we should simply kill that VM (instead of letting it hanging). I understand that this is out of the scope of this PR. Could you create an issue and link it here?

csegarragonz: If I understand correctly, [`lock()`](https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.lock) will only return an error if the mutex is poisoned, which means that another thread panicked while holding the lock.

If this happens this means that something is going very wrong, so maybe we should just error-out? (It is unlikely that we will be able to clean this user VM anyway, neither).

src/daemons/linuxd/src/worker_thread.rs
```diff
@@ -842,44 +851,41 @@ impl WorkerThreadHandle {
...
-                            );
-                            break Ok(ReadResponse::eof(source));
+                            error!("gateway stream mutex poisoned (error={e:?})");
+                            return Ok(ReadResponse::eof(source));
                         },
                     };
+
+                // Read from the gateway thread.
+                let mut response_buf: [u8; ReadResponse::BUFFER_SIZE] =
+                    [0u8; ReadResponse::BUFFER_SIZE];
```
ppenna, line 865: Please use the EOF helper function.

src/libs/syscomm/src/lib.rs
```diff
@@ -19,16 +32,14 @@ use ::std::{
...
-        UnixListener,
-        UnixStream,
-    },
 };
 
+//==================================================================================================
+// Constants
+//==================================================================================================
+
+const BLOCKING_THREAD_TOKEN: Token = Token(0);
```
ppenna, line 42: Could you add a comment with a rationale for this constant here?

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: Created a tracking issue: https://github.com/nanvix/nanvix/issues/787

ppenna (approved):

PR #793, opened 2025-08-13
[daemons] Make Linuxd Poll Over A Single Connection + Control-Plane Socket

csegarragonz:
In this PR I move linuxd to poll instead of spin-loop when trying to receive a message from the user VM. In addition of being more resource efficient, this also makes it easy to poll, together with the user VM socket, over the control-plane socket, and process events from both in a unified manner.

The lines changed in this PR are very small. The diff is large because I had to indent a lot of the logic, and put it inside the right match arm.

### DIFF
--- a/src/daemons/linuxd/src/linuxd/mod.rs
+++ b/src/daemons/linuxd/src/linuxd/mod.rs
@@ -21,6 +21,13 @@ use crate::{
     worker_thread::WorkerThreadHandle,
 };
 use ::anyhow::Result;
+use ::mio::{
+    Events,
+    Interest,
+    Poll,
+    Token,
+};
+use ::nanvixd::control_plane;
 use ::std::{
     collections::VecDeque,
     io::ErrorKind,
@@ -49,7 +56,10 @@ use ::syscomm::SocketStream;
 // Constants
 //==================================================================================================
 
-const DEFAULT_CONNECTION_ID: usize = 0;
+/// We use ID 0 for the control-plane socket in the main poll structure.
+const CONTROL_PLANE_CONNECTION_ID: usize = 0;
+/// We use ID 1 for the user VM connection in the main poll structure.
+const DEFAULT_CONNECTION_ID: usize = 1;
 
 //==================================================================================================
 // Structures
@@ -57,6 +67,7 @@ const DEFAULT_CONNECTION_ID: usize = 0;
 
 pub struct LinuxDaemon {
     assembler: Arc<Mutex<RequestAssembler>>,
+    control_plane_stream: SocketStream,
     uvm_stream: SocketStream,
     gateway_stream: Option<SocketStream>,
     venv: Arc<Mutex<VirtualEnviromentDirectory>>,
@@ -68,18 +79,106 @@ pub struct LinuxDaemon {
 
 impl LinuxDaemon {
     pub fn init(
+        control_plane_stream: SocketStream,
         uvm_stream: SocketStream,
         gateway_stream: Option<SocketStream>,
     ) -> Result<Self, Error> {
         Ok(Self {
             assembler: Arc::new(Mutex::new(RequestAssembler::default())),
+            control_plane_stream,
             uvm_stream,
             gateway_stream,
             venv: Arc::new(Mutex::new(VirtualEnviromentDirectory::new())),
         })
     }
 
-    pub fn run(self) -> Result<(), Error> {
+    /// Helper method to close a connection to a user VM identified by the connection id. Closing
+    /// the connection also involves stopping all associated worker threads.
+    fn close_connection(
+        uvm_handle: UserVmHandle,
+        poll: &Poll,
+        worker_threads: Option<VecDeque<WorkerThreadHandle>>,
+    ) {
+        // De-register the user VM socket from the poll structure.
+        let user_vm_stream: Arc<Mutex<(SocketStream, VecDeque<u8>)>> =
+            uvm_handle.get_user_vm_stream();
+        match user_vm_stream.lock() {
+            Ok(mut guard) => {
+                let (locked_uvm_stream, _): &mut (SocketStream, VecDeque<u8>) = &mut guard;
+                if let Err(e) = poll.registry().deregister(locked_uvm_stream) {
+                    error!("failed to deregister user VM from poll (error={e:?})");
+                }
+            },
+            Err(e) => {
+                error!("error acquiring lock on user VM stream (error={e:?})");
+            },
+        };
+
+        // Send a shutdown message to all worker threads associated
+        // with this user VM.
+        if let Some(mut worker_threads) = worker_threads {
+            for worker_thread in worker_threads.drain(..) {
+                trace!("sending interrupt to worker thread (thread_id={:?})", worker_thread.id);
+
+                // Each worker thread may be in one of three states:
+                // 1. Running
+                // 2. Blocked on a system call
+                // 3. Blocked waiting for a new message from the channel
+                //
+                // To gracefully shutdown the thread, we enqueue a shutdown message to the
+                // message channel. In case the thread is blocked on a system call, we also
+                // send it an interrupt signal and handle EINTR accordingly. Note that a signal
+                // interrupt will not unblock a thread waiting on a queue, so we need both
+                // mechanisms.
+                //
+                // If any of the commands fail, continue trying to drain the remaining
+                // threads.
+                if let Err(e) = worker_thread.cmd_tx.send(VenvCommand::Shutdown) {
+                    error!(
+                        "error sending shutdown command to worker thread (thread_id={:?}, \
+                         error={e:?})",
+                        worker_thread.id
+                    );
+                }
+                if let Err(e) = worker_thread.stop() {
+                    error!(
+                        "error sending interrupt to worker thread (thread_id={:?}, error={e:?})",
+                        worker_thread.id
+                    );
+                }
+                if let Err(e) = worker_thread.handle.join() {
+                    error!(
+                        "error joining worker thread (thread_id={:?}, error={e:?})",
+                        worker_thread.id
+                    );
+                }
+            }
+        }
+    }
+
+    fn log_and_error(code: ErrorCode, msg: &'static str) -> Error {
+        error!("{msg}");
+        Error::new(code, msg)
+    }
+
+    pub fn run(mut self) -> Result<(), Error> {
+        const CONTROL_PLANE_TOKEN: Token = Token(CONTROL_PLANE_CONNECTION_ID);
+        const USER_VM_TOKEN: Token = Token(DEFAULT_CONNECTION_ID);
+
+        let mut events: Events = Events::with_capacity(config::syscomm::MAX_NUM_POLL_EVENTS);
+        let mut poll: Poll = Poll::new()
+            .map_err(|_| Self::log_and_error(ErrorCode::IoErr, "failed to create Poll"))?;
+        poll.registry()
+            .register(&mut self.control_plane_stream, CONTROL_PLANE_TOKEN, Interest::READABLE)
+            .map_err(|_| {
+                Self::log_and_error(ErrorCode::IoErr, "failed to register cotnrol-plane to poll")
+            })?;
+        poll.registry()
+            .register(&mut self.uvm_stream, USER_VM_TOKEN, Interest::READABLE)
+            .map_err(|_| {
+                Self::log_and_error(ErrorCode::IoErr, "failed to register user VM listener to poll")
+            })?;
+
         // Map keeping track of the worker threads associated the user VM.
         let mut worker_threads: VecDeque<WorkerThreadHandle> = VecDeque::new();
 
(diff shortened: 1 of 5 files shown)


src/utils/nanvixd/src/control_plane.rs
```diff
@@ -28,16 +27,19 @@ pub enum Command {
...
 // the main nanvixd binary, hence why we need to add this annotation to silent clippy.
 #[allow(dead_code)]
-pub fn read_command(stream: &mut BlockingSocketStream) -> Result<Command, SocketError> {
+pub fn try_read_command(stream: &mut SocketStream) -> Result<Command, SocketError> {
     let mut buf: [u8; 1] = [0u8; 1];
-    stream.read_exact(&mut buf)?;
+
+    // Try read exact returns the number of bytes written `n` with 0 < n <= buf.len(). In this case
+    // it is safe to ignore the return value because n can only ever be 1.
+    stream.try_read_exact(&mut buf)?;
```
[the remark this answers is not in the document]
csegarragonz: The function `try_read_command` behaves in the same way than the other `try_*` functions, potentially returning `WouldBlock`. The comment clarifies why it is safe to ignore the rerutn value of `try_read_exact`.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,272 @@ pub struct LinuxDaemon {
...
+        worker_threads: Option<VecDeque<WorkerThreadHandle>>,
+    ) {
+        // De-register the user VM socket from the poll structure.
+        let user_vm_stream: Arc<Mutex<(SocketStream, VecDeque<u8>)>> =
+            uvm_handle.get_user_vm_stream();
+        match user_vm_stream.lock() {
+            Ok(mut guard) => {
+                let (locked_uvm_stream, _): &mut (SocketStream, VecDeque<u8>) = &mut guard;
+                if let Err(e) = poll.registry().deregister(&mut *locked_uvm_stream) {
+                    error!("failed to de-revister user VM from poll (error={e:?})");
```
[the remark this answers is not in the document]
csegarragonz: The method close_connection is only used once per user VM connection, and the connection is dropped afterwards.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,272 @@ pub struct LinuxDaemon {
...
+                                );
+                            },
+                        };
+
+                        // Check if process is associated with a virtual environment.
+                        let (channel_tx, channel_rx): (
+                            Sender<VenvCommand>,
+                            Option<Receiver<VenvCommand>>,
+                        ) = {
+                            let mut venv: MutexGuard<'_, VirtualEnviromentDirectory> =
```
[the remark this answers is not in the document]
csegarragonz: This code has just been indented in this PR, and is not new. I did not want to modify any of the indented code (to avoid confusions), but I will leave this comment open to decide whether it makes sense to change it or not.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -49,16 +56,17 @@ use ::syscomm::SocketStream;
 // Constants
 //==================================================================================================
 
-const DEFAULT_CONNECTION_ID: usize = 0;
+const DEFAULT_CONNECTION_ID: usize = 1;
```
ppenna, line 62: Why did we have to change this?

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -49,16 +56,17 @@ use ::syscomm::SocketStream;
...
 //==================================================================================================
 // Structures
 //==================================================================================================
 
 pub struct LinuxDaemon {
-    assembler: Arc<Mutex<RequestAssembler>>,
+    control_plane_stream: SocketStream,
     uvm_stream: SocketStream,
     gateway_stream: Option<SocketStream>,
+    assembler: Arc<Mutex<RequestAssembler>>,
```
ppenna: Please move this member to its original position to keep diffs minimal.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,275 @@ pub struct LinuxDaemon {
...
+        control_plane_stream: SocketStream,
         uvm_stream: SocketStream,
         gateway_stream: Option<SocketStream>,
     ) -> Result<Self, Error> {
         Ok(Self {
-            assembler: Arc::new(Mutex::new(RequestAssembler::default())),
+            control_plane_stream,
             uvm_stream,
             gateway_stream,
+            assembler: Arc::new(Mutex::new(RequestAssembler::default())),
```
ppenna: Please move this line to its original position to keep diffs minimal.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,275 @@ pub struct LinuxDaemon {
...
             venv: Arc::new(Mutex::new(VirtualEnviromentDirectory::new())),
         })
     }
 
-    pub fn run(self) -> Result<(), Error> {
+    /// Helper method to close a connection to a user VM identified by the connection id. Closing
+    /// the connection also involes stopping all associated worker threads.
```
ppenna: Typo "involves"

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,275 @@ pub struct LinuxDaemon {
...
+        worker_threads: Option<VecDeque<WorkerThreadHandle>>,
+    ) {
+        // De-register the user VM socket from the poll structure.
+        let user_vm_stream: Arc<Mutex<(SocketStream, VecDeque<u8>)>> =
+            uvm_handle.get_user_vm_stream();
+        match user_vm_stream.lock() {
+            Ok(mut guard) => {
+                let (locked_uvm_stream, _): &mut (SocketStream, VecDeque<u8>) = &mut guard;
+                if let Err(e) = poll.registry().deregister(&mut *locked_uvm_stream) {
+                    error!("failed to de-revister user VM from poll (error={e:?})");
```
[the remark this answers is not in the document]
ppenna: +1

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,275 @@ pub struct LinuxDaemon {
...
+        poll: &Poll,
+        worker_threads: Option<VecDeque<WorkerThreadHandle>>,
+    ) {
+        // De-register the user VM socket from the poll structure.
+        let user_vm_stream: Arc<Mutex<(SocketStream, VecDeque<u8>)>> =
+            uvm_handle.get_user_vm_stream();
+        match user_vm_stream.lock() {
+            Ok(mut guard) => {
+                let (locked_uvm_stream, _): &mut (SocketStream, VecDeque<u8>) = &mut guard;
+                if let Err(e) = poll.registry().deregister(&mut *locked_uvm_stream) {
```
ppenna: Why not simply:

```rust
if let Err(e) = poll.registry().deregister(locked_uvm_stream) {
    error!("failed to de-revister user VM from poll (error={e:?})");
}
```

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,275 @@ pub struct LinuxDaemon {
...
+            },
+        };
+
+        // Send a shutdown message to all worker threads associated
+        // with this user VM.
+        if let Some(mut worker_threads) = worker_threads {
+            for worker_thread in worker_threads.drain(..) {
+                trace!("sending interrupt to worker thread (thread_id={:?})", worker_thread.id);
+
+                // If any of the commands fail, continue trying to drain the remainnig
```
ppenna: Typo "remaining".

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,275 @@ pub struct LinuxDaemon {
...
+                        error!("error sending interrupt to worker thread (thread_id={e:?})");
+                        continue;
+                    },
+                }
+                match worker_thread.handle.join() {
+                    Ok(_) => {},
+                    Err(e) => {
+                        error!("error joining worker thread (thread_id={e:?})");
+                    },
+                }
```
ppenna: As we care only about the error paths, switch to the `if let Err(e) = ...` pattern to keep the code easier to read.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,275 @@ pub struct LinuxDaemon {
...
+                // If any of the commands fail, continue trying to drain the remainnig
+                // threads.
+                match worker_thread.cmd_tx.send(VenvCommand::Shutdown) {
+                    Ok(_) => {},
+                    Err(e) => {
+                        error!("error sending shutdown command to worker thread (thread_id={e:?})");
+                        continue;
+                    },
+                }
+                match worker_thread.stop() {
```
ppenna: As the worker thread may not immediately react due to scheduling latencies, we may introduce a small delay here to avoid false-positive kills.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,275 @@ pub struct LinuxDaemon {
...
+            }
+        }
+    }
+
+    pub fn run(mut self) -> Result<(), Error> {
+        const CONTROL_PLANE_TOKEN: Token = Token(0);
+        const USER_VM_TOKEN: Token = Token(DEFAULT_CONNECTION_ID);
+
+        let mut events = Events::with_capacity(config::syscomm::MAX_NUM_POLL_EVENTS);
+        let mut poll =
```
ppenna: Please add type annotations.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,275 @@ pub struct LinuxDaemon {
...
+            .register(&mut self.control_plane_stream, CONTROL_PLANE_TOKEN, Interest::READABLE)
+            .map_err(|_| {
+                Error::new(ErrorCode::IoErr, "failed to register user VM listener to poll")
+            })?;
+        poll.registry()
+            .register(&mut self.uvm_stream, USER_VM_TOKEN, Interest::READABLE)
+            .map_err(|_| {
+                Error::new(ErrorCode::IoErr, "failed to register user VM listener to poll")
+            })?;
+
```
ppenna: Please log errors before failing. As `map_err` is commonly used, you may want to create a helper function that logs a given errors message and returns an error.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -68,179 +76,275 @@ pub struct LinuxDaemon {
...
+                }
+            },
+            Err(e) => {
+                error!("error acquiring lock on user VM stream (error={e:?})");
+            },
+        };
+
+        // Send a shutdown message to all worker threads associated
+        // with this user VM.
+        if let Some(mut worker_threads) = worker_threads {
```
ppenna, line 119: We should likely run similar logic on the destructor.

src/utils/nanvixd/src/control_plane.rs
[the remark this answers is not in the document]
ppenna: +1

src/utils/nanvixd/src/control_plane.rs
```diff
@@ -28,16 +27,19 @@ pub enum Command {
 // This function is actually used by linuxd, consuming nanvixd as a library. It is never used from
 // the main nanvixd binary, hence why we need to add this annotation to silent clippy.
 #[allow(dead_code)]
-pub fn read_command(stream: &mut BlockingSocketStream) -> Result<Command, SocketError> {
+pub fn try_read_command(stream: &mut SocketStream) -> Result<Command, SocketError> {
     let mut buf: [u8; 1] = [0u8; 1];
-    stream.read_exact(&mut buf)?;
+
+    // Try read exact returns the number of bytes written `n` with 0 < n <= buf.len(). In this case
+    // it is safe to ignore the return value because n can only ever be 1.
```
ppenna: Please catch the return value and `debug_assert`

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: I have added a comment explaining it. It is because we now use connection id 0 for the control-plane. I have also added a constant for the control-plane connection id for completenes.

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: This logic is ran twice: (1) when we read a EoF from the user VM socket, indicating that the connection has been closed, and (2) when we receive a shutdown message from the control plane.

In this PR I just copied and pasted the logic from (1) into a separate method so that it could be called in (2).

When you say destructor do you have any other situation in mind?

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: We are joining the thread right after in the next command. Is that what you are refering to?

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: Good idea. I also noticed that the error messasges were wrong.

src/daemons/linuxd/src/linuxd/mod.rs
ppenna: Thanks! As the rust doc wil lbe applied just to the first constant, could you decouple documentation? Here is a suggestion:

```
/// Connection ID used to register control-place sockets with a poll structure.
const CONTROL_PLANE_CONNECTION_ID: usize = 0;
/// Connection ID used to register user VM sockets sockets with a poll structure.
const DEFAULT_CONNECTION_ID: usize = 1;
```

src/daemons/linuxd/src/linuxd/mod.rs
ppenna: No. You are sending the shutdown command and right after you are sending a stop signal. I suggested:
- Send shutdown command
- Wait for a few milliseconds
- Send stop signal
- Join the thread

csegarragonz: Ah I see. The stop signal is just a signal to interrupt threads that are blocked on a system call. The signal handler is empty, so threads will just resume execution.

If they were not blocked on a system call, they were either blocked waiting to receive a command from the queue (in which case the shutdown command will make them stop) or doing some non-blocking work. Once they finish the work they will read from the queue, and realize they need to shut down.

What is the scheduling latency you are talking about? I may be missing something.

ppenna: Thanks, now it is clearer. Can you add precisely this explantion to the code?

PR #795, opened 2025-08-13
[microvm] Connect to Control Plane

csegarragonz:
In this PR I connect the micro VM to the control-plane in Nanvixd by establishing a socket stream at start-up time. The mechanism is identical to the one for linuxd. For the time being, the micro VM does not support processing any control-plane commands, this is because the micro VM lacks the ability to poll over the linuxd and control-plane sockets. I will address this limitation in a follow-up PR.

Each linuxd and micro VM instance get a different socket stream, accepted from the same socket listener. I had to move the binding logic in the socket listener from linuxd's spawn method, to the caller context and persist it in the sandbox cache.

@LucasGdosR: when this set of PRs is ready we should be able to send SNAPSHOT commands to the micro VM from nanvixd.

### DIFF
--- a/src/libs/config/src/lib.rs
+++ b/src/libs/config/src/lib.rs
@@ -74,6 +74,15 @@ pub mod syscomm {
     /// the same time.
     ///
     pub const MAX_NUM_POLL_EVENTS: usize = 128;
+
+    ///
+    /// # Description
+    ///
+    /// Provides the timeout we should use when accepting a connection. This timeout is used when
+    /// manually accepting connections, and it is needed to account for races between both ends.
+    /// These races should be in the order of milliseconds.
+    ///
+    pub const ACCEPT_TIMEOUT_SECS: u64 = 1;
 }
 
 //==================================================================================================
--- a/src/libs/syscomm/src/lib.rs
+++ b/src/libs/syscomm/src/lib.rs
@@ -32,6 +32,10 @@ use ::std::{
         Read,
         Write,
     },
+    time::{
+        Duration,
+        Instant,
+    },
 };
 
 //==================================================================================================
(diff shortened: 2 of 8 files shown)


src/utils/nanvixd/src/cache/mod.rs
```diff
@@ -16,21 +16,28 @@ use ::std::{
...
+};
 use ::tokio::sync::Mutex;
 
 //==================================================================================================
 // Structures
 //==================================================================================================
 
 /// A cache of sandboxes.
-#[derive(Clone)]
 pub struct SandboxCache {
```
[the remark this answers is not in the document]
ppenna: This is intentional.

src/utils/nanvixd/src/sandbox/microvm.rs
```diff
@@ -41,6 +53,8 @@ impl Microvm {
             program.to_string(),
             "-system-vm-addr".to_string(),
             addr.to_string(),
+            "-control-plane-addr".to_string(),
+            control_plane_addr.to_string(),
```
ppenna, line 63: Unrelated to this PR, we should have a good first issue that would expose the name of this CLI args so we could consume them using a constant.

src/microvm/src/args.rs
```diff
@@ -313,6 +345,34 @@ impl Args {
...
+    /// The control-plane address that was passed as a command-line argument to the program.
+    ///
+    pub fn control_plane_addr(&mut self) -> Option<String> {
+        self.control_plane_addr.take()
+    }
+
+    ///
+    /// # Description
+    ///
+    /// Returns the socket address type of the control-plane socket that was passed as a command-line
```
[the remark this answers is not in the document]
csegarragonz: This is the same format used across the file.

src/utils/nanvixd/src/sandbox/microvm.rs
```diff
@@ -20,20 +28,27 @@ use ::tokio::process::{
 pub struct Microvm {
     child: Option<Child>,
     addr: String,
+    #[allow(dead_code)]
+    // FIXME: the micro VM still does not support processing messages from the control-plane.
```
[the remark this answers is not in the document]
csegarragonz: This is temporary as will be fixed in a follow-up PR.

src/utils/nanvixd/src/sandbox/microvm.rs
csegarragonz: Good point, the constants already somewhat exist in  the args.rs module.

Tracking issue: https://github.com/nanvix/nanvix/issues/803

src/libs/config/src/lib.rs
```diff
@@ -74,6 +74,15 @@ pub mod syscomm {
     /// the same time.
     ///
     pub const MAX_NUM_POLL_EVENTS: usize = 128;
+
+    ///
+    /// # Description
+    ///
+    /// Provides the timeout we should use when accepting a connection. This timeout is used when
+    /// manually accepting connections, and it is needed to account for small races between both
+    /// ends.
```
ppenna: Please update last sentence to clarify "small races". This will be critical for high-density deployment.

src/libs/syscomm/src/lib.rs
```diff
@@ -106,6 +110,51 @@ impl SocketListener {
...
+    /// a poll structure.
+    ///
+    /// This is different from the BlockingSocketStream structure, that uses an internal poll to
+    /// receive in a blocking fashion.
+    pub fn accept_timeout(
+        &self,
+        poll: &mut Poll,
+        timeout: Duration,
+    ) -> Result<SocketStream, SocketError> {
+        let deadline = Instant::now() + timeout;
```
ppenna: Add a type annotation here.

src/libs/syscomm/src/lib.rs
```diff
@@ -106,6 +110,51 @@ impl SocketListener {
...
+                Err(ref e) if e.kind() == ErrorKind::Interrupted => continue,
+                // If there was a transient error in the connection, retry immediately.
+                Err(ref e) if e.kind() == ErrorKind::ConnectionAborted => continue,
+                Err(e) => {
+                    error!("error accepting connection (error={e:?})");
+                    return Err(e);
+                },
+            }
+
+            let now = Instant::now();
```
ppenna: Add type annotation here.

src/libs/syscomm/src/lib.rs
```diff
@@ -106,6 +110,51 @@ impl SocketListener {
...
+            if now >= deadline {
+                error!("accept timed-out waiting for connection");
+                return Err(io::Error::new(ErrorKind::TimedOut, "accept timed out").into());
```
ppenna: Create a static string and reuse it on the log error, as well as to return the error.

ppenna (approved):

PR #806, opened 2025-08-14
[daemons] Make Linuxd Poll For Gateway Connections

csegarragonz:
In this commit I modify linuxd to keep a poll structure monitoring the gateway listener socket, and accept incoming connections from the gateway by calling accept_timeout on the listener socket itself.

We need a different polling structure than the one used for the user VMs because we want to accept connections from the gateway on-demand, as user VMs connect. On top of that, once we have accepted the connection, and have the corresponding stream, we don't need the poll anymore as we want each worker thread to read/write to the stream directly.

### DIFF
--- a/src/daemons/linuxd/src/linuxd/mod.rs
+++ b/src/daemons/linuxd/src/linuxd/mod.rs
@@ -40,6 +40,7 @@ use ::std::{
         Mutex,
         MutexGuard,
     },
+    time::Duration,
 };
 use ::sys::{
     error::{
@@ -50,7 +51,10 @@ use ::sys::{
     pm::ThreadIdentifier,
 };
 use ::syscall::venv::VirtualEnvironmentIdentifier;
-use ::syscomm::SocketStream;
+use ::syscomm::{
+    SocketListener,
+    SocketStream,
+};
 
 //==================================================================================================
 // Constants
@@ -60,6 +64,9 @@ use ::syscomm::SocketStream;
 const CONTROL_PLANE_CONNECTION_ID: usize = 0;
 /// We use ID 1 for the user VM connection in the main poll structure.
 const DEFAULT_CONNECTION_ID: usize = 1;
+/// We use ID 0 for the gateway listener socket in the gateway poll structure. Given that this
+/// socket is monitored in a different poll, we can re-use the connection ID 0.
+const GATEWAY_LISTENER_CONNECTION_ID: usize = 0;
 
 //==================================================================================================
 // Structures
@@ -69,7 +76,7 @@ pub struct LinuxDaemon {
     assembler: Arc<Mutex<RequestAssembler>>,
     control_plane_stream: SocketStream,
     uvm_stream: SocketStream,
-    gateway_stream: Option<SocketStream>,
+    gateway_listener: Option<SocketListener>,
     venv: Arc<Mutex<VirtualEnviromentDirectory>>,
 }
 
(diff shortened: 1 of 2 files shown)


src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -164,33 +171,94 @@ impl LinuxDaemon {
...
+                )
+                .map_err(|_| {
+                    Self::log_and_error(
+                        ErrorCode::IoErr,
+                        "failed to register gateway listener to poll",
+                    )
+                })?;
+
+            Some(gateway_poll)
+        } else {
```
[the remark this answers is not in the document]
ppenna: The suggestion is incorrect.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -164,33 +171,94 @@ impl LinuxDaemon {
...
+            None
+        };
+
+        // Accept one incoming connection from the gateway.
+        let gateway_stream: Option<SocketStream> = if let Some(mut gateway_poll) = gateway_poll {
```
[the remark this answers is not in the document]
ppenna: This is ok for now.

ppenna (dismissed):

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -164,33 +171,94 @@ impl LinuxDaemon {
...
+        // the gateway we want to give each worker thread the mutex-protected socket stream
+        // connected to the gateway.
+        let gateway_poll: Option<Poll> = if self.gateway_listener.is_some() {
+            let gateway_poll: Poll = Poll::new()
+                .map_err(|_| Self::log_and_error(ErrorCode::IoErr, "failed to create Poll"))?;
+            gateway_poll
+                .registry()
+                .register(
+                    // We need to check is_some() and get the mutable in-place here instead of
+                    // using an if let Some(... to keep the borrow-checker happy.
```
ppenna, line 204: Typo: missing close ')'

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -164,33 +171,94 @@ impl LinuxDaemon {
...
+                    Self::log_and_error(
+                        ErrorCode::IoErr,
+                        "failed to register gateway listener to poll",
+                    )
+                })?;
+
+            Some(gateway_poll)
+        } else {
+            None
+        };
```
ppenna: Why not:

```rust
let gateway_poll: Option<Poll> =
    if let Some(gateway_listener) = self.gateway_listener.as_mut() {
        let gateway_poll: Poll = Poll::new()
            .map_err(|_| Self::log_and_error(ErrorCode::IoErr, "failed to create Poll"))?;
        gateway_poll
            .registry()
            .register(gateway_listener, GATEWAY_LISTENER_TOKEN, Interest::READABLE)
            .map_err(|_| {
                Self::log_and_error(
                    ErrorCode::IoErr,
                    "failed to register gateway listener to poll",
                )
            })?;

        Some(gateway_poll)
    } else {
        None
    };
```

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -164,33 +171,94 @@ impl LinuxDaemon {
...
+                    .map_err(|_| {
+                        Self::log_and_error(
+                            ErrorCode::IoErr,
+                            "error accepting connection from gateway",
+                        )
+                    })?,
+            )
+        } else {
+            None
+        };
```
ppenna: Why not:

```rust
let gateway_stream: Option<SocketStream> = if let Some(mut gateway_poll) = gateway_poll {
    if let Some(gateway_listener) = self.gateway_listener.as_mut() {
        Some(
            gateway_listener
                .accept_timeout(
                    &mut gateway_poll,
                    Duration::from_secs(config::syscomm::ACCEPT_TIMEOUT_SECS),
                )
                .map_err(|_| {
                    Self::log_and_error(
                        ErrorCode::IoErr,
                        "error accepting connection from gateway",
                    )
                })?,
        )
    } else {
        None
    }
} else {
    None
};
```

csegarragonz: Applied your patch, thanks! :)

src/daemons/linuxd/src/linuxd/mod.rs
csegarragonz: Applied your patch, thanks! :)

PR #818, opened 2025-08-15
[microvm] Interrupt vCPU With Signal

csegarragonz:
In this commit I implement support to interrupt the vCPU thread by sending it a SIGUSR1 signal.

According to the KVM docs [1], sending a SIGUSR1 to the vCPU thread will cause KVM_RUN to return with -EINTR. For the time being we treat an interrupt as an outside request to halt execution, but we can extend this functionality in the future.

I have tested this functionality manually, but there is no way to trigger it programatically yet as we are lacking a connection with the control-plane socket.

[1] https://www.kernel.org/doc/html/v5.8/virt/kvm/api.html#kvm-set-signal-mask

### DIFF
--- a/src/microvm/src/vmm/microvm/mod.rs
+++ b/src/microvm/src/vmm/microvm/mod.rs
@@ -37,14 +37,20 @@ use crate::{
     },
 };
 use ::anyhow::Result;
+use ::libc::pthread_self;
 use ::std::{
     fs::File,
     io::Write,
     mem,
     sync::{
         Arc,
+        Barrier,
         Mutex,
         MutexGuard,
+        atomic::{
+            AtomicUsize,
+            Ordering,
+        },
         mpsc,
         mpsc::{
             Receiver,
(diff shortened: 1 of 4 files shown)


src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -236,8 +241,19 @@ impl Vmm {
...
+        let pthread_id_holder: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
+
         let microvm_clone: Arc<Mutex<MicroVm>> = microvm.clone();
+        let pthread_id_holder_clone: Arc<AtomicUsize> = pthread_id_holder.clone();
         let vcpu_thread: JoinHandle<Result<u16>> = std::thread::spawn(move || {
+            // Store the tid so that the caller can send signals to the vCPU thread.
+            // SAFETY: we are calling pthread_self() right after creating the thread so this is
+            // safe.
+            let pthread_id: libc::pthread_t = unsafe { pthread_self() };
+            pthread_id_holder_clone.store(pthread_id as usize, Ordering::Relaxed);
```
ppenna, line 259: There maybe be a race condition if an interrupt is sent before this line is reached. We should have a barrier here to prevent this corner case.

csegarragonz: Good call. I have added a barrier to make sure we don't start the vCPU thread until the atomic value has been set.

ppenna (approved):

PR #826, opened 2025-08-16
[daemons] Index Venvs By Thread and User VM Id

csegarragonz:
In this commit I extend the virtual environment abstraction in linuxd to index thread virtual environments by a combination of the thread ID, unique per user VM, and the user VM ID, unique per linuxd instance.

Collectively, both things represent what I called a global thread identifier, where I put the connection ID on the top 32 bits of a u64, and the thread ID on the bottom ones.

### DIFF
--- a/src/daemons/linuxd/src/venv.rs
+++ b/src/daemons/linuxd/src/venv.rs
@@ -28,7 +28,7 @@ use ::syscall::venv::VirtualEnvironmentIdentifier;
 ///
 /// Unique identifier for each user VM.
 ///
-pub type UserVmIdentifier = usize;
+pub type UserVmIdentifier = u32;
 
 ///
 /// # Description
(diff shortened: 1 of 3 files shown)


src/daemons/linuxd/src/venv.rs
```diff
@@ -23,6 +23,20 @@ use ::sys::{
 };
 use ::syscall::venv::VirtualEnvironmentIdentifier;
 
+///
+/// # Description
+///
+/// Unique identifier for each user VM.
+///
+pub type UserVmIdentifier = usize;
```
ppenna: Consider using `u32`  as `usize` is not portable.

src/daemons/linuxd/src/venv.rs
```diff
@@ -115,6 +129,46 @@ impl VirtualEnviromentDirectory {
...
+    /// # Returns
+    ///
+    /// On success, a u64 where the user VM id sits in the first 32 bits, and the thread id in the
+    /// second 32 bits. On error an error is returned.
+    ///
+    fn get_gtid(
+        uvmid: UserVmIdentifier,
+        tid: ThreadIdentifier,
+    ) -> Result<GlobalThreadIdentifier, Error> {
+        let uvmid: u32 = match u32::try_from(uvmid) {
```
ppenna: By using `u32` for user VMs you avoid this.

ppenna: By using `u32` for user VMs you avoid this.

src/daemons/linuxd/src/venv.rs
csegarragonz: I chose `usize` because that is the key value for `Slab`, so I was using it to index the different data structures. In the long run it is better to have `u32`, because we, at some point, will start indexing things with the user VM ID assigned by nanvixd, not the connection id.

In the last commit I change the indexing to use u32, and convert when appropriate.

ppenna (approved):

PR #854, opened 2025-08-20
[microvm] F: Adding pause and resume functionality

LucasGdosR:
# Feedback requested

I want feedback on the pause functionality. I'm more confident in the resume functionality, and I don't need feedback on it for now.

1. Is it ok to write a flag for pausing the MicroVM at `::config::microvm::DEFAULT_MICROVM_CTRL_BASE`, or is this address already taken? If so, where should I write?
2. I think there're 32 bits free at that address. However, I only need a single bit for this flag. Should I use a u32 or a u8 for this, or something else?
3. In the kernel, I should probably wrap the pausing functionality with a `cfg_if::cfg_if!`. What feature should I use?
4. How should I issue the VM exit and set the exit context in the kernel? I see that @csegarragonz used the KVM documentation to figure out what the exit context would be when sending a particular signal. Can I arbitrarily choose an exit context? If so, how?

# Implementation Description

This implements pausing and resuming via control-plane messages to the VMM via co-design with the kernel. Let's go through a whole pause-resume cycle to understand the implementation.

## Pause

The changes start at `mod.rs`. The VMM now has communication channels to the MicroVM, which are used for syncing the main thread and the vCPU thread. In the main thread, when the VMM receives a `ControlCommand::PauseMicroVm`, it calls `pause_protocol()` inside `handle_command()`. `pause_protocol()` now calls `MicroVm.asnc_pause()`, then waits for a confirmation from the channel to the MicroVM that the vCPU thread operates. I will add a timeout check at this point, as per the TODO.

Let's now switch to `microvm.rs`. It also has the channels to communicate between the vCPU thread and the main thread. The `async_pause()` function writes to the guest OS's kernel's memory a bit telling it to stop execution. Conceptually, we could say there's a state transition in the MicroVM from "running" to "pausing". I would like feedback on the address that's written to. Is this address already taken? If so, what address should I use?

Let's now briefly switch to `lib.rs`. It just defines the values that are supposed to be written / read in the kernel memory. I would like feedback on using an u8 vs an u32 for this.

Let's now go deep into the kernel. This runs in the vCPU thread. In `clock.rs`, the `timer_handler()` function runs whenever there's a timer interrupt. It reads the same address that was written to by `async_pause()`. If the value read corresponds to the pause request, it issues a VM exit with an exit context of `Paused`. I would like feedback on what to wrap this statement with, and how to issue the VM exit and set the exit context.

Let's now go back to `microvm.rs`. Once the VM exit is executed, the vCPU thread goes back to the body of the `MicroVm::run()` function. It's done with `self.vcpu.run()?`, and matches on the `exit_context.reason()`. A new reason was added for `Paused`. It sends a message to the VMM confirming the vCPU has stopped. That was the message it was waiting for from the beginning of this explanation. Conceptually, this changes the MicroVM state from "pausing" to "paused". To actually implement being paused, it waits for a message telling it to resume. I would like feedback on this. Should this be a blocking read? If not, what else should we do besides polling for the resume message? P.S.: I just realized that I need to poll this instead if I want to implement a timeout.

Let's briefly switch to `exit.rs`. All it does is define the new exit context and exit reason.

Now that the MicroVM is paused, we could take a consistent snapshot. However, that's for another PR. Let's turn our attention to resuming the MicroVM.

## Resume

Back at the VMM (`mod.rs`) in the main thread, we're now handling a `ControlCommand::ResumeMicroVm`. We call `resume_microvm()`, which calls `MicroVm::resume()`. It then sends a message to the channel that was expecting a message to continue the `MicroVm::run()` loop in the vCPU thread. Receiving the message conceptually changes the MicroVM state from `paused` to `running`. It's important to say that `resume_microvm()` is synchronous, so its effects will be done before the message is sent, and therefore before it is received. We'll dive deep into this soon.

Let's go to `microvm.rs` and look at `resume_microvm()`. All it does is clear the bit that requested the kernel to pause execution.

Let's focus on ordering. The VMM thread clears the bit, then it sends a message to the vCPU thread telling it to resume execution. This is an asynchronous message, to the VMM may continue handling control commands while the vCPU hasn't really resumed. Suppose the VMM receives a `ControlCommand::Pause` before the vCPU is schedulled. Then, the VMM would execute `pause_protocol()`, and write to the kernel's memory the `PAUSE_REQUESTED` value.  The VMM would eventually stop while waiting for a message from the MicroVM confirming it has paused. At this point, the vCPU thread must be schedulled. It goes back to the `MicroVm::run()` loop, and executes for a quantum. When the interrupt goes off, the kernel reads the memory with the `PAUSE_REQUESTED` value and pauses execution once more. Then, it sends the message to the main thread. Therefore, there are no race conditions in this case, and there's no need to wait for a confirmation message in `Vmm::resume_microvm()`. Just sending a message is enough.

### DIFF
--- a/src/kernel/src/pm/clock.rs
+++ b/src/kernel/src/pm/clock.rs
@@ -90,7 +90,7 @@ impl TimerTicks {
 ///
 /// # Description
 ///
-/// Handles a timer interrupt.
+/// Handles a timer interrupt. It may cause a VM exit if a pause has been requested.
 ///
 /// # Parameters
 ///
@@ -104,6 +104,19 @@ impl TimerTicks {
 ///
 pub unsafe fn timer_handler(_intnum: InterruptNumber) {
     TIMER_TICKS.increment();
+
+    // Check if a pause has been requested.
+    #[cfg(feature = "microvm")]
+    if core::ptr::read_volatile(
+        ::config::microvm::DEFAULT_MICROVM_CTRL_PAUSE_REQUESTED as *const u32,
+    ) == ::config::microvm::PAUSE_REQUEST
+    {
+        // Cause a VM exit.
+        ::arch::io::out32(
+            ::config::microvm::DEFAULT_VMM_PORT,
+            (::config::microvm::DEFAULT_VMM_PAUSE_CMD as u32) << 16,
+        )
+    }
 
     // Determine if a context switch is required. The kernel's running state is checked first to
     // prevent reentrant calls to the scheduler, which could lead to undefined behavior.
--- a/src/libs/config/src/lib.rs
+++ b/src/libs/config/src/lib.rs
@@ -252,6 +252,9 @@ pub mod microvm {
     /// Default VMM shutdown command
     pub const DEFAULT_VMM_SHUTDOWN_CMD: u16 = 0x2000;
 
+    /// Default VMM pause command. This MUST be a value that's never an exit code.
+    pub const DEFAULT_VMM_PAUSE_CMD: u16 = 0x3000;
+
     /// Default base address for MicroVM control registers.
     pub const DEFAULT_MICROVM_CTRL_BASE: usize = 0x00000000;
 
@@ -260,6 +263,15 @@ pub mod microvm {
 
     /// Default base address for MicroVM credits register (32-bit wide read-only register)
     pub const DEFAULT_MICROVM_CTRL_CREDITS: usize = 0x00000004;
+
+    /// Default base address for MicroVM pause-requested register (32-bit wide read-only register)
+    pub const DEFAULT_MICROVM_CTRL_PAUSE_REQUESTED: usize = 0x00000008;
+
+    /// Magic value that identifies the running state in the pause-requested register.
+    pub const RUNNING: u32 = 0x00000000;
+
+    /// Magic value that flags that the VMM requested the guest OS to pause MicroVM execution.
+    pub const PAUSE_REQUEST: u32 = 0x00000001;
 }
 
 #[cfg(feature = "pc")]
--- a/src/microvm/src/vmm/microvm/kvm/emulator.rs
+++ b/src/microvm/src/vmm/microvm/kvm/emulator.rs
@@ -120,6 +120,9 @@ impl Emulator {
                             let status: u16 = (data & 0xffff) as u16;
                             return Ok(Some(status));
                         },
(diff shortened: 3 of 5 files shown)


ppenna:
> Is it ok to write a flag for pausing the MicroVM at ::config::microvm::DEFAULT_MICROVM_CTRL_BASE, or is this address already taken? If so, where should I write?

Yes, the entire page is reserved for virtual registers [[1](https://github.com/nanvix/nanvix/blob/7a7f51a08adb25660a84cf8d5a55242123a2d210/src/kernel/src/hal/platform/microvm.rs)]. Note that the following addresses are already taken:
- 0-3: Null register https://github.com/nanvix/nanvix/blob/7a7f51a08adb25660a84cf8d5a55242123a2d210/src/libs/config/src/lib.rs#L259
- 4-7: Credits register https://github.com/nanvix/nanvix/blob/7a7f51a08adb25660a84cf8d5a55242123a2d210/src/libs/config/src/lib.rs#L262

ppenna:
> Should I use a u32 or a u8 for this, or something else?

It should be a 32-bit register for performance reasons.

ppenna:
> What feature should I use.

Check if you are running on `microvm`, like here:

https://github.com/nanvix/nanvix/blob/7a7f51a08adb25660a84cf8d5a55242123a2d210/src/kernel/src/hal/platform/mod.rs#L17

ppenna:
> Can I arbitrarily choose an exit context? If so, how?

Let's not take arbitrarily decisions when writing software :)

For causing a VM exit for your case, you can create a new command that is multiplexed on `DEFAULT_VMM_PORT`, like https://github.com/nanvix/nanvix/blob/7a7f51a08adb25660a84cf8d5a55242123a2d210/src/libs/config/src/lib.rs#L253

And handle it accordingly: https://github.com/nanvix/nanvix/blob/7a7f51a08adb25660a84cf8d5a55242123a2d210/src/microvm/src/vmm/microvm/kvm/emulator.rs#L105-L125

src/kernel/src/pm/clock.rs
```diff
@@ -153,3 +162,86 @@ pub fn now() -> SystemTime {
         },
     }
 }
+// _____________________________________________
+// THIS IS TRASH. DON'T LOOK. YOU'VE BEEN WARNED
+// _____________________________________________
```
ppenna: 😆

PR #856, opened 2025-08-20
[microvm] Add pause and resume functionality

### LINKED ISSUE #791: [kernel] Cause VM exit when a magic number is read from a magic address
## TLDR

In order to implement MicroVM pauses, when should the kernel opportunistically check for requests to issue VM exits? Where in the kernel source code can I find code that runs in that time?

## Context

The MicroVM executes the guest code when it calls `self.fd.run()`. In order to pause execution (for snapshot consistency), this call must be interrupted if it's already in place. This can be done via signals, or via the kernel running in the guest OS. I want the kernel to opportunistically check if it should issue a VM exit to end `self.fd.run()` instead of using signals.

## Proposed Solution

I want a request to pause the MicroVM to arrive through the IO thread, travel to the VMM, which would write in a designated address, and the kernel would periodically read this designated address to check if it should pause.

Given that a guest program may never issue syscalls, pause latency would be unbounded if it relied only on syscalls. I believe this opportunistic check should be implemented at the end of a timer interrupt. I'm not entirely sure how this would behave with an I/O-bound application that never uses up its quantum. Would the timer still work, or would it be necessary to implement this opportunistic check also in syscalls (for I/O) or maybe when I/O is completed?

Actually, I have an idea. Kernels are sort of stateless, right? I mean, they keep some process state in a process table, but syscalls are kind of stateless, or something like that. I don't know the details, but there's something here. Could we exploit this "stateless" nature to choose a point in execution to pause where a large portion of the kernel memory is always the same? For example, is the kernel memory mostly the same at the beginning of every syscall? Or at the end of every syscall? Or whenever a timer interrupt occurs? If so, all snapshots could share this large portion of the kernel memory, and we wouldn't need to save it.

LucasGdosR:
# Overview
This implements pausing and resuming via control-plane messages to the VMM via co-design with the kernel. The VMM writes to a kernel register. The kernel polls this register during timer interrupts, and exits the VM if flagged. It pauses execution and waits for a resume messages, which might be later sent by the VMM.

This does not implement pausing and resuming for hyperlight.

# Attention points:
- `DEFAULT_VMM_PAUSE_CMD` must not collide with other exit codes for the MicroVM;
- When pausing, the main thread sends to the memory thread, then receives from the vCPU thread. The main thread only receives from the vCPU thread after the vCPU thread has read the memory in the kernel that the memory thread wrote, so there's a total order in these events.;
- When resuming, the main thread sends to the memory thread, then receives from the memory thread, then sends to the vCPU thread. This means there's also a total order, and the vCPU thread will never resume the MicroVM with a pause request written in the kernel's memory.

### DIFF
--- a/src/microvm/src/vmm/microvm/microvm.rs
+++ b/src/microvm/src/vmm/microvm/microvm.rs
@@ -15,6 +15,7 @@
 #[cfg(target_os = "linux")]
 use crate::{
     orchestrator::{
+        TIMEOUT_WARNING_INTERVAL_IN_MS,
         VcpuControlCommand,
         VcpuControlResponse,
     },
@@ -38,13 +39,17 @@ use ::libc::{
     sigaction,
     sigemptyset,
 };
-use ::std::sync::{
-    Arc,
-    Mutex,
-    mpsc::{
-        Receiver,
-        Sender,
+use ::std::{
+    sync::{
+        Arc,
+        Mutex,
+        mpsc::{
+            Receiver,
+            Sender,
+            TryRecvError,
+        },
     },
+    time::Instant,
 };
 
 pub const INTERRUPT_SIGNAL: c_int = SIGUSR1;
@@ -70,9 +75,9 @@ pub struct MicroVm {
     // If present, initial RAM disk location and size.
     initrd: Option<(u64, usize)>,
     // Channel to receive commands from the VMM.
-    _control_rx: Receiver<VcpuControlCommand>,
+    control_rx: Receiver<VcpuControlCommand>,
     // Channel to send control responses to the VMM.
-    _control_tx: Sender<VcpuControlResponse>,
+    control_tx: Sender<VcpuControlResponse>,
 }
 
 unsafe impl Send for MicroVm {}
@@ -138,8 +143,8 @@ impl MicroVm {
             vcpu,
             emulator,
             initrd: None,
-            _control_rx: control_rx,
-            _control_tx: control_tx,
+            control_rx,
+            control_tx,
         })
     }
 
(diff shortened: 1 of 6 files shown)


ppenna:
@LucasGdosR does this superspeed #854?

LucasGdosR:
> @LucasGdosR does this superspeed #854?

Kind of. It's the same commit, I just switched the branch name to match the standard for regular PRs.

I'll do some manual testing before asking for a proper review.

ppenna:
Oh, I see. Thanks for being careful. The good news is that you can always push matching branches. The CI Will run only on non-draft PRs.

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -554,9 +561,35 @@ impl Vmm {
...
     fn pause_protocol(&mut self) -> Result<()> {
-        // TODO: pause MicroVM (Running -> Paused)
-        // and tell linuxd to flush (Running -> Flushing)
-        // This TODO requires pausing the vCPU and a control plane communication with linuxd
+        // TODO: tell linuxd to flush (Running -> Flushing)
+        // This TODO requires control plane communication with linuxd
+        self.microvm
+            .lock()
+            .map_err(|e| anyhow::anyhow!("failed to acquire lock {e:?}"))?
+            .async_pause()?;
```
LucasGdosR: I'm testing by hardcoding messages into the I/O thread (not in this commit). This call to `async_pause()` never happens, because `self.microvm` cannot be locked (although the error message never shows up). It seems to be locked elsewhere. The `vcpu_thread` calls:
```
microvm_clone
                .lock()
                .map_err(|e| anyhow::anyhow!("failed to acquire lock {e:?}"))?
                .run()
```
I'm thinking that's the problem. There shouldn't be a single lock that's used both for running the microvm and for pausing it, otherwise there's no way to interrupt it. Do you have suggestions on how to organize this into two different locks, or something else?

ppenna: It is likely that some rework is required:

At a high level, to handle a snapshot you:
- Receive snapshot signal
- Sinal the kernel that a snapshot will happen
- The kernel issues a VM exit
- The vCPU unlocks the required resources

ppenna:
@LucasGdosR could you reopen #919 and have it independently merged with this one? We should have 1 PR for each feature/enhancement/bugfix whenever possible. As you stated in #919 the changed that you introduced there, but then later incorporated here, can be merged independently.

LucasGdosR:
I have reopened https://github.com/nanvix/nanvix/pull/919 and I'll include the `orchestrator.rs` part in it, without the pause-resume functionality. It's also duplicated code.

LucasGdosR:
@ppenna @csegarragonz this PR is now ready to be reviewed. I hope it'll be easy now that most of the code was merged in previous PRs. If there are no issues with it, I'll convert it to a regular PR.

I have tested the pause-resume functionality locally with a custom user program which you can find here, if you'd like https://github.com/nanvix/nanvix/tree/feature-long-running-rust-nostd-user-program

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -319,7 +324,45 @@ impl MicroVm {
...
+                            loop {
+                                match self.control_rx.try_recv() {
+                                    Ok(VcpuControlCommand::Resume) => break,
+                                    // NOTE: Should we add an option for shutting down? Like so:
+                                    // Ok(VcpuControlCommand::Shutdown) => self.vcpu.poweroff(0),
+                                    Err(TryRecvError::Empty) => (),
+                                    Err(TryRecvError::Disconnected) => {
+                                        let reason: String = "the vmm has disconnected".to_string();
+                                        error!("run(): {reason}");
+                                        anyhow::bail!(reason)
```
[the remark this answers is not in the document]
ppenna: This suggestion is incorrect.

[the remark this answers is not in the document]
ppenna: This suggestion is incorrect.

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -319,7 +324,45 @@ impl MicroVm {
...
+                                // should we use a larger constant, considering snapshots might take long?
+
+                                // Log a warning and increment the counter every TIMEOUT_WARNING_INTERVAL_IN_MS ms.
+                                let elapsed_time: usize = start.elapsed().as_millis() as usize;
+                                if elapsed_time > TIMEOUT_WARNING_INTERVAL_IN_MS * counter {
+                                    warn!(
+                                        "{}ms have passed waiting for `ResumeMicroVm` message",
+                                        TIMEOUT_WARNING_INTERVAL_IN_MS * counter
+                                    );
+                                    counter += 1;
```
ppenna, line 362: Could you link the issue of the exponential back-off timeout here?

ppenna (dismissed):

PR #919, opened 2025-08-29
[microvm] Collapses common code between microvm and hyperlight

### LINKED ISSUE #873: [microvm] Deduplicate Code for microvm and hyperlight Machine Types
## Context

The `microvm` and `hyperlight` machines have feature parity. We shall merge code duplicate between them.

LucasGdosR:
The I/O thread code can be exactly the same. For now, hyperlight doesn't use every channel that's created, so they're left with a leading underscore.

The memory thread is not exactly the same. A closure is used to encapsulate the different code.


LucasGdosR:
These changes have been incorporated into https://github.com/nanvix/nanvix/pull/856. They can be merged independently or together.

LucasGdosR:
Superseded by https://github.com/nanvix/nanvix/pull/856. It also colapses orchestrating logic.

LucasGdosR:
All the code regarding the snapshot protocol is now part of a separate file that's reused between the microvm and hyperlight. It will also use callback functions to deal with different implementations between the microvm and hyperlight, particularly for creating snapshots and loading snapshots.

src/microvm/src/memory.rs
```diff
@@ -0,0 +1,77 @@
...
+///
+/// # Parameters
+///
+/// - `data_rx`: Receives data messages from the I/O thread.
+/// - `data_tx`: Sends data messages to the virtual machine's stdin.
+/// - `add_credit`: Closure that adds a credit to the virtual machine credit pool.
+///
+/// # Returns
+///
+/// A handle to the I/O thread.
```
csegarragonz: A handle to the memory thread.

src/microvm/src/orchestrator.rs
```diff
@@ -0,0 +1,275 @@
...
+use crate::io::{
+    ControlCommand,
+    ControlCommandResponse,
+};
+use ::anyhow::Result;
+use ::std::sync::mpsc::{
+    Receiver,
+    Sender,
+};
+use std::{
```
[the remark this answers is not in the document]
ppenna: @LucasGdosR please address this.

LucasGdosR:
Many thanks for the review, and sorry for being sloppy with the description of the memory thread and the imports. I have corrected those and squashed the bug fixing commits.

src/microvm/src/memory.rs
```diff
@@ -0,0 +1,77 @@
+// Copyright(c) The Maintainers of Nanvix.
+// Licensed under the MIT License.
+
```
csegarragonz: I would personally name this module `memory_thread` as it is more explainatory than `memory`.

Can we also add a module level block comment explaining why we need a memory_thread and what is its purpose?

Something like:

```rust
//! Bla bla
```

src/microvm/src/memory.rs
```diff
@@ -0,0 +1,77 @@
...
+use ::sys::ipc::Message;
+
+//==================================================================================================
+// Standalone Functions
+//==================================================================================================
+
+///
+/// # Description
+///
+/// Spawns a new memory thread.
```
csegarragonz: Same as above. What is a memory thread, why do we need one?

src/microvm/src/orchestrator.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+//==================================================================================================
+// Enumerations
+//==================================================================================================
+
+///
+/// # Description
+///
+/// States relating to snapshots functionality. Snapshots may be loaded at PreBoot, and created at Paused.
```
csegarragonz: I think this is not particular to snapshots, but to the VMM itself?

src/microvm/src/orchestrator.rs
```diff
@@ -0,0 +1,275 @@
+// Copyright(c) The Maintainers of Nanvix.
+// Licensed under the MIT License.
+
```
csegarragonz, line 3: Maybe nit-pick, but I don't think `orchestrator` is the right name for this module neither? The way I see it this module handles the state machine for the VMM, and the state machine transitions.

csegarragonz (commented):
Some comments regarding naming, happy to hear other's opinions.

src/microvm/src/memory.rs
LucasGdosR: I agree with the module name proposal. I used `memory` because the previous module for the io thread was called `io`, but I think that's also inappropriate. I'd change them both to include `_thread`. I also agree a module level block comment would make it clearer for newcomers, and it would also be useful in the `io_thread` module and the (to be renamed) `orchestrator` module.

src/microvm/src/orchestrator.rs
LucasGdosR: It holds all control channels between the main thread and all other threads, and those control channels are used for orchestrating the other threads.

For example, the vCPU thread can only resume running the MicroVM once the memory thread is done writing to the kernel's register, otherwise there would be a race condition where the vCPU thread might resume and pause again before the memory thread writes to the kernel. The logic to avoid these race conditions would be in this module, using those control channels.

I think this module holds the main thread's functionality of communicating with the other threads, and the state machine transitions based on this communication.

src/microvm/src/orchestrator.rs
LucasGdosR: I agree. For example, these states could be used for live migrations as well.

ppenna:
@LucasGdosR and @csegarragonz I am taking over this PR to get it merged.

src/microvm/src/memory.rs
ppenna: Introduced module-level documentation.

src/microvm/src/memory.rs
ppenna: Improved documentation.

src/microvm/src/orchestrator.rs
ppenna: Improved documentation.

src/microvm/src/memory.rs
ppenna: Renamed modules.

src/microvm/src/orchestrator.rs
ppenna: Keeping orchestrator for now.

src/microvm/src/vmm/hyperlight/mod.rs
```diff
@@ -254,54 +277,27 @@ impl Vmm {
...
-                        break Ok(());
-                    },
-                    Err(TryRecvError::Empty) => {
-                        // No message available.
-                    },
-                }
-            }
-        });
+        let memory_thread: JoinHandle<Result<(), anyhow::Error>> =
+            memory::spawn(memory_thread_rx, memory_thread_tx, add_credit);
```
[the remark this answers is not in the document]
ppenna: This suggestion is invalid.

src/microvm/src/memory_thread.rs
```diff
@@ -0,0 +1,84 @@
...
+    F: FnMut() -> Result<()> + std::marker::Send + 'static,
+{
+    thread::spawn(move || {
+        loop {
+            match data_rx.try_recv() {
+                Ok(mut msg) => {
+                    profiler::timestamp_message!(
+                        &mut msg.payload,
+                        mem::offset_of!(syscall::LinuxDaemonMessage, payload)
+                            + mem::offset_of!(syscall::unistd::message::ReadResponse, buffer)
```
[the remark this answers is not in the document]
ppenna: The suggestion is wrong.

src/microvm/src/vmm/hyperlight/mod.rs
```diff
@@ -18,22 +12,47 @@ extern crate kvm_ioctls;
 
 use crate::{
     Gateway,
-    vmm::hyperlight::io::IoThread,
+    io::{
+        ControlCommand,
+        ControlCommandResponse,
+        IoThread,
+    },
+    memory,
```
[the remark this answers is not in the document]
ppenna: Fixed.

src/microvm/src/vmm/hyperlight/mod.rs
```diff
@@ -18,22 +12,47 @@ extern crate kvm_ioctls;
 
 use crate::{
     Gateway,
-    vmm::hyperlight::io::IoThread,
+    io::{
+        ControlCommand,
+        ControlCommandResponse,
+        IoThread,
+    },
```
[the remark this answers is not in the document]
ppenna: Fixed.

csegarragonz (approved):
LGTM! Thanks @LucasGdosR, this will help a lot!

PR #935, opened 2025-09-04
[user] Add long running user application for testing snapshots

LucasGdosR:
This program runs a loop from 0 to u64::MAX and prints every 500_000_000 iterations. It doesn't rely on any Linux daemon state. It should be useful to test pausing and resuming the MicroVM and seeing that it continues from where it left off, and also for creating and loading snapshots and seeing that it continues too.

Later on, a new application will be needed to include Linux daemon interactions.


src/user/snapshot-rust-nostd/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+    let mut pos = buf.len() - 1;
+    buf[pos] = b'\n';
+    while num > 0 {
+        pos -= 1;
+        buf[pos] = (num % 10) as u8 + b'0';
+        num /= 10;
+    }
+
+    unsafe { core::str::from_utf8_unchecked(&buf[pos..]) }
+}
```
ppenna: You can leverage `alloc::format!()` to do this for you (https://doc.rust-lang.org/alloc/macro.format.html).

ppenna (changes requested):

ppenna:
I am cancelling CI runs while you work on the requested changes.

LucasGdosR:
Should be good to go. I compiled it and ran it.

src/user/snapshot-rust-nostd/src/main.rs
```diff
@@ -0,0 +1,44 @@
...
+//==================================================================================================
+// Standalone Functions
+//==================================================================================================
+
+#[unsafe(no_mangle)]
+pub fn main() -> Result<(), Error> {
+    const ITERATIONS_STEP: u64 = 500_000_000;
+
+    for i in 0..u64::MAX {
+        if i % ITERATIONS_STEP == 0 {
```
ppenna, line 37: Why do we need this condition? We should keep examples and tests the simplest possible.

src/user/snapshot-rust-nostd/src/main.rs
```diff
@@ -0,0 +1,44 @@
...
+#![no_main]
+
+//==================================================================================================
+// Imports
+//==================================================================================================
+
+extern crate alloc;
+extern crate nvx;
+
+use alloc::{
```
ppenna, line 19: Please prefix with `::`

src/user/snapshot-rust-nostd/src/main.rs
```diff
@@ -0,0 +1,44 @@
...
+    format,
+    string::String,
+};
+use sys::{
```
ppenna, line 23: Please prefix with `::`

ppenna (changes requested):
I left a minor comment.

src/user/snapshot-rust-nostd/src/main.rs
LucasGdosR: It slows down the printing, making it easier to read the output. I could use `sleep()` instead, if it's available in no_std and you find it simpler.

ppenna: Yes, I imagined this was the motivation.  This should not be a requirement for testing the snapshot functionality in an automated way. The sleep would slightly improve, but it is not realiable 

I also realized that this app is not integrated with the testing infrastructure.

We should rethink how this test should be.

LucasGdosR: This is a manual utility program for testing during development. It is not meant to be automated.

An automated test might be:
- A long running loop (maybe ~1s)
- A pause in the middle of the loop (maybe ~0.5s)
- A resume issued ~1s after the pause
- A check that "Hello, world!" is printed  between ~1.3s and ~1.8s

This would ensure that:
- It actually paused (or it would print earlier than the lower bound)
- It actually resumed execution rather than starting fresh (or it would finish at roughly ~2.5s)
```
    __________________________________________________________________________________________
    |                 |               |                  |                 |                 |
  START             PAUSE             |                RESUME            FINISH       NO-RESUME FINISH
                                      |
                                  NO-PAUSE FINISH
```

ppenna: The automated utility should not depend on timing requirements, as it is not portable.

Let's discuss alternatives offline.

LucasGdosR: Should we have this manual test alternative in the repository, or only the automated utility?

ppenna: We should not have manual tests committed to the repository because:
- No one will run them consistently
- It's extra code to maintain
- They may not be reliable

This does not mean that we don't need a testing program for the snapshot functionality.

Finally, we could turn this into an `example`, but there are simpler ways to demo this.

LucasGdosR:
I'll reopen this when we're thinking about implementing automatic tests for snapshots.

Issue #937, opened 2025-09-04
[utils] Cold-Start L2 Benchmark Fails For MicroVM


ppenna:
## Description

Cold-start benchmarks with the L2 VM are failing intermittently.

We should investigate and fix the issue.

ppenna:
https://github.com/nanvix/nanvix/actions/runs/17532584149/job/49791462670?pr=954

csegarragonz:
@ppenna note that in this case it is the `warm-start-l2` benchmark that is failing.

Now, however, we at least have an error message that we can work with:

```
Error restoring VM: VmRestore(Restore(MigrateSend(Destination is not a directory)))
```

PR #938, opened 2025-09-04
[ci] Disabling Benchmarks with L2 VM

ppenna:
This pull request is a workaround for https://github.com/nanvix/nanvix/issues/937

### DIFF
--- a/.github/actions/benchmark/action.yml
+++ b/.github/actions/benchmark/action.yml
@@ -34,6 +34,10 @@ runs:
       )
 
       for bench in "${benchmarks[@]}"; do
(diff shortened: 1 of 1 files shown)


csegarragonz (commented):
Is the `warm-start-l2` also failing? It does not seem so. I would personally prefer to only skip the cold-start-l2 for microvm by adding an if statement in `.github/actions/benchmark/action.yml` and adding a FIXME linking to the issue.

This way it is more clear that there is a bug to be fixed.

csegarragonz (approved):

PR #950, opened 2025-09-05
[daemons] Use A Separate Gateway Address Per User VM

### LINKED ISSUE #892: [scripts] Kill `nanvixd` On Test Script
## Description

We should kill nanvixd when we terminate test script.

csegarragonz:
In this PR I make each different user VM gateway bind to a different address.

Before deployment, `nanvixd` decides on the gateway address, as well as the socket type, and passes both as arguments to the user VM during start-up. Then the user VM passes this as part of the handshake with the system VM once they have established a connection. The user VM then blindly tries to listen to that address.

This is relatively straightforward for UNIX sockets, as we can just use the sandbox ID as part of the socket name, but becomes a bit more involved for TCP sockets, in L2 deployments, as we need to keep track of the available/used ports. For the moment I implement a very simple mechanism that takes a range as part of the build configuration, and then assigns ports from that range. Note that, for each gateway, linuxd binds to `GUEST_TAP_IP:PORT_FROM_RANGE` inside the L2 VM, so we should not need to worry about port clashes with other services in the host system.

It may now happen that the gateway socket is not immediately available after spawning the VM, so I modify `scripts/test-nanvixd.sh` to wait in a loop. In the process, I fix #892.

### DIFF
--- a/build/linuxd_config.toml
+++ b/build/linuxd_config.toml
@@ -13,4 +13,11 @@ snapshot_name = "l2-sysvm-snapshot"
 # Ports used for different services in linuxd.
 control_plane_port = 9000
 user_vm_port = 9001
-gateway_port = 9002
+
+# Each user VM instance gets a different gateway port, so we define the allowed port range
+# here. Note that this range limits the total number of concurrent user VMs we can have
+# at any point in time. We pick large values that are unlikely to overlap with other
+# services, and with the ephemeral port range used in Linux, readable with:
+# cat /proc/sys/net/ipv4/ip_local_port_range
+gateway_port_range_begin = 49152
+gateway_port_range_end = 65535
--- a/doc/run.md
+++ b/doc/run.md
@@ -101,10 +101,10 @@ nc -lU /tmp/control-plane.socket
 in another terminal, start linuxd:
 
 ```bash
-./bin/linuxd.elf -control-plane-addr /tmp/control-plane.socket -user-vm-bind-addr /tmp/user-vm-bind.socket -gateway-bind-addr /tmp/gw-bind.socket
+./bin/linuxd.elf -control-plane-addr /tmp/control-plane.socket -user-vm-bind-addr /tmp/user-vm-bind.socket
 ```
 
-all `-x-addr` flags have a corresponding `-x-socket-type` flag to switch between `tcp` or `unix` sockets. The default values are `unix`. The `-gateway-bind-addr` is optional, and should only be used if you want to connect to the VM's stdin/stdout.
+all `-x-addr` flags have a corresponding `-x-socket-type` flag to switch between `tcp` or `unix` sockets. The default values are `unix`.
 
 To enable logging, consider prepending the above command with `RUST_LOG=debug` or `RUST_LOG=trace`.
 
@@ -113,16 +113,16 @@ To enable logging, consider prepending the above command with `RUST_LOG=debug` o
 Open another terminal to run the MicroVM. Use the `-initrd` option to specify which application to run, and pass it additional arguments with `-initrd-args`.
 
 ```bash
-./bin/microvm.elf -user-vm-id 1 -system-vm-addr /tmp/user-vm-bind.socket -kernel bin/kernel.elf -initrd bin/hello-rust-nostd.elf [-initrd-args <args>]
+./bin/microvm.elf -user-vm-id 1 -system-vm-addr /tmp/user-vm-bind.socket -kernel bin/kernel.elf -initrd bin/hello-rust-nostd.elf [-gateway-addr /tmp/gw.sock] [-initrd-args <args>]
 ```
 
-If you passed a `-gateway-bind-addr` flag to `linuxd` in step 1, you will need to open a netcat session to connect to it:
+The `-gateway-addr` argument is optional, and should only be used if you want to connect to the VM's stdin/stdout. If you do set it, you next need to open a netcat session to connect to it:
 
 ```bash
-nc -U /tmp/gw-bind.socket
+nc -U /tmp/gw.socket
 ```
 
-For example, the `./bin/echo-c.elf` binary uses the gateway. Inside the netcat terminal, once you have written your message press `Ctrl-D` so that it is flushed to linuxd.
+For example, the `./bin/echo-c.elf` binary uses the gateway because it reads input from stdin. Inside the netcat terminal, once you have written your message press `Ctrl-D` so that it is flushed to linuxd.
 
 Alternatively to an interactive netcat session, you can use something like the following:
 
--- a/scripts/test-nanvixd.sh
+++ b/scripts/test-nanvixd.sh
@@ -2,6 +2,23 @@
 
 set -euo pipefail
 
+#===================================================================================================
+# Global Variables
+#===================================================================================================
+
+NANVIX_HOME=$(git rev-parse --show-toplevel)
+
+#===================================================================================================
+# Utilities
+#===================================================================================================
+
+source "${NANVIX_HOME}/scripts/common/logging.sh"
+source "${NANVIX_HOME}/scripts/common/utils.sh"
+
+#===================================================================================================
+# Command line arguments
+#===================================================================================================
+
 NANVIXD_SOCKADDR=$1
 PROGRAM_NAME=$2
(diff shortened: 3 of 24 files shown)


build/linuxd_config.toml
```diff
@@ -13,4 +13,11 @@ snapshot_name = "l2-sysvm-snapshot"
...
 user_vm_port = 9001
-gateway_port = 9002
+
+# Each user VM instance gets a different gateway port, so we define the allowed port range
+# here. Note that this range limits the total number of concurrent user VMs we can have
+# at any point in time. We pick large values that are unlikely to overlap with other
+# services, and with the ephemral port range used in Linux, readable with:
+# cat /proc/sys/net/ipv4/ip_local_port_range
+gateway_port_range_begin = 61000
+gateway_port_range_end = 65000
```
ppenna: This can technically be any port in the range 49152–65535.

doc/run.md
```diff
@@ -113,16 +113,15 @@ To enable logging, consider prepending the above command with `RUST_LOG=debug` o
 Open another terminal to run the MicroVM. Use the `-initrd` option to specify which application to run, and pass it additional arguments with `-initrd-args`.
 
 ```bash
-./bin/microvm.elf -user-vm-id 1 -system-vm-addr /tmp/user-vm-bind.socket -kernel bin/kernel.elf -initrd bin/hello-rust-nostd.elf [-initrd-args <args>]
+./bin/microvm.elf -user-vm-id 1 -system-vm-addr /tmp/user-vm-bind.socket -kernel bin/kernel.elf -initrd bin/hello-rust-nostd.elf [-gateway-addr /tmp/gw.sock] [-initrd-args <args>]
 ```
-
```
ppenna, line 118: Please rollback this change. We should have a newline here.

doc/run.md
```diff
@@ -113,16 +113,15 @@ To enable logging, consider prepending the above command with `RUST_LOG=debug` o
...
-If you passed a `-gateway-bind-addr` flag to `linuxd` in step 1, you will need to open a netcat session to connect to it:
+ The `-gateway-addr` argument is optional, and should only be used if you want to connect to the VM's stdin/stdout. If you do set it, you next need to open a netcat session to connect to it:
```
ppenna: Remove extra space in the beginning of this line.

scripts/test-nanvixd.sh
```diff
@@ -11,55 +28,100 @@ TIMEOUT=${6:-90}
...
-done
 
-# Check again after waiting.
-if ! ss -tln | grep -q ":${NANVIXD_PORT_NUMBER} "; then
-    echo "nanvixd failed to start"
-    exit 2 # Error Code 2: No such file or directory (ENOENT)
-fi
+    rm -rf "${TMP_DIR}"
+}
+trap 'cleanup' EXIT
```
ppenna, line 117: This function is not invoked. Are we missing that?

ppenna (changes requested):

ppenna:
@csegarragonz the CI error on the benchmark is intermittent.

src/utils/nanvixd/src/sandbox/microvm.rs
```diff
@@ -109,7 +113,11 @@ impl Microvm {
                 anyhow::anyhow!(reason)
             })?;
 
-            Some(port_str.parse()?)
+            get_tcp_port_allocator()
```
csegarragonz: @ppenna why do we call `try_allocate()` here, and also `allocate()` in `src/utils/nanvixd/src/http.rs`? Are we not allocating two different ports?

csegarragonz: We should re-use the latter, and maybe split the gateway host/gateway port in two separate args?

src/utils/nanvixd/src/http.rs
```diff
@@ -110,7 +111,14 @@ impl HttpClient {
         let mut locked_sandbox_cache = sandbox_cache.lock().await;
 
         let gateway_l2_port: Option<TcpPort> = if in_l2 {
-            Some(locked_sandbox_cache.get_available_port()?)
+            match get_tcp_port_allocator().lock().await.allocate().await {
```
csegarragonz, line 115: @ppenna I have a concern here regarding locking. We are already taking a lock on the sandbox_cache in line 111, and then we are taking another lock inside. I am worried this could cause a dead-lock if something in the allocation logic hangs?

src/utils/nanvixd/src/sandbox/microvm.rs
ppenna: In your original code, you were trying to allocate a port number based on the port number string that you received. This code does the same thing.

src/utils/nanvixd/src/http.rs
ppenna: It will not deadlock in respect to `get_tcp_port_allocator()`, because we are grabbing locks in order. But we can also move the port allocation logic before the logic that locks the sandbox cache.

build/linuxd_config.toml
csegarragonz: Thanks, applied.

scripts/test-nanvixd.sh
csegarragonz: The function is always called on `exit *` because we set-up the `trap`. We were doing the same before to `rm -rf "${TMP_DIR}"`, I extend it to also clean-up nanvixd if it is dangling.

src/daemons/linuxd/src/linuxd/mod.rs
```diff
@@ -233,31 +229,60 @@ impl LinuxDaemon {
...
+                                // only take a &str, and we want a formatted String with the error
+                                // message.
+                                error!(
+                                    "error binding to gateway socket (addr={gateway_sockaddr}, \
+                                     error={e:?})"
+                                );
+                                return Err(Self::log_and_error(
+                                    ErrorCode::IoErr,
+                                    "failed to bind gateway socket for user VM",
+                                ));
```
ppenna: You can have a `&static str` for the common string "failed to bind gateway socker for user VM" and use it in both statements.

src/utils/nanvixd/src/sandbox/config.rs
```diff
@@ -1,14 +1,15 @@
...
+use crate::sandbox::tcp_port::TcpPort;
 use hwloc::HwLoc;
 
 //==================================================================================================
 // Structures
 //==================================================================================================
 
 /// Packs configuration for a sandbox.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
```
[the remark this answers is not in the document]
csegarragonz: This is intentional.

src/utils/nanvixd/src/cache/mod.rs
```diff
@@ -103,7 +103,7 @@ impl SandboxCache {
     pub async fn get(
         &mut self,
         tag: &SandboxTag,
-        config: Option<&SandboxConfig>,
+        config: Option<SandboxConfig>,
```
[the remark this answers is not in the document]
csegarragonz: This is intentional, and I have added a comment where needed.

src/utils/nanvixd/src/sandbox/tcp_port.rs
```diff
@@ -0,0 +1,175 @@
...
+            ports.push(port);
+        }
+
+        Self {
+            inner: TcpPortAllocatorInner {
+                ports: Arc::new(Mutex::new(ports)),
+            },
+        }
+    }
+
```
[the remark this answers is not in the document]
csegarragonz: It already has a block comment.

src/utils/nanvixd/src/sandbox/tcp_port.rs
```diff
@@ -0,0 +1,175 @@
...
+    ///
+    /// # Description
+    ///
+    /// Allocate a free TCP port from the pool.
+    ///
+    /// # Returns
+    ///
+    /// A TCP port if there are any in the pool, None otherwise.
+    ///
+    async fn allocate(&mut self) -> Option<RawTcpPortNum> {
```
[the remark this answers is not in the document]
csegarragonz: It does modify the internal structure so I guess it is better to keep the mut.

PR #960, opened 2025-09-06
[microvm] E: Reusing channel instead of barrier and atomic

### LINKED ISSUE #959: [microvm] Reuse vCPU channel instead of using atomic + barrier to get thread id
## Context

The vCPU thread id is stored in an atomic variable shared by two threads and synced via a barrier. Meanwhile, now we have a channel between those threads which could be used to pass the tid in a more straightforward manner.

## Proposed Solution

Pass the tid through the channel instead of using an atomic variable and a barrier.

LucasGdosR:
A `pthread_id_holder` was created when there was no channel between the main thread and the vCPU thread. Now that we have  a channel, we could use it to pass the thread id instead of creating an atomic variable and using a barrier.

Closes https://github.com/nanvix/nanvix/issues/959

### DIFF
--- a/src/microvm/src/orchestrator.rs
+++ b/src/microvm/src/orchestrator.rs
@@ -128,9 +128,10 @@ pub enum VcpuControlCommand {
 ///
 /// Control plane command responses from the vCPU thread to the VMM.
 ///
-#[derive(PartialEq)]
+#[derive(Debug, PartialEq)]
 pub enum VcpuControlResponse {
     Paused,
+    Tid(u64),
 }
 
 //==================================================================================================
@@ -294,6 +295,10 @@ impl Orchestrator {
         loop {
             match self.vcpu_control_rx.try_recv() {
                 Ok(VcpuControlResponse::Paused) => break,
+                Ok(VcpuControlResponse::Tid(tid)) => unreachable!(
+                    "The tid is only sent when the vCPU thread is spawned. Sending it in the \
+                     middle of a pause protocol is against the protocol. tid=({tid})"
+                ),
                 Err(TryRecvError::Empty) => (),
                 Err(TryRecvError::Disconnected) => {
                     let reason: String = "the vmm has disconnected".to_string();
--- a/src/microvm/src/vmm/hyperlight/mod.rs
+++ b/src/microvm/src/vmm/hyperlight/mod.rs
@@ -48,13 +48,8 @@ use ::std::{
     io::Write,
     sync::{
         Arc,
-        Barrier,
         Mutex,
         OnceLock,
-        atomic::{
-            AtomicUsize,
-            Ordering,
-        },
         mpsc::{
             self,
             Sender,
@@ -297,23 +292,16 @@ impl Vmm {
             resume_microvm,
         );
 
-        // We use an atomic to pass the id of the created thread back to the caller context. We
-        // need this because std::thread's JoinHandle does not expose the tid. We synchronize the
-        // update using a barrier.
-        let pthread_id_holder: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
-        let barrier: Arc<Barrier> = Arc::new(Barrier::new(2));
-
-        let barrier_clone: Arc<Barrier> = Arc::clone(&barrier);
-        let pthread_id_holder_clone: Arc<AtomicUsize> = pthread_id_holder.clone();
         let vcpu_thread: JoinHandle<Result<u16>> = std::thread::spawn(move || {
             // Store the tid so that the caller can send signals to the vCPU thread.
             // SAFETY: we are calling pthread_self() right after creating the thread so this is
             // safe.
             let pthread_id: libc::pthread_t = unsafe { pthread_self() };
-            pthread_id_holder_clone.store(pthread_id as usize, Ordering::Relaxed);
-
-            // Notify the outside thread that the thread id is ready.
-            barrier_clone.wait();
+            if let Err(e) = vcpu_thread_control_tx.send(VcpuControlResponse::Tid(pthread_id)) {
+                let reason: String = format!("failed to send the tid (error={e:?})");
+                error!("spawn(): {reason}");
+                anyhow::bail!(reason)
+            }
 
             match sandbox.evolve() {
                 Ok(res) => anyhow::bail!("Expected DEFAULT_VMM_SHUTDOWN_CMD, got: {:#?}", res),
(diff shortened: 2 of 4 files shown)


src/microvm/src/orchestrator.rs
```diff
@@ -294,6 +295,10 @@ impl Orchestrator {
         loop {
             match self.vcpu_control_rx.try_recv() {
                 Ok(VcpuControlResponse::Paused) => break,
+                Ok(VcpuControlResponse::Tid(_)) => unreachable!(
```
ppenna: (nitpick) Please match with `(tid)` and print `tid` on the unreachable message. It will help on debugging if we hit this eventually.

src/microvm/src/vmm/hyperlight/mod.rs
```diff
@@ -297,23 +292,13 @@ impl Vmm {
...
 
-        // We use an atomic to pass the id of the created thread back to the caller context. We
-        // need this because std::thread's JoinHandle does not expose the tid. We synchronize the
-        // update using a barrier.
-        let pthread_id_holder: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
-        let barrier: Arc<Barrier> = Arc::new(Barrier::new(2));
-
-        let barrier_clone: Arc<Barrier> = Arc::clone(&barrier);
-        let pthread_id_holder_clone: Arc<AtomicUsize> = pthread_id_holder.clone();
+        let _pthread_id_holder: u64;
```
ppenna: You don't need this.

src/microvm/src/vmm/hyperlight/mod.rs
```diff
@@ -333,7 +318,9 @@ impl Vmm {
         });
         // Wait right after spawning the vCPU thread such that we populate the pthread id holder
         // before actually starting the vCPU.
-        barrier.wait();
+        if let VcpuControlResponse::Tid(tid) = vcpu_control_rx.recv()? {
+            _pthread_id_holder = tid;
```
ppenna: You can replace this by a trace statement.

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -221,34 +216,25 @@ impl Vmm {
             },
         );
 
-        // We use an atomic to pass the id of the created thread back to the caller context. We
-        // need this because std::thread's JoinHandle does not expose the tid. We synchronize the
-        // update using a barrier.
-        let pthread_id_holder: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
-        let barrier: Arc<Barrier> = Arc::new(Barrier::new(2));
-
+        let _pthread_id_holder: u64;
```
ppenna: You don't need this.

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -221,34 +216,25 @@ impl Vmm {
...
+                .map_err(|e| anyhow::anyhow!("failed to acquire lock {e:?}"))?;
+            locked_vm.send_tid(pthread_id)?;
+            locked_vm.run()
         });
 
         // Wait right after spawning the vCPU thread such that we populate the pthread id holder
         // before actually starting the vCPU.
-        barrier.wait();
+        if let VcpuControlResponse::Tid(tid) = vcpu_control_rx.recv()? {
+            _pthread_id_holder = tid;
```
ppenna: You can replace this with a trace statement.

src/microvm/src/vmm/hyperlight/mod.rs
```diff
@@ -333,7 +318,9 @@ impl Vmm {
         });
         // Wait right after spawning the vCPU thread such that we populate the pthread id holder
         // before actually starting the vCPU.
-        barrier.wait();
+        if let VcpuControlResponse::Tid(tid) = vcpu_control_rx.recv()? {
```
ppenna: Please log an error before failing.

src/microvm/src/vmm/microvm/mod.rs
```diff
@@ -221,34 +216,25 @@ impl Vmm {
...
-            // Notify the outside thread that the thread id is ready.
-            barrier_clone.wait();
-
-            microvm_clone
+            let mut locked_vm = microvm_clone
                 .lock()
-                .map_err(|e| anyhow::anyhow!("failed to acquire lock {e:?}"))?
-                .run()
+                .map_err(|e| anyhow::anyhow!("failed to acquire lock {e:?}"))?;
+            locked_vm.send_tid(pthread_id)?;
```
ppenna: Please log an error before failing.

ppenna (changes requested):

src/microvm/src/vmm/hyperlight/mod.rs
```diff
@@ -297,23 +292,14 @@ impl Vmm {
...
             // SAFETY: we are calling pthread_self() right after creating the thread so this is
             // safe.
             let pthread_id: libc::pthread_t = unsafe { pthread_self() };
-            pthread_id_holder_clone.store(pthread_id as usize, Ordering::Relaxed);
-
-            // Notify the outside thread that the thread id is ready.
-            barrier_clone.wait();
+            vcpu_thread_control_tx
+                .send(VcpuControlResponse::Tid(pthread_id))
+                .map_err(|e| anyhow::anyhow!("failed to send tid {e:?}"))?;
```
ppenna: Now you are just rethrowing the error with a different message.

To log and then rethrow you must use a `if let Err(error)` pattern and inside the if you `error!()` and then `bail!()`

You may create an auxiliary function to do this work (`log_error_and_fail()`) and pass it in as a closure.

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -399,4 +399,22 @@ impl MicroVm {
...
+    /// - `tid`: The vCPU thread's tid.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, returns empty. Otherwise, returns an error.
+    ///
+    pub fn send_tid(&self, tid: u64) -> Result<()> {
+        self.control_tx.send(VcpuControlResponse::Tid(tid))?;
+        Ok(())
+    }
```
ppenna: Why can't this be:

```
    pub fn send_tid(&self, tid: u64) -> Result<()> {
        self.control_tx.send(VcpuControlResponse::Tid(tid))
    }
```

ppenna (changes requested):
We are almost there.

src/microvm/src/vmm/microvm/microvm.rs
LucasGdosR: It can't be because anyhow::Error and SendError are different types. I can change the function's return type and keep it like that, though.

ppenna (approved):

PR #988, opened 2025-09-12
[microvm] Create and load MicroVM snapshots

### LINKED ISSUE #947: [microvm] Create MicroVM snapshots with no linuxd interaction
## Context

Given a paused VMM, it should be possible to save its state and terminate it to stop consuming resources, and then resume execution wherever and whenever is desired.

## Proposed Solution

After https://github.com/nanvix/nanvix/issues/791 is done, we should add functionality to take a VMM in the paused state and save all of its relevant state (vmem and vcpu mostly) to a file. linuxd state would not get persisted at this point.

### LINKED ISSUE #948: [microvm] Load MicroVM snapshots with no linuxd interaction
## Context

Given the capability of creating snapshots of MicroVMs, we should add functionality to load those snapshots and continue execution from where it left off.

## Proposed Solution

After https://github.com/nanvix/nanvix/issues/947 is done, we should add functionality to create a MicroVM and put the state that was persisted in files directly into its vmem and vcpu, then run it from where it left off.

LucasGdosR:
# Overview

The `Vmm` passes callback funtions to the `Orchestrator` so it calls them to create or load snapshots when it needs to. These closures are implemented by the `MicroVm`, which uses both the `VirtualProcessor` and the `VirtualMemory` to gather state and persist it, or reads from files and writes to those fields.

# Details

- How to name the snapshots? For now, I just name them by the executable name and the `.vmem` and `.kvm.json` extensions, and put them in a `snapshots` directory in the repository's root;
- On the subject of naming, `make_snapshot_paths` might need to be changed based on that;
- `kvm_fpu` does not support serialization through `serde`, so I implemented serialization and deserialization manually;
- I created a custom header for the virtual memory snapshot file. It needs to be improved upon;
- I'm not sure whether loading a snapshot to a `MicroVm` with a larger memory size should be an error;
- I used `serde-json` for serializing the vCPU. This is fine for a PoC, but should be changed for production, I think.

Closes https://github.com/nanvix/nanvix/issues/947 and https://github.com/nanvix/nanvix/issues/948.

### DIFF
--- a/src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
+++ b/src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
@@ -13,6 +13,14 @@ mod timer;
 // Exports
 //==================================================================================================
 
+pub use exit::*;
+
+//==================================================================================================
+// Imports
+//==================================================================================================
+
+use crate::vmm::microvm::kvm::partition::VirtualPartition;
+use ::anyhow::Result;
 use ::arch::cpu::{
     cpuid::{
         CPUID_FEATURES,
@@ -30,22 +38,16 @@ use ::kvm_bindings::{
     CpuId,
     KVM_MAX_CPUID_ENTRIES,
     kvm_fpu,
-};
-pub use exit::*;
-
-//==================================================================================================
-// Imports
-//==================================================================================================
-
-use crate::vmm::microvm::kvm::partition::VirtualPartition;
-use ::anyhow::Result;
-use ::kvm_bindings::{
     kvm_regs,
     kvm_sregs,
 };
 use ::kvm_ioctls::{
     VcpuExit,
     VcpuFd,
+};
+use ::serde::{
+    Deserialize,
+    Serialize,
 };
 use ::std::sync::{
     Arc,
@@ -82,6 +84,25 @@ pub struct VirtualProcessor {
     /// Handle to timer.
     _timer: Timer,
     /// Processor state.
+    online: bool,
+    /// Exit status code.
+    exit_status: u16,
+}
+
+///
+/// # Description
+///
+/// Virtual CPU state that can be serialized and saved to disk.
+///
+#[derive(Serialize, Deserialize)]
+pub struct VirtualProcessorState {
+    /// General purpose registers.
+    regs: kvm_regs,
+    /// System registers (segment registers, control registers, etc.).
+    sregs: kvm_sregs,
+    /// FPU/SIMD state.
+    fpu: Vec<u8>,
+    /// Whether the processor is online.
     online: bool,
     /// Exit status code.
     exit_status: u16,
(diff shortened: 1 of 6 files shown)


src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -87,6 +89,25 @@ pub struct VirtualProcessor {
...
+    pub regs: kvm_regs,
+    /// System registers (segment registers, control registers, etc.).
+    pub sregs: kvm_sregs,
+    /// FPU/SIMD state.
+    pub fpu: Vec<u8>,
+    /// Whether the processor is online.
+    pub online: bool,
+    /// Exit status code.
+    pub exit_status: u16,
+}
```
ppenna: Structures should not have public member fields.

Create helper functions to get/set values and apply validations.

src/microvm/src/vmm/microvm/kvm/vmem.rs
```diff
@@ -386,6 +393,179 @@ impl VirtualMemory {
...
+        // Write snapshot header with metadata
+        // Header format (64 bytes total):
+        // - Memory size (8 bytes): u64
+        // - Kernel base (8 bytes): u64
+        // - Kernel size (8 bytes): u64
+        // - Initrd base (8 bytes): u64
+        // - Initrd size (8 bytes): u64
+        // - Credits (4 bytes): u32
+        // - Padding (20 bytes): zeros
+        // - Memory contents (size bytes)
```
ppenna: Place each comment before each concerned statement. It is very easy to get this unsynced.

ppenna (changes requested):
- Log error before failing in all `map_err()` statements. Use `log_and_error()` function idea to do that.
- Remember to save/load CPUID table.
- Remember to save/load PIT configuration.
- Check if when loading the snapshot has the same memory size as well as same features in the CPUID table

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
LucasGdosR: I just stripped the `pub` qualifications and it works without the getter functions, because it's only accessed in this same module. I suppose this is fine. It's even more encapsulated than creating getters.

src/microvm/src/vmm/microvm/kvm/vmem.rs
LucasGdosR: This is a job for the type system, not for comments. Added a SnapshotHeader struct :)

LucasGdosR:
> * Log error before failing in all `map_err()` statements. Use `log_and_error()` function idea to do that.

I don't get why `map_err()` is not enough in this case. If an error occurs, it will propagate back to `Orchestrator::handle_command()`, which already has an `if let Err(e) = ... { // log and error` construct. Every error is already logged and followed by `bail!()`. These are leaf functions which can very well propagate their error messages to a handler function.

> * Remember to save/load CPUID table.
> * Remember to save/load PIT configuration.
> * Check if when loading the snapshot has the same memory size as well as same features in the CPUID table

I will figure out how CPUID table and PIT configuration work and add them to the `VirtualProcessorState` struct.

LucasGdosR:
> I will figure out how CPUID table and PIT configuration work and add them to the `VirtualProcessorState` struct.

I went looking through the `kvm_ioctls::VcpuFd` [documentation](https://docs.rs/kvm-ioctls/latest/kvm_ioctls/struct.VcpuFd.html) and I found a bunch of extra state on top of the CPUID table. Do we need to keep them all? Some of it is X86-only. Namely, here it is:

- LAPIC (Local Advanced Programmable Interrupt Controller)
- MSRS (Model-Specific RegisterS)
- MP_STATE (MultiProcessing State)
- VCPU_EVENTS (pending exceptions, interrupts, NMIs, and related states)
- TSC_KHZ (TSC frequency)
- SYNC_REGS (kvm_sync_regs)
- NESTED_STATE
x86 only:
- XSAVE and XSAVE2 (xsave struct)
- XCRS (xcrs)
- DEBUGREGS (debug registers)

I looked through [VmFd](https://docs.rs/kvm-ioctls/latest/kvm_ioctls/struct.VmFd.html) and found that I can `get_pit2` from it. In the codebase, I can access it through `VirtualProcessor._partition.lock()?.vm()`. However, there's also:
- CLOCK
x86 only:
- IRQCHIP

Should I include these too?

ppenna:
> I don't get why `map_err()` is not enough in this case. If an error occurs, it will propagate back to `Orchestrator::handle_command()`, which already has an `if let Err(e) = ... { // log and error` construct. Every error is already logged and followed by `bail!()`. These are leaf functions which can very well propagate their error messages to a handler function.

Yes, you have to:
- Logging errors on leaf functions make it possible to pin-point the precise line where the bug happened.
- Messages in `bail!()` do not end up in log files, thus are not included in telemetry data.
- This is a production requirement.
- This is a project standard.
- This is the most predominant programming pattern in the source code and following it is a golden rule in software engineering.

ppenna:
> > I will figure out how CPUID table and PIT configuration work and add them to the `VirtualProcessorState` struct.
> 
> I went looking through the `kvm_ioctls::VcpuFd` [documentation](https://docs.rs/kvm-ioctls/latest/kvm_ioctls/struct.VcpuFd.html) and I found a bunch of extra state on top of the CPUID table. Do we need to keep them all? Some of it is X86-only. Namely, here it is:
> 
> * LAPIC (Local Advanced Programmable Interrupt Controller)
> * MSRS (Model-Specific RegisterS)
> * MP_STATE (MultiProcessing State)
> * VCPU_EVENTS (pending exceptions, interrupts, NMIs, and related states)
> * TSC_KHZ (TSC frequency)
> * SYNC_REGS (kvm_sync_regs)
> * NESTED_STATE
>   x86 only:
> * XSAVE and XSAVE2 (xsave struct)
> * XCRS (xcrs)
> * DEBUGREGS (debug registers)
> 
> I looked through [VmFd](https://docs.rs/kvm-ioctls/latest/kvm_ioctls/struct.VmFd.html) and found that I can `get_pit2` from it. In the codebase, I can access it through `VirtualProcessor._partition.lock()?.vm()`. However, there's also:
> 
> * CLOCK
>   x86 only:
> * IRQCHIP
> 
> Should I include these too?

You must save and restore every configurable parameter of the running VM, otherwise the snapshot may be potentially out of sync with the guest state after restored.

src/microvm/src/vmm/microvm/kvm/vmem.rs
```diff
@@ -45,14 +52,43 @@ pub struct VirtualMemory {
...
+    /// Kernel size (8 bytes): usize
+    kernel_size: usize,
+    /// Initrd base (8 bytes): u64
+    initrd_base: u64,
+    /// Initrd size (8 bytes): usize
+    initrd_size: usize,
+    /// Credits (4 bytes): u32
+    credits: u32,
+    /// Padding (20 bytes): zeros
+    padding: [u8; 20],
```
ppenna, line 80: Create a private constant for this in `Self`. Also, document ther eason why this padding is required. If it is to make the structure 64-bytes long, consider adding a static check for that.

src/microvm/src/vmm/microvm/kvm/vmem.rs
```diff
@@ -386,6 +422,116 @@ impl VirtualMemory {
...
+        let memory_slice: &[u8] = unsafe { slice::from_raw_parts(self.ptr, self.size) };
+        file.write_all(memory_slice)
+            .map_err(|e| anyhow::anyhow!("failed to write memory contents: {}", e))?;
+
+        file.sync_all()
+            .map_err(|e| anyhow::anyhow!("failed to sync snapshot file: {}", e))?;
+
+        trace!("save_snapshot(): successfully saved snapshot to {:?}", path);
+        Ok(())
+    }
```
ppenna, line 479: Log before failing. Always. Anywhere.

src/microvm/src/vmm/microvm/kvm/vmem.rs
```diff
@@ -386,6 +422,116 @@ impl VirtualMemory {
...
+        if header.memory_size != self.size {
+            return Err(anyhow::anyhow!(
+                "memory size mismatch: snapshot has {} bytes, current VM has {} bytes",
+                header.memory_size,
+                self.size
+            ));
+        }
+
+        // Read memory contents
+        let memory_slice: &mut [u8] = unsafe { slice::from_raw_parts_mut(self.ptr, self.size) };
```
ppenna, line 523: Please comment why is this safe. Also introduce checks alignment checks.

src/microvm/src/vmm/microvm/kvm/vmem.rs
```diff
@@ -398,3 +544,52 @@ impl Drop for VirtualMemory {
...
+    ///
+    /// Serializes the snapshot header (which has `repr(C)`) as a slice of bytes.
+    ///
+    /// # Returns
+    ///
+    /// A slice of bytes containing the snapshot header.
+    ///
+    fn as_bytes(&self) -> &[u8; SIZE_OF_HEADER] {
+        // SAFETY: The size of the slice is the same of the struct.
+        unsafe { &*(self as *const SnapshotHeader as *const [u8; SIZE_OF_HEADER]) }
```
ppenna, line 579: Consider using memory transmute instead.

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -480,4 +490,104 @@ impl MicroVm {
...
+    /// # Parameters
+    ///
+    /// - `filepath`: Path to the executable file.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, returns empty. Otherwise, returns an error.
+    ///
+    pub fn load_snapshot(&mut self, filepath: &str) -> Result<()> {
+        let (vmem_filepath, kvm_filepath) = Self::make_snapshot_paths(filepath);
```
ppenna, line 577: Add missing type annotation.

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -480,4 +490,104 @@ impl MicroVm {
...
+    /// # Parameters
+    ///
+    /// - `filepath`: Path to the executable file.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, returns empty. Otherwise, returns an error.
+    ///
+    pub fn create_snapshot(&self, filepath: &str) -> Result<()> {
+        let (vmem_filepath, kvm_filepath) = Self::make_snapshot_paths(filepath);
```
ppenna, line 542: Add missing type annotation.

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -52,6 +58,10 @@ use ::std::{
     },
     time::Instant,
 };
+use std::path::{
+    Path,
+    PathBuf,
+};
```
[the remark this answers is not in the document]
ppenna: +1

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -381,4 +402,145 @@ impl VirtualProcessor {
...
+    /// # Description
+    ///
+    /// Serializes `kvm_fpu` and creates a new `VirtualProcessorState`.
+    ///
+    /// # Parameters
+    ///
+    /// - `regs`: General purpose registers.
+    /// - `sregs`: System registers (segment registers, control registers, etc.).
+    /// - `fpu`: FPU/SIMD state.
+    /// - `onlline`: Whether the processor is online.
```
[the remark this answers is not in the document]
ppenna: +1

src/microvm/src/vmm/microvm/kvm/vmem.rs
LucasGdosR: Sorry, I didn't have time to modify these files between your preliminary review and this one. I understood why error logging is crucial. These are fixes I already had locally, I assure you.

src/microvm/src/vmm/microvm/microvm.rs
LucasGdosR: Type annotations cannot be added to multiple return values.

src/microvm/src/vmm/microvm/microvm.rs
LucasGdosR: Type annotations cannot be added to multiple return values.

LucasGdosR:
I'm splitting up this PR into multiple ones. The first one is https://github.com/nanvix/nanvix/pull/996

Issue #500, opened 2025-03-07
[posix] Support `pthread_condattr_setclock()`


ppenna:
## Description

Implement `pthread_condattr_setclock()`.

## References

https://pubs.opengroup.org/onlinepubs/9799919799/functions/pthread_condattr_getclock.html

Issue #483, opened 2025-03-07
[posix] Support `pthread_atfork()`


ppenna:
## Description

Implement `pthread_atfork()`.

## References

https://pubs.opengroup.org/onlinepubs/9799919799/functions/pthread_atfork.html

PR #990, opened 2025-09-15
[lib] Dummy System Call Implementations

ppenna:
# Description

This PR a dummy implementation for the following system calls:
- `pthread_condattr_setclock()`
- `pthread_atfork()`

This PR is a workaround for the following issues:
- https://github.com/nanvix/nanvix/issues/500
- https://github.com/nanvix/nanvix/issues/483


danbugs (dismissed):
LGTM

PR #996, opened 2025-09-16
[microvm] Create MicroVM memory snapshot

LucasGdosR:
# Overview

This PR's commits are sequenced in this way:
- Adds dependencies and rename variable (mostly boilerplate);
- Adds the plumbing so the main thread can call the snapshot function (a callback);
- High level abstract implementation of snapshots at the `MicroVm` struct level;
- Implementation of memory snapshots at the `VirtualMemory` struct level. TODOs for the `VirtualProcessor` and for loading snapshots. Removes dependencies for loading snapshots.

I'll squash them once they're reviewed.

# Next steps

This leaves snapshots of the `VirtualProcessor` for another PR, and loading snapshots for another one after that.

### DIFF
--- a/src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
+++ b/src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
@@ -13,6 +13,14 @@ mod timer;
 // Exports
 //==================================================================================================
 
+pub use exit::*;
+
+//==================================================================================================
+// Imports
+//==================================================================================================
+
+use crate::vmm::microvm::kvm::partition::VirtualPartition;
+use ::anyhow::Result;
 use ::arch::cpu::{
     cpuid::{
         CPUID_FEATURES,
@@ -30,22 +38,16 @@ use ::kvm_bindings::{
     CpuId,
     KVM_MAX_CPUID_ENTRIES,
     kvm_fpu,
-};
-pub use exit::*;
-
-//==================================================================================================
-// Imports
-//==================================================================================================
-
-use crate::vmm::microvm::kvm::partition::VirtualPartition;
-use ::anyhow::Result;
-use ::kvm_bindings::{
     kvm_regs,
     kvm_sregs,
 };
 use ::kvm_ioctls::{
     VcpuExit,
     VcpuFd,
+};
+use ::serde::{
+    Deserialize,
+    Serialize,
 };
 use ::std::sync::{
     Arc,
@@ -85,6 +87,16 @@ pub struct VirtualProcessor {
     online: bool,
     /// Exit status code.
     exit_status: u16,
+}
+
+///
+/// # Description
+///
+/// Virtual CPU state that can be serialized and saved to disk.
+///
+#[derive(Serialize, Deserialize)]
+pub struct VirtualProcessorState {
+    // TODO: fill struct with relevant state https://github.com/nanvix/nanvix/issues/947
 }
 
 impl VirtualProcessor {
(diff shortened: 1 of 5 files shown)


src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -87,6 +89,16 @@ pub struct VirtualProcessor {
...
 }
 
+///
+/// # Description
+///
+/// Virtual CPU state that can be serialized and saved to disk.
+///
+#[derive(Serialize, Deserialize)]
+pub struct VirtualProcessorState {
+    // TODO: fill struct with relevant state.
```
ppenna: Create an issue and link it here.

src/microvm/src/vmm/microvm/kvm/vmem.rs
```diff
@@ -45,14 +49,45 @@ pub struct VirtualMemory {
...
 unsafe impl Send for VirtualMemory {}
 unsafe impl Sync for VirtualMemory {}
 
+//==================================================================================================
+// Constants
+//==================================================================================================
+
+const SIZE_OF_HEADER: usize = mem::size_of::<SnapshotHeader>();
+/// This will fail to compile if the size is not 64 bytes
+const _: () = assert!(SIZE_OF_HEADER == 64, "SnapshotHeader must be 64 bytes");
```
ppenna: Use our lib `static_assert`. Also add a comment explaining why this is required.

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -480,4 +486,86 @@ impl MicroVm {
...
+        let filename: &str = filepath
+            .strip_suffix(".elf")
+            .unwrap_or(filepath)
+            .strip_prefix("bin/")
+            .unwrap_or(filepath);
+        // NOTE: if a different prefix might occur, we must either:
+        // - Remove all prefixes (use just the file name);
+        // - Create subdirectories in the snapshot directory;
+        // - Replace the `/`s from the path and concatenate them all into the snapshot filename.
+        // Otherwise we'll get errors from writing to a directory that doesn't exist.
```
ppenna: Can you solve this here, or create an issue and link it here?

ppenna (changes requested):
I left some minor comments.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
LucasGdosR: Is it okay if I reference the existing issue on creating snapshots? This is a needed step. https://github.com/nanvix/nanvix/issues/947

src/microvm/src/vmm/microvm/kvm/vmem.rs
```diff
@@ -45,14 +49,45 @@ pub struct VirtualMemory {
     /// Kernel location and size.
     kernel: Option<(u64, usize)>,
     /// Initial RAM disk location and size.
```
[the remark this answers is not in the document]
LucasGdosR: It's the same field, except it's now used, so the trailing underscore was removed. The existing documentation still applies.

src/microvm/src/vmm/microvm/microvm.rs
LucasGdosR: I used `Path::file_stem()` to get the filename.

src/microvm/src/vmm/microvm/kvm/vmem.rs
LucasGdosR: I have added a comment. I'm assuming that 64-byte alignment is enough for the purposes of 512-bit SIMD operations, and that we don't require page alignment. If we do require page alignment, than the header would need enough padding to get to 4kB.

ppenna (approved):

LucasGdosR:
Sorry for the mixup with the squash after turning CI/CD on.

PR #999, opened 2025-09-16
[microvm] MicroVM VirtualProcessor snapshots

LucasGdosR:
# Overview

This PR:
- Fills up a struct with the relevant state for vCPU snapshots
- Implements a getter to populate this struct with state from KVM using the KVM API
- Implements custom serialization for kvm structs that don't have serde support.

### DIFF
--- a/src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
+++ b/src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
@@ -37,21 +37,45 @@ use ::arch::cpu::{
 use ::kvm_bindings::{
     CpuId,
     KVM_MAX_CPUID_ENTRIES,
+    Msrs,
+    Xsave,
+    kvm_clock_data,
+    kvm_debugregs,
     kvm_fpu,
+    kvm_irqchip,
+    kvm_lapic_state,
+    kvm_mp_state,
+    kvm_msr_entry,
+    kvm_pit_state2,
     kvm_regs,
     kvm_sregs,
+    kvm_vcpu_events,
+    kvm_xcrs,
+    kvm_xsave,
 };
 use ::kvm_ioctls::{
+    Cap,
+    Kvm,
     VcpuExit,
     VcpuFd,
+    VmFd,
 };
 use ::serde::{
     Deserialize,
     Serialize,
 };
-use ::std::sync::{
-    Arc,
-    Mutex,
+use ::std::{
+    mem,
+    slice,
+    sync::{
+        Arc,
+        Mutex,
+        MutexGuard,
+    },
+};
+use ::vmm_sys_util::fam::{
+    FamStruct,
+    FamStructWrapper,
 };
 use irqchip::IrqChip;
 use timer::Timer;
@@ -76,7 +100,7 @@ const FP_TAG_WORD_DEFAULT: u8 = 0xff;
 ///
 pub struct VirtualProcessor {
     /// Handle to underlying virtual partition.
-    _partition: Arc<Mutex<VirtualPartition>>,
+    partition: Arc<Mutex<VirtualPartition>>,
     /// Handle to underlying virtual processor.
     fd: VcpuFd,
     /// Handle to underlying interrupt controller.
@@ -94,9 +118,57 @@ pub struct VirtualProcessor {
 ///
 /// Virtual CPU state that can be serialized and saved to disk.
 ///
+/// The virtual CPU is composed of three structs from the `kvm_*` crates:
+/// - `KVM`: A field of `VirtualPartition`.
+/// - `VcpuFd`: A field of `VirtualProcessors`.
+/// - `VmFd`: A field of `VirtualPartition`.
+///
+/// This structure holds the state that can be extracted from `VcpuFd` and `VmFd`, as `KVM` does not
+/// hold state directly. Also, it holds the direct fields of `VirtualProcessor`: `online` and
+/// `exit_status`.
+///
 #[derive(Serialize, Deserialize)]
 pub struct VirtualProcessorState {
-    // TODO: fill struct with relevant state https://github.com/nanvix/nanvix/issues/947
+    // `VirtualProcessor` direct state:
+    /// Whether the processor is online.
+    online: bool,
+    /// Exit status code.
+    exit_status: u16,
+
+    // `VirtualProcessor` indirect state:
+    // VcpuFd state:
+    /// General purpose registers.
+    regs: kvm_regs,
+    /// System registers (segment registers, control registers, etc.).
+    sregs: kvm_sregs,
+    /// FPU/SIMD state. Natively a `kvm_bindings::kvm_fpu`.
+    fpu: Vec<u8>,
+    /// CPUID table. Natively a `kvm_bindings::CpuId`.
+    cpuid: Vec<u8>,
+    /// Local Advanced Programmable Interrupt Controller.
+    lapic: kvm_lapic_state,
+    /// Model-Specific RegisterS. Natively a `kvm_bindings::Msrs`
+    msrs: Vec<u8>,
+    /// MultiProcessing State.
+    mp_state: kvm_mp_state,
+    /// KVM's xsave struct (x86 only). Natively a `kvm_bindings::Xsave`.
+    xsave: Vec<u8>,
+    /// XCRS (x86 only).
+    xcrs: kvm_xcrs,
+    /// Debug registers (x86 only).
+    debugregs: kvm_debugregs,
+    /// Pending exceptions, interrupts, NMIs, and related states.
+    vcpu_events: kvm_vcpu_events,
+    /// TSC frequency in kHz.
+    tsc_khz: u32,
+
+    // VmFd state:
+    /// Interrupt controller.
+    irqchip: kvm_irqchip,
+    /// Timer.
+    pit_state: kvm_pit_state2,
+    /// Timestamp of kvmclock.
+    clock_data: kvm_clock_data,
 }
 
 impl VirtualProcessor {
@@ -132,7 +204,7 @@ impl VirtualProcessor {
         fd.set_fpu(&fpu)?;
 
         Ok(Self {
-            _partition: partition,
+            partition,
             fd,
             _irqchip: irqchip,
             _timer: timer,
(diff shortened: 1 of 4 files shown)


src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -399,14 +468,17 @@ impl VirtualProcessor {
...
+    ///
     /// # Returns
     ///
     /// Upon successful completion, returns the current processor state that can be serialized and
     /// saved to a file. Otherwise, returns an error.
     ///
     pub fn get_state(&self) -> Result<VirtualProcessorState> {
-        // TODO: get virtual processor state https://github.com/nanvix/nanvix/issues/947
-        Ok(VirtualProcessorState {})
+        todo!()
```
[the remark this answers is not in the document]
LucasGdosR: I will fill the `get_state()` method in this PR once the `VirtualProcessorState` is approved. The `todo!()` macro will not be merged.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -94,9 +111,61 @@ pub struct VirtualProcessor {
...
+    /// CPUID table. Natively a `kvm_bindings::CpuId`.
+    cpuid: Vec<u8>,
+    /// Local Advanced Programmable Interrupt Controller.
+    lapic: kvm_lapic_state,
+    /// Model-Specific RegisterS. Natively a `kvm_bindings::Msrs`
+    msrs: Vec<u8>,
+    /// MultiProcessing State.
+    mp_state: kvm_mp_state,
+    /// KVM's xsave struct (x86 only). Natively a `kvm_bindings::Xsave`.
+    xsave: Vec<u8>,
```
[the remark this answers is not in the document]
LucasGdosR: Actually, MSRS is the acronym. I put 'S' to indicate what the MSRS acronym stands for.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -37,21 +37,38 @@ use ::arch::cpu::{
...
-    Arc,
-    Mutex,
+use ::std::{
+    mem,
+    slice,
+    sync::{
+        Arc,
+        Mutex,
+    },
 };
```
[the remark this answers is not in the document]
LucasGdosR: They will be used in this PR. `get_state()` will be implemented in this PR. If `get_state()` doesn't use them, they'll be removed.

src/microvm/src/vmm/microvm/microvm.rs
```diff
@@ -548,11 +547,19 @@ impl MicroVm {
...
-                let json: String = serde_json::to_string(&kvm_state)?;
                 let mut file: File = File::create(kvm_filepath)?;
-                file.write_all(json.as_bytes())?;
-                Ok(())
+                match bincode::serde::encode_into_std_write(
+                    kvm_state,
+                    &mut file,
+                    bincode::config::standard(),
+                ) {
+                    Ok(_) => Ok(()),
```
ppenna: Missing explicity match.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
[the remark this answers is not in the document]
ppenna: Unresolving this so it does not pass unnoticed.

ppenna (dismissed):
Please, address the `todo!()` impl.

src/microvm/src/vmm/microvm/microvm.rs
LucasGdosR: Is replacing `_` with `_n` fine?

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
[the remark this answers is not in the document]
LucasGdosR: This commit addressed the `todo!()` macro.

src/microvm/src/vmm/microvm/microvm.rs
ppenna: Yes, but why not emit a trace message? :)

LucasGdosR:
All changes are in the new commit.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -399,14 +467,228 @@ impl VirtualProcessor {
...
     pub fn get_state(&self) -> Result<VirtualProcessorState> {
-        // TODO: get virtual processor state https://github.com/nanvix/nanvix/issues/947
-        Ok(VirtualProcessorState {})
+        // Plain getters:
+        let regs: kvm_regs = match self.fd.get_regs() {
+            Ok(v) => v,
+            Err(e) => {
+                let reason: String = format!("failed getting kvm_regs (error={e:?})");
+                error!("get_state(): {reason}");
+                anyhow::bail!(reason)
```
csegarragonz, line 493: Any reason why you are bailing-out everywhere instead of returning an Err?

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -399,14 +467,228 @@ impl VirtualProcessor {
...
+            let mut xsave2: Xsave = match Xsave::new(fam_units) {
+                Ok(v) => v,
+                Err(e) => {
+                    let reason: String = format!("failed creating xsave2 (error={e:?})");
+                    error!("get_state(): {reason}");
+                    anyhow::bail!(reason)
+                },
+            };
+            // SAFETY: This is safe because we've checked the number of elements before allocating.
+            unsafe {
```
csegarragonz: What is the `unsafe` call here? I would recommend reduing the `unsafe` scope to the minimum possible.

csegarragonz: Also, I don't see any allocation in your unsafe block.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -427,3 +709,48 @@ impl VirtualProcessor {
         Ok(())
     }
 }
+
+//==================================================================================================
```
csegarragonz, line 749: Why do we need these helper methods? Can we not use `bincode`?

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
LucasGdosR: I should probably read up on `anyhow`. I was told:
> To log and then rethrow you must use a `if let Err(error)` pattern and inside the if you `error!()` and then `bail!()`
https://github.com/nanvix/nanvix/pull/960#discussion_r2333738273

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
LucasGdosR: `get_xsave2()` is unsafe. I could store its return and match on it outside the `unsafe` scope.

> Safety
> This function is unsafe because there is no guarantee xsave is allocated with enough space to hold the entire xsave state.
>
> The required size in bytes can be retrieved via KVM_CHECK_EXTENSION(KVM_CAP_XSAVE2) and can vary depending on features that have been dynamically enabled by arch_prctl(). Thus, any features must not be enabled dynamically after the required size has been confirmed.
>
> If xsave is not large enough, KVM_GET_XSAVE2 copies data beyond the allocated area, possibly causing undefined behavior.

The allocation happens on `Xsave::new(fam_units)`. We calculate what `fam_units` should be with all the previous code, and then allocate enough memory to `xsave2` such that calling `get_save2(&mut save2)` is safe. This is the relevant code to the safety claim:
```
let xsave_total_bytes: usize = kvm.check_extension_int(Cap::Xsave2) as usize;
let header_size: usize = mem::size_of::<kvm_bindings::kvm_xsave2>();
let fam_entries: usize = xsave_total_bytes.saturating_sub(header_size);
let fam_units: usize = fam_entries.div_ceil(mem::size_of::<u32>());
let mut xsave2: Xsave = match Xsave::new(fam_units) {
    Ok(v) => v,
    Err(e) => {
        let reason: String = format!("failed creating xsave2 (error={e:?})");
        error!("get_state(): {reason}");
        anyhow::bail!(reason)
    },
};
```

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
LucasGdosR: I'm not entirely sure on standalone `bincode` with `Encode` and `Decode` using the 2.0+ API. I'm sure we cannot use `serde`, though, or `bincode::serde`. There's no support for `kvm_fpu`, `FamStructWrapper<kvm_xsave2>`, `FamStructWrapper<kvm_cpuid2>`, and `FamStructWrapper<kvm_msrs>`. I think it might have to do with most of them being `FamStructWrapper`s, which are variable length structs. They have members which are arrays with flexible size (Flexible Array Member), so the size is unknown at compile time.

If you have suggestions on how to make this simpler, I'm all ears. I struggled with this part.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
csegarragonz: You are right, my bad. I checked the anyhow docs and I realised that `anyhow::bail!` is equivalent to `return Err(anyhow::anyhow())`.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
csegarragonz: > get_xsave2() is unsafe. I could store its return and match on it outside the unsafe scope.

This is more or less what I was thinking of:

```rust
if let Err(e) = unsafe { self.fd.get_xsave2(&mut xsave2) } { 
    let reason: String = format!("failed getting xsave2 (error={e:?})"); 
    error!("get_state(): {reason}"); 
    anyhow::bail!(reason) 
}
````

File: src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs (hunk 7 above)
Comment (csegarragonz) [reply src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs:749]: I thought a bit more about it and I think you are right. To use `bincode` we would need (at least?) to annotate the raw structures with `Encode/Decode` stuff.

My only suggestion would then be that, if we are serializing a `&[T]` (which I am interpreting is a C-style array), you could probably get away with casting the whole slice at once, instead of doing it element by element:

```rust
fn serialize_slice<T>(data: &[T]) -> Vec<u8> {
    unsafe {
        slice::from_raw_parts(
            data.as_ptr() as *const u8,
            data.len() * mem::size_of::<T>(),
        )
        .to_vec()
    }
}
```

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
LucasGdosR: Nice, thanks a bunch! Pattern matching off of the unsafe block signals clearly what's unsafe while maintaining ergonomics.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
LucasGdosR: You're right, there's no need to iterate through it. I realized I also need to include the header metadata, though, and this only stores the entries. I changed it to something else that stores the header and the array in a single cast. Thanks a lot!

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -94,9 +118,57 @@ pub struct VirtualProcessor {
...
+    /// TSC frequency in kHz.
+    tsc_khz: u32,
+
+    // VmFd state:
+    // Interrupt controller.
+    irqchip: kvm_irqchip,
+    // Timer.
+    pit_state: kvm_pit_state2,
+    // Timestamp of kvmclock.
+    clock_data: kvm_clock_data,
```
ppenna: Fix documentation to have Rust doc syntax.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -427,3 +731,49 @@ impl VirtualProcessor {
...
+///
+/// - `t`: An instance of the T type.
+///
+/// # Returns
+///
+/// A vector of bytes with the same contents as the structure.
+///
+fn serialize_plain<T: Sized>(t: &T) -> Vec<u8> {
+    // We cannot use `transmute`, because "generic parameters may not be used in const operations".
+    //SAFETY: We're casting a `Sized` type to a `&[u8]` of its own length, so the sizes match.
```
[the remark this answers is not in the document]
ppenna: +1

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -427,3 +731,49 @@ impl VirtualProcessor {
...
+///
+/// # Parameters
+///
+/// - `wrapper`: A `FamStructWrapper` with array entries of the T type.
+///
+/// # Returns
+///
+/// A vector of bytes with the same contents as the original wrapper.
+///
+pub fn serialize_fam_struct<T: Default + FamStruct>(
```
ppenna: Does this structure need to be `pub`?

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -399,14 +471,246 @@ impl VirtualProcessor {
...
+            .collect();
+        let mut msrs: Msrs = match Msrs::from_entries(&msr_entries) {
+            Ok(v) => v,
+            Err(e) => {
+                let reason: String = format!("failed creating msrs (error={e:?})");
+                error!("get_state(): {reason}");
+                anyhow::bail!(reason)
+            },
+        };
+        let _nmsrs_read: usize = match self.fd.get_msrs(&mut msrs) {
```
ppenna: Should we sanity check that `_nmsrs_read == msr_entries.len()` ?

ppenna (dismissed):
Please address reviews, amend commits, rebase, push and open PR for regular reviews.

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -402,14 +474,259 @@ impl VirtualProcessor {
...
+                } else {
+                    nmsrs_read
+                }
+            },
+            Err(e) => {
+                let reason: String = format!("failed mutating msrs (error={e:?})");
+                error!("get_state(): {reason}");
+                anyhow::bail!(reason)
+            },
+        };
```
ppenna: As you are not using the return of the match statement, you can rewrite as it:

```rust
match self.fd.get_msrs(&mut msrs) {
  Ok(nmsrs_read) => {
    // Sanity check.
    if nmsrs_read != msr_entries.len() {
      let reason: String = format!(
        "`nmsrs_read`(={}) is different from `msr_entries.len()`(={})",
        nmsrs_read,
        msr_entries.len(),
      );
      error!("get_state(): {reason}");
      anyhow::bail!(reason)
    }
  },
  Err(e) => {
    let reason: String = format!("failed mutating msrs (error={e:?})");
    error!("get_state(): {reason}");
    anyhow::bail!(reason)
  },
};
```

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
```diff
@@ -430,3 +747,49 @@ impl VirtualProcessor {
...
+/// - `t`: An instance of the T type.
+///
+/// # Returns
+///
+/// A vector of bytes with the same contents as the structure.
+///
+fn serialize_plain<T: Sized>(t: &T) -> Vec<u8> {
+    // We cannot use `transmute`, because "generic parameters may not be used in const operations".
+    // SAFETY: We're casting a `Sized` type to a `&[u8]` of its own length, so the sizes match.
+    unsafe { slice::from_raw_parts((t as *const T) as *const u8, mem::size_of::<T>()).to_vec() }
```
ppenna, line 769: Shouldn't it be `t.len()*mem::size_of::<T>()` ?

ppenna (changes requested):

src/microvm/src/vmm/microvm/kvm/vcpu/mod.rs
LucasGdosR: No. This helper function is for structs with fixed size, with no arrays. Despite having no arrays, `kvm_fpu` does not derive `serde` traits, so it must be serialized manually. The other helper function, `serialize_fam_struct`, deals with arrays.

PR #1003, opened 2025-09-22
[microvm] Remove Gateway Structure

csegarragonz:
In this commit I remove the gateway.rs structure used in the micro VM to monitor the connection to the system VM.

This commit is a first step towards implementing asynchronous polling for the micro VM. In particular, it will become necessary to have direct access to all the streams/queues managed by the I/O thread in order to multiplex them over a poll.

In addition, the gateway structure was not encapsulating too much complexity and was a bit misleading in terms of what it encapsulared (system VM vs Gateway).


src/microvm/src/io_thread.rs
```diff
@@ -149,30 +151,75 @@ impl IoThread {
...
+                if n + num_filled < buf.len() {
+                    self.system_vm_partial_read_buffer
+                        .extend(&buf[..(n + num_filled)]);
+
+                    // A partial read corresponds to a WouldBlock, so we return false.
+                    return Ok(false);
+                }
             },
+            // If WouldBlock, return false.
+            Err(e) if e.kind() == ErrorKind::WouldBlock => return Ok(false),
```
ppenna: If there were outstanding bytes not read yet from a previous call, and the call would block, we are throwing those bytes away.

src/microvm/src/io_thread.rs
```diff
@@ -149,30 +151,75 @@ impl IoThread {
...
     ///
     /// Upon success, the received message is pushed into the `incoming` queue, and `true` is returned.
     /// Otherwise, if it would block, `false` is returned. Otherwise, an error is returned.
     ///
-    fn try_receive_from_gateway(&mut self) -> Result<bool> {
-        match self.gateway.try_receive() {
-            Ok(message) => {
-                self.incoming.push_back(message);
-                Ok(true)
+    fn try_receive_from_system_vm(&mut self) -> Result<bool> {
```
ppenna, line 162: (nitpick) We are missing some type annotations in this function.

ppenna (changes requested):

src/microvm/src/io_thread.rs
csegarragonz: Good catch! I just copied-and-pasted this code from `gateway.rs`, but this was indeed a pre-existing bug.

ppenna (approved):

PR #1006, opened 2025-09-23
[microvm] Graceful Shutdown

csegarragonz:
In this PR I implement support to gracefully shutdown the micro VM using the control-plane API from nanvixd.

This change involves a few moving pieces. First, we need to pass the `control_plane_stream` to the IO thread, so that it can monitor it. Then, we need to actually trigger the interrupt support that we introduced in #818 (guess what, there was a bug that I fix in #1007). Lastly, and this is the biggest architectural change, we need the VMM thread to actually decide when has execution finished, in order to join on all the other components.

Following-up on the last point, right now the main VMM loop (the one that calls the orchestrator) is wrapped in an while loop bound to the lifecycle of the vCPU thread (i.e. `while !self.vcpu_thread.is_finished()`). It is cleaner if we make the orchestrator `run()` in its own loop, and decide when to break out of it in its own accord. 

Right now, there are two possible exit reasons: 
1. Nanvix sends a shutdown command: I/O thread sends shutdown to orchestrator, orchestrator sends interrupt to vCPU, and waits for vCPU to send a Shutdown message via control channel.
2. vCPU runs to completion: after finishing it sends a Shutdown message to the VMM via control channel like the last step above.

Lastly, the simple interrupt approach does not work to actually interrupt hyperlight. I implement a work-around in this PR (at the cost of a few error messages during shutdown), but we should consider how to implement this properly (see #1010).

### DIFF
--- a/src/microvm/src/io_thread.rs
+++ b/src/microvm/src/io_thread.rs
@@ -10,6 +10,7 @@ use crate::orchestrator::{
     IoControlResponse,
 };
 use ::anyhow::Result;
+use ::control_plane_api;
 use ::std::{
     collections::VecDeque,
     io::ErrorKind,
@@ -41,10 +42,12 @@ use ::syslog::{
 /// Private data of the I/O thread.
 ///
 pub struct IoThread {
-    /// Connection to the system VM.
-    system_vm_stream: SocketStream,
+    /// Optional connection to the system VM.
+    system_vm_stream: Option<SocketStream>,
     /// Buffer to handle partial reads from the system VM.
     system_vm_partial_read_buffer: VecDeque<u8>,
+    /// Optional connection to the external control-plane, nanvixd.
+    control_plane_stream: Option<SocketStream>,
     /// Gateway receiver.
     data_rx: Receiver<Message>,
     /// Gateway sender.
@@ -54,7 +57,7 @@ pub struct IoThread {
     /// Queue of outgoing messages.
     outgoing: VecDeque<Message>,
     /// Command sender to the VMM.
-    _control_tx: Sender<IoControlCommand>,
+    control_tx: Sender<IoControlCommand>,
     /// Response receiver from the VMM.
     control_rx: Receiver<IoControlResponse>,
     // TODO: channels to linuxd and nanvixd https://github.com/nanvix/nanvix/issues/945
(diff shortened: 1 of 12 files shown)


src/microvm/src/orchestrator.rs
```diff
@@ -181,7 +201,27 @@ impl Orchestrator {
...
+                break;
+            }
+            if self.try_receive_from_vcpu()? == Break(()) {
+                break;
+            }
+        }
+
+        // If we break out of the main loop, we need to shutdown the I/O and
+        // the memory thread.
+        debug!("run(): exitted run loop, cleaning-up");
```
LucasGdosR: Typo: exitted → exited

src/microvm/src/vmm/hyperlight/mod.rs
```diff
@@ -110,9 +111,14 @@ impl Vmm {
         _initrd_args: Option<String>,
         stderr: Option<String>,
         gateway_conn: Option<Gateway>,
+        control_plane_stream: Option<SocketStream>,
     ) -> Result<u16> {
         crate::timer!("vmm_creation");
 
+        // TODO (#1009): all IO-related channels should be Optional and only be initialized if
```
csegarragonz, line 124: Addressing this TODO in this PR creates an unnecessarily large diff.

src/microvm/src/io_thread.rs
```diff
@@ -164,75 +179,79 @@ impl IoThread {
     /// Otherwise, if it would block, `false` is returned. Otherwise, an error is returned.
     ///
     fn try_receive_from_system_vm(&mut self) -> Result<bool> {
-        let mut buf: [u8; mem::size_of::<Message>()] = [0; mem::size_of::<Message>()];
+        if let Some(system_vm_stream) = self.system_vm_stream.as_mut() {
```
csegarragonz, line 182: Only change in this method is adding a level of indentation with the `if let Some() = `. (And removing the `self.` from `self.system_vm_stream` to use the unpacked value.)

Seems like Github is doing a better job at formatting the diff.

src/microvm/src/io_thread.rs
```diff
@@ -277,30 +296,71 @@ impl IoThread {
...
-                    std::mem::offset_of!(syscall::LinuxDaemonMessage, payload)
-                        + std::mem::offset_of!(syscall::unistd::message::WriteRequest, buffer)
-                );
-                match self.system_vm_stream.write_all(&message_clone.to_bytes()) {
-                    Ok(()) => Ok(()),
-                    Err(e) if e.kind() == ErrorKind::WouldBlock => {
-                        self.outgoing.push_front(message);
-                        Ok(())
-                    },
+        if let Some(system_vm_stream) = self.system_vm_stream.as_mut() {
```
csegarragonz, line 299: Ditto as above. Only change is adding a level of indentation to unwrap the optional.

Alas, Github botching the diff.

ppenna (approved):

ppenna:
@danbugs we should chat about how to introduce a definitive solution for Hyperlight.

PR #1033, opened 2025-10-01
[libs] Make Control-Plane API Bi-directional

csegarragonz:
In this PR I enhance the control-plane API to support more complex messages, by moving to `bincode`-powered (de)-serialization, and supporting replies.

For the time-being, the control-plane API is bi-directional from nanvixd to the system/user VM and vice-versa. When the message is sent from nanvixd, we don't differentiate between recipients.

In follow-up PRs I will add messages from the system VM to nanvixd, to help with synchronization during system provisioning. 

We can also explore merging the user-vm API with this control-plane API. In this case, however, we should also consider having a control-plane stream between the system and the user VM. If I am not mistaken this feature is also desirable for snapshots.

### DIFF
--- a/src/libs/config/src/lib.rs
+++ b/src/libs/config/src/lib.rs
@@ -57,10 +57,18 @@ pub mod syscomm {
     ///
     /// # Description
     ///
+    /// Provides the maximum length of a message we receive. We set a limit to
+    /// prevent a malformed payload to trigger unbounded allocations in linuxd.
+    ///
+    pub const MAX_MESSAGE_LEN: usize = 128 * 1024 * 1024;
+
+    ///
+    /// # Description
+    ///
     /// Provides the maximum length of a message we receive in linuxd. We set a limit to
     /// prevent a malformed payload to trigger unbounded allocations in linuxd.
     ///
-    pub const MAX_LINUXD_MESSAGE_LEN: usize = 128 * 1024 * 1024;
+    pub const MAX_LINUXD_MESSAGE_LEN: usize = MAX_MESSAGE_LEN;
 
     ///
     /// # Description
--- a/src/libs/config/src/lib.rs
+++ b/src/libs/config/src/lib.rs
@@ -65,14 +65,6 @@ pub mod syscomm {
     ///
     /// # Description
     ///
-    /// Provides the maximum length of a message we receive in linuxd. We set a limit to
-    /// prevent a malformed payload to trigger unbounded allocations in linuxd.
-    ///
-    pub const MAX_LINUXD_MESSAGE_LEN: usize = MAX_MESSAGE_LEN;
-
-    ///
-    /// # Description
-    ///
     /// Provides the maximum number of poll events we can buffer. We set the limit to 128 as, for
     /// the time being, it is unlikely we will have more than 128 connections being activated at
     /// the same time.
--- a/src/libs/control-plane-api/src/lib.rs
+++ b/src/libs/control-plane-api/src/lib.rs
@@ -1,13 +1,31 @@
 // Copyright(c) The Maintainers of Nanvix.
 // Licensed under the MIT License.
 
-//! This file contains the messages used in the control-plane API between
-//! nanvixd and linuxd/user VM.
+//!
+//! This file contains the messages used in the control-plane API between the control-plane
+//! (nanvixd), the system VM (linuxd), and the user VM. It defines a simple wire-format to allow
+//! for bi-directional communication. The specification for the wire-format is as follows:
+//! - source: u8 -> Source of the control-plane message (Nanvixd, SystemVm, or UserVm).
+//! - length: u32 LE -> Length of the message payload.
+//! - bytes: [u8; length]
+//!
+
+#![forbid(clippy::cast_possible_truncation)]
+#![forbid(clippy::cast_possible_wrap)]
+#![forbid(clippy::cast_precision_loss)]
+#![forbid(clippy::unnecessary_cast)]
+#![forbid(invalid_reference_casting)]
+#![forbid(clippy::fn_to_numeric_cast)]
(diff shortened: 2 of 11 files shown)


src/libs/control-plane-api/src/lib.rs
```diff
@@ -17,32 +27,150 @@ use ::syscomm::{
...
+        let reason: String = format!("unexpected control-plane command source (got={}, expected={})", source as u8, T::SOURCE as u8);
+        error!("{reason}");
+        return Err(Error::new(ErrorKind::InvalidData, reason).into());
+    }
+
+    // Read command length.
+    let mut length_bytes: [u8; 4] = [0u8; 4];
+    let num_read: usize = stream.try_read_exact(&mut length_bytes)?;
+    debug_assert_eq!(num_read, 4);
+    let len: usize = u32::from_le_bytes(length_bytes) as usize;
```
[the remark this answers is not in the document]
csegarragonz: I think it is safe to assume that `size_of::u32` <= `size_of::usize`?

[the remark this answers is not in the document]
ppenna: Yes, it is. Please add the following (if possible) to `lib.rs`:

```
#![forbid(clippy::cast_possible_truncation)]
#![forbid(clippy::cast_possible_wrap)]
#![forbid(clippy::cast_precision_loss)]
#![forbid(clippy::unnecessary_cast)]
#![forbid(invalid_reference_casting)]
#![forbid(clippy::fn_to_numeric_cast)]
#![forbid(clippy::fn_to_numeric_cast_with_truncation)]
```

src/libs/config/src/lib.rs
```diff
@@ -54,13 +54,21 @@ pub mod system {
...
 pub mod syscomm {
 
+    ///
+    /// # Description
+    ///
+    /// Provides the maximum length of a message we receive. We set a limit to
+    /// prevent a malformed payload to trigger unbounded allocations in linuxd.
+    ///
+    pub const MAX_MESSAGE_LEN: usize = 128 * 1024 * 1024;
+
```
ppenna, line 64: Why is this new constant required?

src/libs/control-plane-api/src/lib.rs
```diff
@@ -17,32 +27,156 @@ use ::syscomm::{
...
+///
+/// - `stream`: the control-plane stream where to send the message.
+/// - `msg`: a message that implements the WireCommand trait.
+///
+/// # Returns
+///
+/// In case of success, returns nothing. Otherwise, returns an error.
+///
+pub fn send_command<T: WireCommand>(stream: &mut SocketStream, msg: &T) -> Result<(), SocketError> {
+    let payload: Vec<u8> = bincode::serde::encode_to_vec(msg, bincode_cfg())
```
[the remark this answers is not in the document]
csegarragonz: Messages may get more complicated down the road, so I'd stick to passing by reference.

src/libs/config/src/lib.rs
csegarragonz: It is used as a sanity-check before we allocate a buffer for the newly received message: https://github.com/nanvix/nanvix/blob/enhancement-control-plane-api-bidirectional/src/libs/control-plane-api/src/lib.rs#L159-L164

(We read the message length from the payload itself as part of the wire-format.)

src/libs/control-plane-api/src/lib.rs
```diff
@@ -17,32 +27,156 @@ use ::syscomm::{
...
+            error!("{reason}");
+            Error::new(ErrorKind::InvalidData, reason)
+        })?;
+
+
+    if payload.len() > u32::MAX as usize {
+        let reason: String = format!("payload length exceeds maximum allowed size (length={}, max={})", payload.len(), u32::MAX);
+        error!("{reason}");
+        return Err(Error::new(ErrorKind::InvalidData, reason).into());
+    }
```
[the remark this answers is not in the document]
csegarragonz: I think we do need this check, and you suggested it yourself before :D

src/libs/config/src/lib.rs
ppenna: Sure, but why could't we use directly `MAX_LINUXD_MESSAGE_LEN`, or rename `MAX_LINUXD_MESSAGE_LEN` to `MAX_MESSAGE_LEN` ?

src/libs/control-plane-api/src/lib.rs
```diff
@@ -17,32 +35,162 @@ use ::syscomm::{
...
+    const SOURCE: Source;
+}
+
+///
+/// # Description
+///
+/// Default bincode options to encode/decode control-plane messages. This configuration uses
+/// fixed-size ints, and little-endian encoding.
+///
+fn bincode_cfg() -> impl bincode::config::Config {
```
ppenna: This can be made a `const` function I believe.

ppenna (dismissed):
I understand that control messages may get more complex down the road, still using 'serdes' looks a bit overkill now

csegarragonz:
> I understand that control messages may get more complex down the road, still using 'serdes' looks a bit overkill now

I have a PR where I am adding the user vm ID in a message (a `u32`) and down the road I want to merge the control-plane API with the user-vm API (which also uses `serdes`). 

I am open to re-assessing in the future and moving to something more lightweight.

src/libs/config/src/lib.rs
csegarragonz: Its not exactly a LINUXD message, so that is why I added a new constant. I did not want to refactor MAX_LINUXD in this PR to not add more noise but, if you are happy about it, I can refactor it in this PR.

csegarragonz: Done!

ppenna (approved):

PR #1044, opened 2025-10-03
[libs] Asynchronous Wrapper of Syscomm Library

csegarragonz:
In this PR I add an optional extension to the syscomm library, gated under the `async` feature, that implements `async` versions of `read_exact` and `accept`.

We currently have blocking non-async versions of this methods, but having async versions makes it possible to integrate them better with the parts of the codebase where we use `tokio` as asynchronous runtime.

In particular, I use the new async `accept` method to replace the blocking `accept_timeout` used when spawning linuxd and the user vm fron nanvixd.

### DIFF
--- a/src/libs/syscomm/src/lib.rs
+++ b/src/libs/syscomm/src/lib.rs
@@ -2,6 +2,13 @@
 // Licensed under the MIT License.
 
 //==================================================================================================
+// Modules
+//==================================================================================================
+
+#[cfg(feature = "async")]
+pub mod r#async;
+
+//==================================================================================================
 // Imports
 //==================================================================================================
 
@@ -34,6 +41,10 @@ use ::std::{
         ErrorKind,
         Read,
         Write,
+    },
+    os::fd::{
+        AsRawFd,
+        RawFd,
     },
     thread,
     time::{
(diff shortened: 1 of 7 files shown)


src/libs/syscomm/src/lib.rs
```diff
@@ -1,6 +1,17 @@
 // Copyright(c) The Maintainers of Nanvix.
 // Licensed under the MIT License.
 
+//==================================================================================================
+// Modules
+//==================================================================================================
+
+#[cfg(feature = "async")]
+pub mod r#async;
```
csegarragonz, line 9: I have named the module `async` because that is what it felt more natural, but it has the disadvantage that you need to escape it, as it is a rust keyword.

I was also not feeling very inspired with any new names, but open to suggestions :D

src/utils/nanvixd/src/sandbox/microvm.rs
```diff
@@ -119,19 +120,38 @@ impl Microvm {
...
+            Ok(Ok(stream)) => stream,
+            Ok(Err(e)) => {
                 // If the user VM has not accepted the control-plane connection, it means that
                 // something went wrong during start-up. We kill the process ignoring errors,
                 // and return an error.
                 let reason: String =
                     format!("error connecting control-plane to user VM (error={e:?})");
                 error!("{reason}");
 
+                // FIXME: use send_sigkill_to_child after it lands
```
csegarragonz: I add a `Self::send_sigkill_to_child` in #1033, so, if #1033 lands first, I will remove this `FIXME` and use the method introduced there.

ppenna (approved):

PR #1046, opened 2025-10-03
[libs] Fix BlockingSocketStream Read Implementation

### LINKED ISSUE #1045: [microvm] User VM API Race Condition
## Description

This is where it blocks:
https://github.com/nanvix/nanvix/blob/dev/src/libs/user-vm-api/src/lib.rs#L165

I think it could be a race condition in that  the code here :
https://github.com/nanvix/nanvix/blob/dev/src/libs/syscomm/src/lib.rs#L635

Seems to have been written with the assumption that each write to the socket will result in a distinct readable event, but I am ot 100% sure this is correct, I think it may be possible that the 2 writes at:

https://github.com/nanvix/nanvix/blob/dev/src/libs/user-vm-api/src/lib.rs#L130

to only result in one event hence the code will then wait forever

## Explanation

I think this is indeed a bug. Intuitively, whenever we are awoken in a poll, we need to drain it until it returns WouldBlock. However, in the BlockingSocketStream implementation we don't do such draining. This means that the second message would be lost.

Reported by @simongdavies

csegarragonz:
In this commit I fix a stall bug in the read implementation for the BlockingSocketStream class.

In particular, by calling poll before trying to read, we were creating a situation where we may successfully return a read, but not fully drain the socket stream. This means that the next time we call poll, we may not need to call poll at all because the data would already be there.

This bug could manifest in the reader hanging indefinitely.

Closes #1045

### DIFF
--- a/src/libs/syscomm/src/lib.rs
+++ b/src/libs/syscomm/src/lib.rs
@@ -615,7 +615,9 @@ impl BlockingSocketStream {
     ///
     /// # Description
     ///
-    /// Blocking read implementation.
+    /// Blocking read implementation. We need to be careful to always drain the stream before
+    /// calling poll, as otherwise we may call poll again when we still have data available to be
+    /// read.
     ///
     /// # Parameters
     ///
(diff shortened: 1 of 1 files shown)


csegarragonz:
@simongdavies tagging you FYI. I think this should fix the issue you reported.

simongdavies:
> @simongdavies tagging you FYI. I think this should fix the issue you reported.

@csegarragonz  I tested this locally and it does fix the issue Thanks!!

ppenna:
Superseded by https://github.com/nanvix/nanvix/pull/1052

PR #1079, opened 2025-10-20
[ci] Re-enable L2 Benchmarks

### LINKED ISSUE #926: [ci] Cold-Start L2 Benchmark Fails For Hyperlight
## Description

I have not been able to reproduce it locally.

### LINKED ISSUE #937: [utils] Cold-Start L2 Benchmark Fails For MicroVM

### LINKED ISSUE #975: [utils] Warm-Start L2 Benchmark Fails for MicroVM
## Description

Warm-start benchmarks with the L2 VM are failing intermittently.

We should investigate and fix the issue.

csegarragonz:
In this PR I re-enable the L2 benchmarks for the Micro VM and Hyperlight. After the major re-factor, the race conditions that we were seeing should not no longer appear. I fix a few issues that were introduced precisely because we were not testing the L2 deployment at all.

Closes #975 
Closes #937
Closes #926

### DIFF
--- a/src/utils/nanvixd/src/args.rs
+++ b/src/utils/nanvixd/src/args.rs
@@ -13,6 +13,7 @@ use ::std::{
     fs::File,
     io::BufReader,
 };
+use ::syscomm::SocketType;
 
 //==================================================================================================
 // Structures
@@ -138,6 +139,26 @@ impl Args {
             }
 
             i += 1;
+        }
+
+        // If we deploy linuxd in an L2 VM, we need to make sure that all socket types are set to
+        // TCP.
+        if l2 {
+            if control_plane_socket_type == Some(SocketType::UNIX_STR.to_string()) {
+                anyhow::bail!("control-plane must use a tcp socket in l2 deployments");
+            }
+
+            if gateway_socket_type == Some(SocketType::UNIX_STR.to_string()) {
+                anyhow::bail!("gateway must use a tcp socket in l2 deployments");
+            }
+
+            if system_vm_socket_type == Some(SocketType::UNIX_STR.to_string()) {
+                anyhow::bail!("system vm must use a tcp socket in l2 deployments");
+            }
+
+            control_plane_socket_type = Some(SocketType::TCP_STR.to_string());
+            gateway_socket_type = Some(SocketType::TCP_STR.to_string());
+            system_vm_socket_type = Some(SocketType::TCP_STR.to_string());
         }
 
         Ok(Self {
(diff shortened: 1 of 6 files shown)


src/libs/syscomm/src/socket_unbound.rs
ppenna, line 1: Previous message was correct. We should tag the name of the function where the error message is located.

src/utils/nanvixd/src/args.rs
```diff
@@ -140,6 +142,26 @@ impl Args {
...
+                warn!("{} and {}={} are incompatible options, overwriting {} to {}", Self::OPT_L2, Self::OPT_CONTROL_PLANE_SOCKET_TYPE, SocketType::UNIX_STR, Self::OPT_CONTROL_PLANE_SOCKET_TYPE, SocketType::TCP_STR);
+            }
+
+            if gateway_socket_type == Some(SocketType::UNIX_STR.to_string()) {
+                warn!("{} and {}={} are incompatible options, overwriting {} to {}", Self::OPT_L2, Self::OPT_GATEWAY_SOCKET_TYPE, SocketType::UNIX_STR, Self::OPT_GATEWAY_SOCKET_TYPE, SocketType::TCP_STR);
+            }
+
+            if system_vm_socket_type == Some(SocketType::UNIX_STR.to_string()) {
+                warn!("{} and {}={} are incompatible options, overwriting {} to {}", Self::OPT_L2, Self::OPT_SYSTEM_VM_SOCKET_TYPE, SocketType::UNIX_STR, Self::OPT_SYSTEM_VM_SOCKET_TYPE, SocketType::TCP_STR);
+            }
```
ppenna, line 157: Instead of testing for `UNIX` we should test for `TCP` and bail if not expected, logging will not work as this is not enabled at this point.

csegarragonz: This is not entirely correct. (The logging is).

To enable L2 deployments, we should only need to provide the `-l2` flag to nanvixd. However, for flexibility, we also support setting the individual `-control-plane-socket-type`. What I am trying to capture here is that setting both the `-l2` flag and _any_ `*-socket-type unix` is an invalid choice of arguments.

In addition, given that the default socket type is UNIX we need to explicitly set the socket types to TCP here.

ppenna: Yes, I agree. What I am suggesting is that we should not override any user informed parameters. Instead, we should check for them and fail if the combination is not valid.

csegarragonz: Agree. In this case the only invalid combination is `-l2` and `*-socket-type unix`. In which case we should `bail!`.

However, if `-l2` is set, we need to manually set the socket types to `tcp` because the default, when the argument is missing, is `unix`.

PR #1088, opened 2025-10-21
[scripts] Add Wrapper Script To Launch ./bin/nanvix-bench.elf

csegarragonz:
In this PR I extend the `./scripts/ci_bench_summary.py` script to also wrap around the invocation of `./bin/nanvix-bench.elf`. 

To this extent, I rename the script to `./scripts/benchmark.py` and add three top-level commands: `run`, `ci-summary`, and `copy-results`. The `run` command runs benchmarks and parses the output to CSV files. The `ci-summary` consumes these CSV files to generate markdown tables. In the future we could have different tasks that process the CSV files to different ends. Lastly, the `copy-results` copies the results from one directory to the other.

This keeps all the logic regarding file naming in the same script, reducing the possibility of errors. In this same line, I slightly re-factor the naming convention, which means that we have no `dev` results for any benchmark, as the new results for the new naming convention are still missing.

With this re-factor I also fix the issue of the `round-trip-latency` benchmark results not being persisted, which I highlight in a comment below.


.github/actions/benchmark/action.yml
```diff
@@ -146,15 +177,22 @@ runs:
   - name: Run only on dev branch
     if: github.ref_name == 'dev'
     run: |
-      mkdir -p $HOME/ubench-results-dev
+      target_dir=$HOME/ubench-results-dev
+      mkdir -p ${target_dir}
 
-      for machine in microvm hyperlight; do
-        for benchmark in boot_time cold_start warm_start warm_start_vmm; do
```
csegarragonz, line 152: We had a hard-coded list of what benchmarks to copy results when running the benchmarks in `dev`, and I'd forgotten to add `round-trip-latency` to the list. Now we are just re-using the input.

ppenna (approved):

PR #1160, opened 2025-11-05
[daemons] Spawn One Reader Task Per User VM

csegarragonz:
In this PR I re-factor linuxd to spawn one lightweight task per user VM to read from its socket, and enqueue messages to a shared, fan-in, queue. The main linuxd thread only needs to then `select!` on the queue to receive messages from _all_ user VMs.

This addresses a limitation in the current design where we were only able to poll one user VM at a time, picked at random. This was causing starvation, and deadlocking, in the `concurrent` benchmark that I also introduce in this PR.

### DIFF
--- a/src/daemons/linuxd/src/lib.rs
+++ b/src/daemons/linuxd/src/lib.rs
@@ -15,9 +15,11 @@ use crate::{
     config::{
         restore_gate_sockaddr_builder,
         CONTROL_PLANE_CONNECT_TIMEOUT,
+        READER_TASK_JOIN_TIMEOUT,
     },
     message::RequestAssembler,
     syscalls::SyscallTable,
+    user_vm_event::UserVmEvent,
     user_vm_handle::UserVmHandle,
     venv::{
         VenvCommand,
@@ -34,7 +36,6 @@ use ::control_plane_api::{
     self,
     NanvixdControlMessage,
 };
-use ::rand::seq::SliceRandom;
 use ::std::{
     collections::{
         HashMap,
@@ -79,12 +80,14 @@ use ::tokio::{
     net::TcpListener,
     sync::{
         mpsc::{
+            channel,
             Receiver,
             Sender,
         },
         Mutex,
         MutexGuard,
     },
+    task::JoinHandle,
     time::timeout,
 };
 use ::user_vm_api::{
@@ -102,6 +105,7 @@ mod error;
 mod linux;
 mod message;
 mod time;
+mod user_vm_event;
 mod user_vm_handle;
 mod venv;
 mod worker_thread;
@@ -124,6 +128,14 @@ pub mod syscalls;
 /// Maximum number of messages that can be queued in a channel to a worker thread.
 ///
 pub const WORKER_THREAD_CHANNEL_CAPACITY: usize = 1024;
+
+///
+/// # Description
+///
+/// Maximum number of messages that can be queued in a channel multiplexing messages from all user
+/// VMs being served by this linuxd instance.
+///
+const USER_VM_CHANNEL_CAPACITY: usize = 1024;
 
 //==================================================================================================
 // Structures
@@ -258,6 +270,7 @@ impl<T: Sync + Send + 'static> LinuxDaemon<T> {
     async fn accept_connections(
         &self,
         mut user_vm_stream: SocketStream,
+        user_vm_event_tx: Sender<UserVmEvent>,
     ) -> Result<(UserVmIdentifier, UserVmHandle), Error> {
         trace!("accepted connection from user VM (addr={user_vm_stream:?})",);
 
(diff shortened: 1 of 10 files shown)


src/daemons/linuxd/src/lib.rs
```diff
@@ -536,50 +579,81 @@ impl<T: Sync + Send + 'static> LinuxDaemon<T> {
...
-        for uvm_id in ids {
-            let Some(uvm_handle) = user_vm_connections.get(&uvm_id).cloned() else {
-                continue;
-            };
-            let message: Message = match Self::recv(uvm_handle.get_user_vm_reader()).await {
-                Ok(m) => m,
-                Err(error_kind) => {
-                    // Log error message only if error was not caused by a disconnection.
-                    if error_kind != ErrorKind::ConnectionAborted
-                        && error_kind != ErrorKind::WouldBlock
```
csegarragonz, line 557: Unrelated, but note how here we were propagating `WouldBlock` into a fatal error, even though our code in `Self::recv` is resilient to `WouldBlock`s. I fix that too.

src/daemons/linuxd/src/user_vm_event.rs
```diff
@@ -0,0 +1,23 @@
+use ::std::io::ErrorKind;
```
ppenna, line 8: Please add missing copyright notices, as well as section headers.

src/utils/nanvix-bench/src/main.rs
```diff
@@ -535,6 +536,98 @@ impl Benchmark {
...
+    ///
+    /// This benchmark measures the time to start N concurrent user VMs, all sharing the same
+    /// linuxd instance.
+    ///
+    /// # Arguments
+    ///
+    /// - `l2`: whether to deploy linuxd in an L2 VM or not.
+    /// - `num_concurrent_vms`: number of VMs to start concurrently.
+    ///
+    pub async fn run_concurrent(&mut self, l2: bool, num_concurrent_vms: usize) -> Result<()> {
```
ppenna, line 586: Please explicit type annotations for variable definitions.

src/utils/nanvix-bench/src/main.rs
```diff
@@ -535,6 +536,98 @@ impl Benchmark {
...
+
+            pb.inc(1);
+
+            // Need to give some time to clean-up (a bit longer for L2 benchmarks).
+            if l2 {
+                sleep(Duration::from_millis(CLEANUP_L2_SLEEP_DURATION)).await;
+            } else {
+                sleep(Duration::from_millis(CLEANUP_SLEEP_DURATION)).await;
+            }
+        }
```
ppenna, line 620: This can be generalized to accommodate both types of benchmarks: cold starts and spawn(), if we introduce an additional boolean argument that dictates, whether or not we should persist connections, and then kill them.

src/daemons/linuxd/src/lib.rs
```diff
@@ -536,50 +579,81 @@ impl<T: Sync + Send + 'static> LinuxDaemon<T> {
...
+    ///
+    async fn user_vm_reader_loop(
+        uvm_id: UserVmIdentifier,
+        uvm_handle: UserVmHandle,
+        uvm_events_tx: Sender<UserVmEvent>,
+    ) {
+        trace!("user_vm_reader_loop(): starting (uvm_id={uvm_id})");
+
+        loop {
+            match Self::recv(uvm_handle.get_user_vm_reader()).await {
```
ppenna: Can we just simplify recv to use `read_exact()`, as we don't need to care about cancel safety anymore?

src/daemons/linuxd/src/lib.rs
```diff
@@ -536,50 +579,81 @@ impl<T: Sync + Send + 'static> LinuxDaemon<T> {
...
+                        .send(UserVmEvent::ConnectionClosed { uvm_id })
+                        .await;
+                    break;
+                },
+                Err(ErrorKind::WouldBlock) => {
+                    // Self::recv() properly handles partial reads by persisting them in a buffer,
+                    // so if it surfaces a WouldBlock we can safely just continue.
+                    // Optionally: tokio::task::yield_now().await;
+                    continue;
+                },
```
ppenna: I would not catch this and instead, let it trigger an error, as this is truly not expected and if it happens, it is an error.

src/daemons/linuxd/src/lib.rs
```diff
@@ -536,50 +579,81 @@ impl<T: Sync + Send + 'static> LinuxDaemon<T> {
...
-        // Shuffle the order in which we poll user VMs to avoid starvation.
-        ids.as_mut_slice().shuffle(&mut ::rand::rng());
+    ///
+    /// # Description
+    ///
+    /// This function implements the background asynchronous reading loop for a single user VM.
+    /// Each user VM spawns a task like this, and all messages are fanned in to a single MPSC
+    /// channel monitored by the main select!.
+    ///
+    async fn user_vm_reader_loop(
```
ppenna, line 609: This function should have a return value, so we can distinguish from cases where we ended due to an error. On error we should at least terminate abort the connection to the userVM and harvest resources of that user vm.

src/daemons/linuxd/src/lib.rs
```diff
@@ -442,31 +462,54 @@ impl<T: Sync + Send + 'static> LinuxDaemon<T> {
...
+                                    uvm_id,
+                                    uvm_handle,
+                                    message,
+                                    worker_threads.clone(),
+                                )
+                                .await
+                            {
+                                let reason: &'static str = "error processing message from user VM";
+                                error!("run(): {reason} (uvm_id={uvm_id}, error={e:?})");
+                                return Err(Self::log_and_error(ErrorCode::IoErr, reason));
```
ppenna: We may want to consider only terminating the buggy VM.

src/daemons/linuxd/src/lib.rs
csegarragonz: If we move to `read_exact` this is indeed an error.

ppenna (dismissed):

csegarragonz (commented):
New commit LGTM

PR #1187, opened 2025-12-10
Update hyperlight to v0.12.0

danbugs:
Updates hyperlight dependencies from v0.11.0 to v0.12.0 via nanvix/v0.12.0 branch. Removes unwrap_mgr() call since mgr field is now directly accessible in the updated API.


ppenna (approved):
LGTM

PR #1199, opened 2025-12-15
[nanvix-sandbox] Separate Control-Plane Bind/Connect Addresses

csegarragonz:
In this PR I modify the sandbox crate to differentiate between the socket address of the control-plane listener socket and the connecting socket. 

In an L2 deployment, each sandbox may be deployed in a (potentially different) namespace, whereas the control-plane listener socket is part of nanvixd, which executes in the root namespace. This means that to connect to the control-plane, each sandbox must connect through its VETH pair, which varies from namespace to namespace. On the other hand, the listener socket must bind to all interfaces in the root namespace.

This addresses a bug where we could not run more than one L2 VMs concurrently. Our current benchmarks did not catch this because:
1. In `concurrent-l2` we have concurrent user VMs, not concurrent linuxd instances (I may change this in a follow-up PR).
2. In `cold-start-l2` we were resetting the `nanvixd` instance at each _iteration_, which was generally bad practice.

In this PR I also address 2, by slightly refactoring `nanvix-bench`, so that I could reliably trigger this bug, and fix it.


src/libs/nanvix-sandbox/src/netns.rs
```diff
@@ -958,7 +961,8 @@ fn init_namespace(info: &NetnsInfo) -> Result<()> {
             .arg(info.ns_name())
             .arg("sysctl")
             .arg("-w")
-            .arg("net.ipv4.ip_forward=1"),
+            .arg("net.ipv4.ip_forward=1")
+            .stdout(Stdio::null()),
```
ppenna: Why is this required? If intention is to suppress output, it's better to write it to the log file to help debugging failures.

src/libs/nanvix-sandbox/src/config.rs
```diff
@@ -193,11 +193,15 @@ pub(crate) fn get_clh_api_socket_path(tmp_dir: &str) -> String {
 ///
 /// # Description
 ///
-/// Builds the control plane socket address for a given tenant ID. If nanvixd is configured to
-/// spawn linuxd in an L2 VM, it will return a TCP socket address, otherwise a Unix socket one.
+/// Builds the control plane socket addresses for nanvixd (bind) and for the linuxd and all
+/// user VM instances associated with a tenant ID. If nanvixd is configured to spawn linuxd inside
+/// an L2 VM, it will return TCP socket addresses, otherwise UNIX socket ones.
```
ppenna: > If nanvixd is configured to spawn linuxd inside an L2 VM, it will return TCP socket addresses, otherwise UNIX socket ones.

This is not accurate and does not reflect our current implementation. Note the logic is not gated whether we are in a L2 deployment or not.

src/libs/nanvix-sandbox/src/sandbox_config.rs
```diff
@@ -60,10 +60,14 @@ pub struct SandboxConfig<T> {
     #[cfg(feature = "single-process")]
     syscall_table: Option<::std::sync::Arc<::linuxd::syscalls::SyscallTable<T>>>,
 
-    /// Optional information on control plane socket (address, socket type).
-    /// This must be provided if the control plane socket is not already initialized before
-    /// sandbox initialization. If both socket and info are provided, the existing socket is used.
-    control_plane_socket_info: Option<(String, SocketType)>,
+    /// Optional information on the control plane listener socket (address, socket type).
+    /// This must be provided if the control plane listener socket is not already initialized before
+    /// sandbox initialization. There is one listener socket per deployment. If both socket and info are provided, the existing socket is used.
```
ppenna: This comment line should wrap.

csegarragonz: Interesting, we may want to enable this guy?
https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=#wrap_comments

src/libs/nanvix-sandbox/src/netns.rs
csegarragonz: I cannot remember why I needed this, I have reverted the change.

src/libs/nanvix-sandbox/src/config.rs
csegarragonz: Updated the comment.

src/libs/nanvix-sandbox-cache/src/lib.rs
```diff
@@ -353,8 +353,8 @@ impl<T: Sync + Send + Default + 'static> SandboxCache<T> {
                 #[cfg(feature = "single-process")]
                 let gateway_l2_port: Option<TcpPort> = None;
 
-                // Add Linux Daemon instance to sandbox if one exists for the tenant.
```
ppenna, line 356: No need to move this line.

src/libs/nanvix-sandbox-cache/src/lib.rs
```diff
@@ -509,9 +516,9 @@ impl<T: Sync + Send + Default + 'static> SandboxCache<T> {
                         },
                     };
 
-                // Update control-plane socket.
-                self.control_plane_socket
-                    .replace(initialized_sandbox.control_plane_socket_info());
+                // Update control-plane listener socket (re-used across all sandboxes).
+                self.control_plane_bind_socket
+                    .replace(initialized_sandbox.control_plane_bind_socket_info());
```
ppenna, line 521: Now that we fixed the property that all sandboxes bind to the same control plane address, we probably want to create the control plane socket when instantiating the cache. This lazy creation here is dangerous. We can address this improvement in a good first issue and assign Copilot to work on it. What do you think?

csegarragonz: Yes, I agree. I have created a tracking issue: https://github.com/nanvix/nanvix/issues/1242

PR #1213, opened 2025-12-21
[uservm] E: Remove kvm_fpu from FpuState

### LINKED ISSUE #1208: [uservm] Remove FPU from Snapshot State
## Context

Currently, when saving a snapshot in the UserVm, the VirtualProcessorState saves XSAVE and FPU.

Upon further inspection, it seems like XSAVE already includes FPU, indicating this is redundant state (Firecracker does not save kvm_fpu, for example). We should not include this in the snapshot.

## Proposed Solution

Remove `fpu` from `FpuState` and the corresponding saving logic in `save_state`.

## Alternative Solutions

We could just keep it as is, as `kvm_fpu` is a small struct with only 418 bytes. Although storing it twice is redundant, it is not incorrect.

LucasGdosR:
Removes `kvm_fpu` from `FpuState`, as `Xsave` already encompasses the floating point registers. Closes #1208

### DIFF
--- a/src/uservm/src/vmm/microvm/kvm/vcpu/fpu.rs
+++ b/src/uservm/src/vmm/microvm/kvm/vcpu/fpu.rs
@@ -59,8 +59,6 @@ pub struct Fpu {
 /// FPU state.
 #[derive(Serialize, Deserialize)]
 pub struct FpuState {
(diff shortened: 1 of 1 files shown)


ppenna (approved):
Looks good to me.

PR #1226, opened 2025-12-24
[uservm] Snapshot reading correction

### LINKED ISSUE #1223: [uservm] Vmm::load_snapshot() expects wrong file structure for vCPU
## Description

`Vmm::load_snapshot()` loads the virtual memory file correctly. However, it fails when loading the KVM file. The file is written in this order: `guest_state`, `vcpu_state`, `irqchip_state`, `timer_state`. When reading it, only `vcpu_state` is expected.

## Expected Behavior

After opening the file, all components should be read in the same order as they were written. Then, they should all be loaded back into KVM, like [so](https://github.com/firecracker-microvm/firecracker/blob/f0691f8253d4bde225b9f70ecabf39b7ad796935/src/vmm/src/builder.rs#L433).

LucasGdosR:
Corrects the expected snapshot file structure for loading.

Having 4 different structs inside a file was not helpful. I created a single wrapper structure to make serialization and deserialization straighforward with a single call.

Having a single structure also provides a single source of truth for the ordering inside the file, instead of having to keep the writing and reading orders consistent manually.

Closes #1223.

### DIFF
--- a/src/uservm/src/vmm/microvm/kvm/mod.rs
+++ b/src/uservm/src/vmm/microvm/kvm/mod.rs
@@ -7,8 +7,154 @@
 //! This module provides the backend implementation of MicroVM for Linux KVM.
 //!
 
+//==================================================================================================
+// Exports
+//==================================================================================================
+
 pub mod irqchip;
 pub mod pmio;
 pub mod timer;
 pub mod vcpu;
 pub mod vmem;
+
+//==================================================================================================
+// Imports
+//==================================================================================================
+
+use crate::vmm::{
+    guest::GuestState,
+    kvm::{
+        irqchip::IrqChipState,
+        timer::TimerState,
+        vcpu::VirtualProcessorState,
+    },
+};
+use ::anyhow::Result;
+use ::serde::{
+    Deserialize,
+    Serialize,
+};
+use ::syslog::trace;
+
+//==================================================================================================
+// Structures
+//==================================================================================================
+
+///
+/// # Description
+///
+/// A structure that holds all KVM components for snapshot serialization and deserialization.
+///
+#[derive(Serialize, Deserialize)]
+pub struct KvmSnapshot {
+    /// Guest component.
+    guest_state: GuestState,
+    /// Virtual processor component.
+    vcpu_state: VirtualProcessorState,
+    /// IrqChip component.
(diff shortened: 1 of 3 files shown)


LucasGdosR:
@ppenna I have addressed all of copilot's comments. I have left the PR as a draft, but it is ready for review. Has the workflow changed for having PRs evaluated? Should I keep them as drafts until approved or should I mark them as ready for review early?

ppenna:
> @ppenna I have addressed all of copilot's comments. I have left the PR as a draft, but it is ready for review. Has the workflow changed for having PRs evaluated? Should I keep them as drafts until approved or should I mark them as ready for review early?

@LucasGdosR , the review workflow did not change. Also, please expect a delay from my end to review this because of holidays.

ppenna (approved):
LGTM. This looks like an improvement rather than a bug. What is the buggy behavior of existing code?

LucasGdosR:
[Previously](https://github.com/nanvix/nanvix/pull/1226/files#diff-2bb285216b4339bd543bcb76513e2c616471fe9be88cf670557fd8fc9b7e9624L644) we tried to deserialize a `VirtualProcessorState` from the file, but that was not what was written. This call would error every time.

[Now](https://github.com/nanvix/nanvix/pull/1226/files#diff-2bb285216b4339bd543bcb76513e2c616471fe9be88cf670557fd8fc9b7e9624R606) we deserialize in line with the file's contents.

```
-             match ::serde_cbor::from_reader::<VirtualProcessorState, &mut File>(&mut file) {
+             match ::serde_cbor::from_reader::<KvmSnapshot, &mut File>(&mut file) {
```

ppenna:
> [Previously](https://github.com/nanvix/nanvix/pull/1226/files#diff-2bb285216b4339bd543bcb76513e2c616471fe9be88cf670557fd8fc9b7e9624L644) we tried to deserialize a `VirtualProcessorState` from the file, but that was not what was written. This call would error every time.
> 
> [Now](https://github.com/nanvix/nanvix/pull/1226/files#diff-2bb285216b4339bd543bcb76513e2c616471fe9be88cf670557fd8fc9b7e9624R606) we deserialize in line with the file's contents.

Thanks for the clarification. Please wait for PR #1231 to be merged, rebase this PR and open it for final review.

PR #1233, opened 2026-01-06
[build] Set SINGLE_PROCESS=yes as Default

ppenna:
Refactor nanvix-sandbox for multi-process deployment mode:
- Updated Cargo.toml to replace single-process feature with multi-process.
- Modified config.rs, initialized.rs, and other source files to conditionally compile based on multi-process feature.
- Adjusted documentation to reflect the new deployment modes and their purposes.
- Ensured all relevant structs and functions are compatible with the multi-process architecture.
- Updated nanvix and nanvixd to support the new multi-process feature, including changes to binary paths and startup logic.

### DIFF
--- a/.github/workflows/self-hosted-ci.yml
+++ b/.github/workflows/self-hosted-ci.yml
@@ -222,9 +222,9 @@ jobs:
       name: "Create Release Archive"
       if: ${{ success() && github.ref == 'refs/heads/dev' && matrix.build-type == 'release' &&  matrix.deployment-type != 'l2' }}
       run: |
-        single_process="no"
-        if [[ "${{ matrix.deployment-type }}" == "single-process" ]]; then
-          single_process="yes"
+        single_process="yes"
+        if [[ "${{ matrix.deployment-type }}" == "multi-process" ]]; then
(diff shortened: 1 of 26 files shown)


csegarragonz:
@ppenna one small suggestion. Would it make sense to have two feature flags `single-process` and `multi-process`, instead of defining one as the `not()` of the other?

I think this will help with code readability, and would make it possible to add a third mode (not sure if needed, but we could consider having something like `multi-process-l2`, just brainstorming here). Then we can set the corresponding default feature in `Cargo.toml`.

In the `Makefile` we would then need to make sure that only one is set with (`SINGLE_PROCESS=yes` or `MULTI_PROCESS=yes`), and resolve the default.

As an additional layer of security, we can also add, in the `./src/lib.rs` or `./src/main.rs` of the crates that use the feature:
```rust
#[cfg(all(feature = "single-process", feature = "multi-process"))]
compile_error!("Features \"single-process\" and \"multi-process\" are mutually exclusive; enable only one.");
```

Another argument in favour of the two-feature approach is that it would make it much easier to change the default feature again.

PR #1239, opened 2026-01-08
[nanvix-sandbox] Make Control-Plane Addresses Independent Of Tenant Id

csegarragonz:
In this PR I modify the signature of the control-plane sockaddr builder to return an address that is indpendent of the tenant ID.

It was a logic error to include it in the first place, and it only works because we cache the control-plane socket address in the sandbox cache, so subsequent tenants would use a control-plane address with a previous tenant's id.


ppenna (approved):
LGTM

Issue #710, opened 2025-07-28
[microvm] MicroVM snapshots


LucasGdosR:
## Context

I highly recommend you watch my [video](https://youtu.be/S3QMsB1BVRE) on it.

It should be possible to pause execution of a user VM, take a snapshot of enough state to load it and continue execution from that point, and save the state to files. After the user VM has been paused, it should be possible to shut it down, or resume it, or migrate it to another VMM.

Traditional snapshotting features save the state of a VM. In Nanvix's case, however, there's distributed state, so we need a distributed snapshot. Distributed snapshots consist of the local state of each process, and the distributed channel state. We must store all of it.

## Proposed Solution

The recorded state must be consistent. This means that:
1. The user VM must pause execution, so while the snapshot is being created, its memory is unchanged, and its vCPU state is unchanged;
2. Every message that was sent by the user VM to the system VM and whose response was not received prior to pausing must be saved. If the message was processed by the system VM but was not received prior to pausing, we can save the response. If the message wasn't processed by the system VM, then we must save the message to be replayed after restoring / resuming;
3. If relevant, system VM local state must also be saved. For now, it isn't needed.

I propose the userVM should have 3 states: `pre-boot`, `running`, and `paused`. `pre-boot` is relevant for loading snapshots, so I'll focus on `paused` and `running` for creating snapshots. There's a protocol for transitioning from `running` to `paused` (`pause`), another for `paused` to `running` (`resume`), and another for creating snapshots while in the `paused` state (`snapshot`).

### Dependency graph

This is a dependency graph of all operations in the pause and snapshot protocols. The main point I want to get across with it is that there is a path that is started as part of the pause protocol, but we only await its completion as a part of the snapshot protocol. This is because the system VM can start doing the necessary work for taking snapshots at the start of the pause protocol, but it doesn't need to finish for the pause to be complete.

<img width="660" height="660" alt="Image" src="https://github.com/user-attachments/assets/8a9dc571-b2d2-47f7-991e-ef01243128f4" />

### Pause protocol

<img width="1350" height="728" alt="Image" src="https://github.com/user-attachments/assets/97fbd5e8-b48f-44aa-8188-bdd839f9f264" />

Concurrently:
async 0. `nanvixd` --pause--> system VM
0. `nanvixd` --pause--> user VM (I/O thread)

Synchronously (continuing from `nanvixd` --pause--> user VM):
1. I/O thread --pause--> main thread
2. Main thread writes to virtual memory (pause requested)
[...]

PR #1243, opened 2026-01-09
[uservm] Change snapshot protocol

### LINKED ISSUE #1054: [uservm] Review Snapshot Protocol
## Description

The major refactor in the uservm severely impacted the snapshot protocol. We need to review and fix it.

LucasGdosR:
# Overview

The pause protocol is now is much simpler. There's no need to flush messages back and forth, just pause the vCPU. In contrast, the snapshot protocol became more complex, and includes flushing, so we can save all messages that aren't captured in the virtual memory. Also, the system VM no longer cares about the pause/resume/snapshot protocols. It just keeps responding to messages as they arrive.

## Details

In the snapshot protocol, the main thread sends a marker that will travel through the channels. Since they're FIFO, it ensures that every message sent before the marker is delivered when it arrives back.

The I/O thread drains vCPU's channel before forwarding the marker to the system VM, so those messages arrive at the system VM first. Then, the marker follows the regular path: system VM → I/O thread → memory thread → vCPU thread. This ensures that, when the marker arrives at the vCPU thread, every outstanding message that follows this path was already received.

Theoretically we could buffer vCPU's outstanding messages to system VM rather than drain the channel, and send the marker first, but this would require storing two buffers, and in practice vCPU's channel will highly likely already be empty, since the I/O thread will keep polling the channel while the user VM is paused, before snapshot creation is requested.

## Fixes

Some closures were removed since snapshot creation was refactored to happen in the vCPU thread through message passing, instead of happening in the main thread / orchestrator. This change happened in The Major Refactor, but got fixed now. It is very likely the same will happen for loading snapshots in the near future.

There was another bug introduced in The Major Refactor (#1054 ). A loop was now introduced to fix the bug, making the vCPU thread stay in the paused state upon receiving a "create snapshot" command. Perhaps we should include a "shutdown" command, as the old `NOTE` says, so we can gracefully shutdown a paused user VM. Closes #1054.

### DIFF
--- a/src/libs/control-plane-api/src/lib.rs
+++ b/src/libs/control-plane-api/src/lib.rs
@@ -61,6 +61,16 @@ use ::syslog::error;
 pub enum NanvixdCommand {
     /// Shutdown.
     Shutdown,
+    /// Pause.
+    Pause,
+    /// CreateSnapshot.
+    CreateSnapshot,
+    /// Resume.
+    Resume,
+    /// LoadSnapshot.
+    LoadSnapshot,
+    /// Start.
+    Start,
 }
 
 ///
--- a/src/uservm/src/io_thread.rs
+++ b/src/uservm/src/io_thread.rs
@@ -228,9 +228,12 @@ impl IoThread {
                                 control_plane_buf_len = 0;
 
                                 match msg.cmd() {
-                                    NanvixdCommand::Shutdown => {
-                                        control_tx.send(IoControlCommand::Shutdown).await?;
-                                    },
+                                    NanvixdCommand::Shutdown => {control_tx.send(IoControlCommand::Shutdown).await?;},
+                                    NanvixdCommand::Pause => {control_tx.send(IoControlCommand::Pause).await?;},
+                                    NanvixdCommand::CreateSnapshot => {control_tx.send(IoControlCommand::CreateSnapshot).await?;},
+                                    NanvixdCommand::Resume => {control_tx.send(IoControlCommand::Resume).await?;},
+                                    NanvixdCommand::LoadSnapshot => {control_tx.send(IoControlCommand::LoadSnapshot).await?;},
+                                    NanvixdCommand::Start => {control_tx.send(IoControlCommand::StartMicroVm).await?;},
                                 }
                             }
                         },
@@ -246,15 +249,6 @@ impl IoThread {
                     match result {
                         Some(response) => {
                             match response {
-                                IoControlResponse::FlushInput => {
-                                    debug!("input flush completed");
-                                    // Messages are no longer buffer, nothing to flush. We should remove this.
-                                    control_tx.send(IoControlCommand::LinuxDaemonFlushed).await?;
-                                },
-                                IoControlResponse::FlushOutput => {
-                                    // Messages are no longer buffer, nothing to flush. We should remove this.
-                                    debug!("output flush completed");
-                                },
                                 IoControlResponse::MicroVmPaused => {
                                     debug!("microvm pause acknowledged");
                                 },
@@ -264,6 +258,15 @@ impl IoThread {
                                 IoControlResponse::Shutdown => {
                                     debug!("shutdown completed");
                                     break Ok(());
+                                },
+                                IoControlResponse::CreateSnapshotMarker => {
+                                    // Drain the vCPU's channel and forward it to system VM.
+                                    // TODO (#1241): for now we assume the channel is empty.
+                                    debug!("vcpu's output drained");
+
+                                    // Forward the marker to system VM.
+                                    // TODO (#1241): create a new syscall that echoes a marker.
+                                    debug!("marker forwarded to system VM");
                                 },
                             }
                         },
--- a/src/uservm/src/orchestrator.rs
+++ b/src/uservm/src/orchestrator.rs
@@ -90,8 +90,6 @@ pub struct Orchestrator {
     pause_microvm: Box<PauseFn>,
     /// Callback function to erase a pause request from the kernel's memory.
     resume_microvm: Box<ResumeFn>,
-    /// Callback function to create a snapshot.
-    _create_snapshot: Box<CreateSnapshotFn>,
     /// Callback function to load a snapshot.
     load_snapshot: Box<LoadSnapshotFn>,
 }
@@ -124,12 +122,11 @@ enum State {
 ///
 #[derive(PartialEq)]
 pub enum IoControlCommand {
-    _StartMicroVm,
-    _LoadSnapshotAndRun,
-    _PauseMicroVm,
-    _CreateSnapshot(String),
-    _ResumeMicroVm,
-    LinuxDaemonFlushed,
+    StartMicroVm,
+    LoadSnapshot,
+    Pause,
+    CreateSnapshot,
+    Resume,
     Shutdown,
 }
 
@@ -140,10 +137,9 @@ pub enum IoControlCommand {
 ///
 #[derive(Debug, PartialEq)]
 pub enum IoControlResponse {
+    CreateSnapshotMarker,
     MicroVmPaused,
     SnapshotCreated,
-    FlushOutput,
-    FlushInput,
     Shutdown,
 }
 
@@ -172,7 +168,7 @@ pub enum MemoryControlResponse {}
 ///
 #[derive(PartialEq)]
 pub enum VcpuControlCommand {
-    CreateSnapshot(String),
+    CreateSnapshot,
     Resume,
 }
 
@@ -206,7 +202,6 @@ impl Orchestrator {
         vcpu_control_tx: Sender<VcpuControlCommand>,
         pause_microvm: Box<PauseFn>,
         resume_microvm: Box<ResumeFn>,
-        create_snapshot: Box<CreateSnapshotFn>,
         load_snapshot: Box<LoadSnapshotFn>,
     ) -> Self {
         Self {
@@ -220,7 +215,6 @@ impl Orchestrator {
             vcpu_control_tx,
             pause_microvm,
             resume_microvm,
-            _create_snapshot: create_snapshot,
             load_snapshot,
         }
     }
@@ -386,7 +380,7 @@ impl Orchestrator {
         command: IoControlCommand,
     ) -> Result<ControlFlow<()>> {
         match command {
-            IoControlCommand::_StartMicroVm => {
+            IoControlCommand::StartMicroVm => {
                 if self.state == State::PreBoot {
                     // TODO: separate starting logic from `spawn()` and put it here
                     // This only makes sense when snapshots can already be loaded https://github.com/nanvix/nanvix/issues/948
@@ -395,29 +389,20 @@ impl Orchestrator {
                 }
                 Ok(Continue(()))
             },
-            IoControlCommand::_LoadSnapshotAndRun => {
+            IoControlCommand::LoadSnapshot => {
                 if self.state == State::PreBoot {
                     if let Err(e) = (self.load_snapshot)().await {
                         let reason: String =
-                            format!("LoadSnapshotAndRun: failed to load snapshot: {e:?}");
-                        error!("handle_command(): {reason}");
-                        anyhow::bail!(reason);
-                    }
-                    trace!("State: PreBoot -> Paused");
-
-                    // The Linux daemon should send messages to PreBoot VMMs by default,
-                    // so there's no need to tell it to resume sending messages.
-
-                    if let Err(e) = self.resume_protocol().await {
-                        let reason: String =
-                            format!("LoadSnapshotAndRun: failed to resume microvm: {e:?}");
+                            format!("LoadSnapshot: failed to load snapshot: {e:?}");
                         error!("try_receive_from_io_thread(): {reason}");
                         anyhow::bail!(reason);
                     }
+                    self.state = State::Paused;
+                    trace!("State: PreBoot -> Paused");
                 }
                 Ok(Continue(()))
             },
-            IoControlCommand::_PauseMicroVm => {
+            IoControlCommand::Pause => {
                 if self.state == State::Running
                     && let Err(e) = self.pause_protocol().await
                 {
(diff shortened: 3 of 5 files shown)


src/libs/control-plane-api/src/lib.rs
```diff
@@ -57,10 +57,21 @@ use ::syslog::error;
...
     /// Shutdown.
     Shutdown,
+    /// Pause.
+    Pause,
+    /// CreateSnapshot.
+    CreateSnapshot,
+    /// Resume.
+    Resume,
+    /// LoadSnapshot.
+    LoadSnapshot,
```
LucasGdosR: Where should the name / path come from for snapshots, both creation and loading? Should this be up to nanvixd? Should it be part of the command? Does this enum need to be an `u8`, or can it hold a `String`?

src/uservm/src/io_thread.rs
```diff
@@ -228,9 +228,15 @@ impl IoThread {
...
-                                    },
+                                    // TODO (in this PR): If all we're doing is forwarding,
+                                    // why not reuse the same Enum, change the channel type to it,
+                                    // and forward with no match-statement?
+                                    NanvixdCommand::Shutdown=>{control_tx.send(IoControlCommand::Shutdown).await?;},
+                                    NanvixdCommand::Pause => {control_tx.send(IoControlCommand::Pause).await?;},
+                                    NanvixdCommand::CreateSnapshot => {control_tx.send(IoControlCommand::CreateSnapshot).await?;},
+                                    NanvixdCommand::Resume => {control_tx.send(IoControlCommand::Resume).await?;},
+                                    NanvixdCommand::LoadSnapshot => {control_tx.send(IoControlCommand::LoadSnapshot).await?;},
+                                    NanvixdCommand::Start => {control_tx.send(IoControlCommand::StartMicroVm).await?;},
```
LucasGdosR: Can we merge these two enums? It seems like every control plane command could be forwarded from the I/O thread with no pattern matching, as long as the recipient matches.

src/uservm/src/io_thread.rs
```diff
@@ -246,25 +252,26 @@ impl IoThread {
...
                                 },
+                                IoControlResponse::CreateSnapshotMarker => {
+                                    // Drain the vCPU's channel and forward it to system VM.
+                                    // TODO (#1241): for now we assume the channel is empty.
+                                    debug!("vcpu's output drained");
+
+                                    // Forward the marker to system VM.
+                                    // TODO (#1241): this is not a syscall, only a marker. How
+                                    // should we adapt this so system VM understands it?
+                                    debug!("marker forwarded to system VM");
```
LucasGdosR, line 269: How should we handle sending a control message to linuxd through the channel that takes in syscalls? Do you have any suggestions on it? There's likely a `message_kind` field I could use to include a tag, and the system VM likely already parses it. Sounds good?

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -444,19 +444,26 @@ impl Vmm {
...
+            match self.inner.lock().await.control_rx.recv().await {
+                Some(VcpuControlCommand::Resume) => {
+                    // TODO (#1241): if buffered messages exist, we must ingest them as if they were
+                    // in the `input_queue`, writing them to virtual memory.
+                    break Ok(());
+                },
+                Some(VcpuControlCommand::CreateSnapshot) => {
+                    // TODO (in this PR): how should we name snapshot files? Can they come from nanvixd?
+                    self.handle_create_snapshot("default".to_string()).await?;
+                },
```
LucasGdosR: This links with the what I mentioned earlier in nanvixd commands having strings.

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -569,6 +582,14 @@ impl Vmm {
                 anyhow::bail!(reason)
             },
         }
+
+        // 2) Drain messages to buffer.
+        // TODO (#1241): This requires a significant refactor to get the `input_queue` channel here.
+        // It is currently accessed through `Emulator.stdin_fn`, but we want to read it without
+        // writing it to the virtual memory.
```
LucasGdosR: Not pressing, but how should we wire this to make the channel accessible here?

LucasGdosR (commented):
@ppenna preliminary points. Still in draft.

src/libs/control-plane-api/src/lib.rs
```diff
@@ -57,10 +57,21 @@ use ::syslog::error;
 /// Command
 ///
 #[derive(Debug, Clone, Copy, IntoPrimitive, PartialEq, TryFromPrimitive)]
+// TODO (in this PR): Must this be u8? Can `CreateSnapshot` and `LoadSnapshot` contain strings with the path?
```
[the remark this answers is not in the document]
ppenna: Let's keep this simple and support a static name for snapshots. further configuration can be discussed later.

src/uservm/src/vmm/microvm/mod.rs
ppenna: Keep it simple now: "snapshot.img" someting like that. Once we have an initial implementation we improve it for partial snapshots.

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -444,19 +444,26 @@ impl Vmm {
...
-                Ok(())
-            },
-            // NOTE: Should we add an option for shutting down? Like so:
-            // Some(VcpuControlCommand::Shutdown) => self.vcpu.poweroff(0),
-            None => {
-                let reason: String = "the vmm has disconnected".to_string();
-                error!("run(): {reason}");
-                anyhow::bail!(reason)
-            },
+        // We should only exit this loop when resuming execution.
```
ppenna: This comment is misleading. What if we fail?

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -535,6 +542,12 @@ impl Vmm {
     /// Upon success, returns empty. Otherwise, returns an error.
     ///
     pub async fn create_snapshot(&self, filepath: String) -> Result<()> {
+        // Creating a snapshot has 3 stages for the vCPU thread:
+        // 1) Save the virtual memory and the KVM state to files;
+        // 2) Drain every message and buffer them until we receive a marker;
+        // 3) Save the buffer to a file.
```
ppenna: Move long explanation to the function doc.

src/uservm/src/vmm/microvm/mod.rs
ppenna: Probably it is not making them accessible here :)

src/uservm/src/io_thread.rs
ppenna: Can you name the two enums that you are talking about? Is it `NanvixdCommand`, and `IoControlCommand`? If yes, think of it as now they seem the same, but `nanvixd` may contain other messages that are not only specific to "snapshotting". So no, it does not make sense merging them.

src/uservm/src/io_thread.rs
ppenna: You should not send control messages to linuxd from uservm. Control messages are only sent by nanvixd.

src/uservm/src/io_thread.rs
LucasGdosR: Yes, those are the enums. Even if the commands are not specific to snapshotting, they'll still follow the same route: `nanvixd` -> I/O thread -> main thread, right? Forwarding is I/O's job. It can pass every control message to the main thread blindly, and then the main thread matches the `NanvixdCommand`.

src/uservm/src/io_thread.rs
LucasGdosR: I phrased this poorly. Think of it as a special kind of ping or echo. It's just a token that should flow through the system. It will follow the exact same flow as a syscall. In fact, it can be a new syscall. It enters from the I/O thread, gets processed, gets returned to the I/O thread, forwarded to the memory thread, and forwarded to the vCPU thread, as regular syscalls. It flows like this:
B2 -> (B3 intentionally skipped) B4 -> B5 -> B6 -> B7 
image

src/uservm/src/vmm/microvm/mod.rs
LucasGdosR: How about instead of passing a function to the `Emulator` that receives from the channel and writes to the virtual memory, we pass just the channel, and have the `Emulator` handle writing to the virtual memory? Or pass two functions, one for receiving, and another for writing, and chain them to get the current behavior. That way, we can expose another `Emulator` method to receive from the channel and do something other than writing to virtual memory.

Now:
```
// Emulator.stdin_fn()
input_queue --------------------------------> vmem
```

Proposed:
```
// Emulator.write_stdin(Emulator.receive_input())
input_queue --------------------------------> vmem

// buffer = Emulator.receive_input()
input_queue ----> buffer

// Emulator.write_stdin(buffer)
buffer ----> vmem

// During snapshot protocol use Emulator.receive_input rather than the complete pipeline
---- XXX -------- vmem
   |
   |---> buffer 

// During resume protocol put the buffer into the end of the pipeline
... -------- vmem
               |
... buffer --->|

```

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -535,6 +542,12 @@ impl Vmm {
     /// Upon success, returns empty. Otherwise, returns an error.
     ///
     pub async fn create_snapshot(&self, filepath: String) -> Result<()> {
+        // Creating a snapshot has 3 stages for the vCPU thread:
```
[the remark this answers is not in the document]
LucasGdosR: This is intended. The path to the snapshot is built later in the function based on the name of the executable file. We will add a more robust solution to this later.

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -569,6 +582,14 @@ impl Vmm {
...
         }
+
+        // 2) Drain messages to buffer.
+        // TODO (#1241): This requires a significant refactor to get the `input_queue` channel here.
+        // It is currently accessed through `Emulator.stdin_fn`, but we want to read it without
+        // writing it to the virtual memory.
+
+        // 3) Save the buffer to a file.
+        // TODO (#1241): we must have the buffer in order to save it.
     }
```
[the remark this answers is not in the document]
LucasGdosR: There's no unreachable code, only comments. There's no explicit return, so even if code is later added, it will not be unreachable.

src/uservm/src/io_thread.rs
ppenna: If you want to move where the forwarding logic is done, I'm fine with that. If you want to merge enums, don't do it.

src/uservm/src/io_thread.rs
ppenna: Yes, this looks what we want to do.

src/uservm/src/vmm/microvm/mod.rs
ppenna: Exposing two functions looks what we want to achieve the required modularity.

LucasGdosR:
Updated the description. This is now ready for review. Note #710 also includes details on the snapshot protocol.

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -444,19 +444,25 @@ impl Vmm {
...
+        // We should not exit this loop when creating a snapshot.
+        loop {
+            match self.inner.lock().await.control_rx.recv().await {
+                Some(VcpuControlCommand::Resume) => {
+                    // TODO (#1241): if buffered messages exist, we must ingest them as if they were
+                    // in the `input_queue`, writing them to virtual memory.
+                    break Ok(());
+                },
+                Some(VcpuControlCommand::CreateSnapshot) => {
+                    self.handle_create_snapshot("default".to_string()).await?;
```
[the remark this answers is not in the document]
ppenna: +1

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -524,7 +530,12 @@ impl Vmm {
...
     /// # Description
     ///
-    /// Saves the virtual memory and the KVM state to files.
+    /// Saves the virtual memory, the KVM state, and the channel state to files.
+    ///
+    /// Creating a snapshot has 3 stages for the vCPU thread:
+    /// 1) Save the virtual memory and the KVM state to files;
+    /// 2) Drain every message and buffer them until we receive a marker;
+    /// 3) Save the buffer to a file.
     ///
```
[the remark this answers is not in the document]
LucasGdosR: Doing this seems like setting me up to fail remembering to update the documentation when the TODOs are done. I think it's better to leave the documentation ready for the implementation.

LucasGdosR:
Added a new commit that adds support to shutting down the user VM while in the paused state / `handle_pause()` loop.

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -330,7 +335,13 @@ impl Vmm {
 
                             break Ok(exit_status);
                         } else {
-                            Handle::current().block_on(self.handle_pause())?;
+                            let shutdown_requested: bool =
+                                Handle::current().block_on(self.handle_pause())?;
+                            if shutdown_requested {
+                                let exit_status: u16 = 0;
```
ppenna: We should make this a file-level constant. Name it `EXIT_SUCCESS`

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -420,43 +431,67 @@ impl Vmm {
 
         let stem: &OsStr = Path::new(filepath)
             .file_stem()
-            .unwrap_or(OsStr::new("default"));
+            .unwrap_or(OsStr::new(DEFAULT_SNAPSHOT_FILENAME));
 
         let vmem_filepath: PathBuf = snapshots_dir.join(stem).with_extension("vmem");
-        let kvm_filepath: PathBuf = snapshots_dir.join(stem).with_extension("kvm.json");
+        let kvm_filepath: PathBuf = snapshots_dir.join(stem).with_extension("kvm");
```
ppenna: We should have file-level constants for these as well.

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -420,43 +431,67 @@ impl Vmm {
...
     ///
-    /// Acknowledges a pause request and waits for the next command, either `Resume` or `CreateSnapshot`.
+    /// Acknowledges a pause request and waits for the next command, either `Resume` or
+    /// `CreateSnapshot`.
     ///
     /// # Returns
     ///
-    /// Upon success, return empty. Otherwise, returns an error.
+    /// Upon success, return true if shutdown was requested, and false if `Resume` was issued.
+    /// Otherwise, returns an error.
```
ppenna: This documentation is confusing. Did you mean:
- Return true => VM shutdown
- Return false => VM resume?

We should create an enum for that. Name it `VmPauseReturn`

src/uservm/src/vmm/microvm/mod.rs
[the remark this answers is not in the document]
ppenna: Yes, but it does not force one to actually screen through the function.

src/uservm/src/io_thread.rs
```diff
@@ -246,25 +249,25 @@ impl IoThread {
...
-                                IoControlResponse::FlushOutput => {
-                                    // Messages are no longer buffer, nothing to flush. We should remove this.
-                                    debug!("output flush completed");
-                                },
                                 IoControlResponse::MicroVmPaused => {
                                     debug!("microvm pause acknowledged");
                                 },
                                 IoControlResponse::SnapshotCreated => {
-                                    debug!("snapshot created");
+                                    debug!("snapshot creation acknowledged");
```
ppenna: Is this intentional? If so, it looks like we should rename `IoControlResponse::SnapshotCreated` to `IoControlResponse::SnapshotAcknowleged`

src/uservm/src/orchestrator.rs
```diff
@@ -625,36 +560,75 @@ impl Orchestrator {
...
+        }
+
+        // Receiving from the vCPU means it has saved the local state and the channel
+        // state as files.
+        match self.vcpu_control_rx.recv().await {
+            Some(VcpuControlResponse::SnapshotCreated) => {
+                trace!("Snapshot created");
+                self.io_control_tx
+                    .send(IoControlResponse::SnapshotCreated)
+                    .await?;
```
ppenna: We are missing to catch and log error here.

ppenna (commented):
I left some minor suggestions here.

LucasGdosR:
> I left some minor suggestions here.

@ppenna I have addressed them all. Ready to merge.

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -420,43 +457,66 @@ impl Vmm {
...
-            // NOTE: Should we add an option for shutting down? Like so:
-            // Some(VcpuControlCommand::Shutdown) => self.vcpu.poweroff(0),
-            None => {
-                let reason: String = "the vmm has disconnected".to_string();
-                error!("run(): {reason}");
-                anyhow::bail!(reason)
-            },
+        // We should not exit this loop when creating a snapshot.
+        loop {
+            let mut inner: MutexGuard<'_, InteriorMicroVmHandle> = self.inner.lock().await;
```
ppenna: Holding the lock and invoking `.await` and `drop` is a very dangerous programming pattern.  Why do we actually need it? Cant it be refactored?

LucasGdosR: Refactored into an outer async block.

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -420,43 +457,64 @@ impl Vmm {
...
+                            error!("handle_pause(): {reason}");
+                            anyhow::bail!(reason)
+                        },
+                    }
+                }
+                // Check shutdown flag for exiting the loop.
+                _ = async {
+                    while !SHUTDOWN.with(|s| s.load(Ordering::SeqCst)) {
+                        // Delay shutdown by at most 10%.
+                        tokio::time::sleep(SHUTDOWN_TIMEOUT / 10).await;
```
ppenna: This should be a file-level constant.

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -330,7 +361,13 @@ impl Vmm {
 
                             break Ok(exit_status);
                         } else {
-                            Handle::current().block_on(self.handle_pause())?;
+                            let pause_return: VmPauseReturn =
+                                Handle::current().block_on(self.handle_pause())?;
```
ppenna: What if we fail?

src/uservm/src/orchestrator.rs
```diff
@@ -824,35 +800,26 @@ mod tests {
         let handle: JoinHandle<Result<()>> = h.orchestrator.spawn();
 
         // Move to Running state.
-        h.io_cmd_tx.send(IoControlCommand::_StartMicroVm).await?;
+        h.io_cmd_tx.send(IoControlCommand::StartMicroVm).await?;
```
ppenna: What if we fail?

src/uservm/src/orchestrator.rs
```diff
@@ -824,35 +800,26 @@ mod tests {
...
         // Request pause.
-        h.io_cmd_tx.send(IoControlCommand::_PauseMicroVm).await?;
+        h.io_cmd_tx.send(IoControlCommand::Pause).await?;
```
ppenna: What if we fail?

src/uservm/src/orchestrator.rs
```diff
@@ -824,35 +800,26 @@ mod tests {
...
-        let r3 = recv_io_resp(&mut h.io_resp_rx).await;
-        assert!(matches!(r3, IoControlResponse::MicroVmPaused));
+        // Expect MicroVmPaused.
+        let r = recv_io_resp(&mut h.io_resp_rx).await;
+        assert!(matches!(r, IoControlResponse::MicroVmPaused));
         assert!(h.pause_called.load(Ordering::SeqCst));
 
         // Resume.
-        h.io_cmd_tx.send(IoControlCommand::_ResumeMicroVm).await?;
+        h.io_cmd_tx.send(IoControlCommand::Resume).await?;
```
ppenna: What if we fail?

LucasGdosR:
I'm closing this one and working on my [wip-lucas](https://github.com/nanvix/nanvix/tree/wip-lucas) branch.

PR #1253, opened 2026-01-14
[nanvix-sandbox-cache] Prefill Network Namespace Pool

csegarragonz:
In this PR I configure the sandbox cache to prefill a network namespace pool upon nanvixd start-up. This applies only to HTTP mode.

I expose the size of the pool as a CLI argument to `nanvixd`, `-netns-pool-size` that has a default value of `128`. Setting the value to `0` reverts to the `Lazy` initialization strategy that we were using until now.

In the process I fixed two bugs:
* In nanvixd HTTP mode, the sandbox cache must be initialized _before_ binding to the HTTP server's listener socket. Otherwise clients will try to send requests before nanvixd is ready. I should say nanvixd, given its conservative locking, was resilient to such a situation, but produced counterintuitive results.
* When prefilling the netns pool, we were dropping the handle immediately after creation, which meant that we were effectively only creating one new namespace, and reusing it for subsequent invocations.

Creating `128` network namespaces upfront takes a bit over 10 seconds in my dev server.

### DIFF
--- a/scripts/run-nanvixd.sh
+++ b/scripts/run-nanvixd.sh
@@ -648,6 +648,7 @@ main() {
         -http-addr "${NANVIXD_SOCKADDR}" \
         -toolchain-bin-dir "${TOOLCHAIN_BIN_DIR}" \
         -log-dir "${logs_dir}" \
+        -netns-pool-size "0" \
         "$([ "$L2_VM" = "yes" ] && echo "-l2")" \
         -console-file "${console_file_name}" &
     NANVIXD_PID=$!
--- a/src/libs/nanvix-http/src/server.rs
+++ b/src/libs/nanvix-http/src/server.rs
@@ -110,9 +110,11 @@ impl<T: Send + Sync + Default + Clone + 'static> HttpServer<T> {
     /// describing what went wrong during server operation.
     ///
     pub async fn run(&mut self) -> Result<()> {
+        // Initialize the sandbox cache before binding the socket, as some setups may use socket
+        // readiness to probe nanvixd's readiness.
+        let sandbox_cache: Arc<Mutex<SandboxCache<T>>> = SandboxCache::new(self.config.clone())?;
         let mut signals: Signal = signal(SignalKind::interrupt())?;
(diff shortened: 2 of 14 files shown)


ppenna:
📋 **Failure summary**
- Workflow run 21041160633 (Jan 15, 2026 UTC) → job 60503073866 `Test (release microvm L2)` failed on the very first system test (`test-nanvixd-http-echo-c`).
- Branch `enhancement-prefill-netns`/PR #1253 only: `nanvixd` never opened 127.0.0.1:8181 during the new netns prefill path, so `run-nanvixd.sh` timed out after 70 s and CI aborted before any guest workload was launched.

🧪 **How to reproduce locally**
```
./z build --with-cached-options -- test-nanvixd-http-echo-c MACHINE=microvm RELEASE=yes L2_VM=yes
```
(Or rerun the failed job via the Actions UI.)

🧾 **Evidence**
1. Runner log [`logs/nanvixd-echo-c.elf/runner_2026_01_15_10_35.log`](https://github.com/nanvix/nanvix/actions/runs/21041160633) shows 70 s of socket polling followed by `[MAIN] ERROR: nanvixd failed to start` and cleanup of 13 namespaces (lines 13‑126). No guest binaries were invoked.
2. `nanvixd` trace log [`logs/nanvixd-echo-c.elf/nanvixd_2026-01-15_10-35-47.log`] ends right after creating namespaces `nvxns-0` through `nvxns-12` under the `prefill-128` strategy, with no subsequent kernel/uservm output, so the daemon appears to stall mid-prefill.

🧠 **Hypothesis / next steps**
- The new namespace prefill loop (prefill-128) is blocking `nanvixd` startup on release+L2, preventing the HTTP control plane from binding. Instrumenting or bounding `nanvix_sandbox::netns::init_namespace()` (e.g., per-namespace timeouts, extra logging) should clarify which `ip`/`tc` call is stuck.
- As a mitigation, reducing/guarding the prefill count on L2 runs would let CI progress while the root cause is addressed.

ppenna (approved):

PR #1287, opened 2026-01-23
[nanvixd] Collapse hashmaps into one

### LINKED ISSUE #1286: [nanvixd] Collapse SandboxCache `running_sandboxes` and `sandbox_index` into one
## Context

Currently:
```
pub struct SandboxCache<T> {
    ...
    /// Registry of all currently running sandboxes indexed by their unique tag.
    running_sandboxes: HashMap<SandboxTag, RunningSandbox>,
    /// Reverse index mapping User VM identifiers to their sandbox tags.
    sandbox_index: HashMap<UserVmIdentifier, SandboxTag>,
    ...
}
```

These two hashmaps can be combined into a single `HashMap<UserVmIdentifier, RunningSandbox>`. I have already double-checked in the codebase, and no `SandboxTag` information is ever needed when the `UserVmIdentifier` is available. The tag is not used when iterating over the keys and values of `running_sandboxes` either. This is safe to remove.

It is actually bug-prone to maintain both, as we need to always keep them in sync. Having a single hashmap removes this mental burden.

## Proposed Solution

```
pub struct SandboxCache<T> {
    ...
    /// Registry of all currently running sandboxes indexed by their unique User VM identifier.
    running_sandboxes: HashMap<UserVmIdentifier, RunningSandbox>,
    ...
}
```

This:
```
let tag: &SandboxTag = self.sandbox_index.get(&user_vm_id).ok_or_else(|| {...})?;
if let Some(sandbox) = self.running_sandboxes.remove(tag) {
    ...
```
Becomes:
```
if let Some(sandbox) = self.running_sandboxes.remove(&user_vm_id) {
    ...
```

LucasGdosR:
Every use of sandbox tags to access the hashmap is replaced by a use of the underlying unique `UserVmIdentifier`. This removes the need to keep two hashmaps in sync, and makes `new` and `kill` operations simpler (not to mention future operations), modifying a single hashmap rather than two. Overall, the code is less bug prone than it was.

Closes #1286

### DIFF
--- a/src/libs/nanvix-http/src/server.rs
+++ b/src/libs/nanvix-http/src/server.rs
@@ -154,11 +154,10 @@ impl<T: Send + Sync + Default + Clone + 'static> HttpServer<T> {
                     let mut cache_guard = sandbox_cache_clone.lock().await;
                     let summary: SandboxCacheStateSummary = cache_guard.state_summary();
                     info!(
-                        "shutdown snapshot: running_sandboxes={}, linuxd_instances={}, sandbox_index_entries={}, \
+                        "shutdown snapshot: running_sandboxes={}, linuxd_instances={}, \
                          control_plane_socket={}, l2_enabled={}",
                         summary.running_sandboxes(),
                         summary.linuxd_instances(),
-                        summary.sandbox_index_entries(),
                         summary.has_control_plane_bind_socket(),
                         summary.l2_enabled()
                     );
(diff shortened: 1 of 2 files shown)


ppenna (approved):
LGTM @LucasGdosR. I will take care of rebase and merging this.

Comment on this work, I understand that we should store the `SandboxTag` in `RunningSandbox`, as we may eventually query information about the program that a given UserVm is running and which tenant it belongs to. Could you create an issue for us to keep track of that?

LucasGdosR:
> LGTM @LucasGdosR. I will take care of rebase and merging this.
> 
> Comment on this work, I understand that we should store the `SandboxTag` in `RunningSandbox`, as we may eventually query information about the program that a given UserVm is running and which tenant it belongs to. Could you create an issue for us to keep track of that?

#1288

PR #1295, opened 2026-01-27
[kernel] Refactor I/O Memory Management

ppenna:
This pull request introduces a significant refactor and enhancement of the I/O memory region management system in the kernel. The changes replace the previous address-based allocation API with a new tag-based system, introduce explicit tracking of available and allocated regions, and ensure safe and automatic reclamation of regions when their handles are dropped. The new design improves safety, correctness, and debuggability of MMIO region allocation, and adds robust overlap and double-allocation checks.


ppenna:
Summary: Increased kernel stack to 64 KB for hyperlight to avoid early-boot kstack overflow that corrupted ProcessManager state and triggered the double-init panic. Fix is in build/kernel_config.toml.

Validation: LOG_LEVEL=trace RUST_LOG=trace bin/nanvix-test.elf test/test_multi_process.toml (hyperlight, local toolchain). All HTTP & terminal echo/hello/QuickJS/wasmd cases passed; ProcessManager panic gone.

Follow-up: opened issue #1313 to add guard pages/stack canaries/CI stack-usage gates to catch future kstack overflows.

PR #1305, opened 2026-01-30
[kernel] Heap Padding Calculation in init()

ppenna:
This PR fixes the heap padding calculation in init() to work on the corner cases when there is no heap padding because between the output data base and the kernel pool.


ppenna:
@danbugs you may want to check this.

PR #1339, opened 2026-02-06
Cast function pointers to usize via *const ()

jaylorch:
This eliminates a warning that in some cases blocks builds.


ppenna (approved):
LGTM

PR #1479, opened 2026-02-27
[libs] Fix bitmap alloc_range not wrapping around next_free hint

tangruize:
## Problem

`alloc_range()` scans from `next_free` to the end without wrap-around. When it skips free bits that don't form a contiguous range of the requested size, `next_free` advances past them. Those bits become unreachable, causing spurious `OutOfMemory` errors. For example: alloc 3 bits, free bit 1, `alloc_range(2)` skips bit 1 and advances `next_free` to 5, fill the rest, then `alloc(1)` returns `OutOfMemory` even though bit 1 is free.

This was discovered during Verus formal verification — the postcondition `exists_contiguous_free_range ==> Ok` could not be proven.

## Fix

Retry from position 0 when the scan from `next_free` reaches the end, stopping at the initial position. Preserves O(N) amortized fast path. Added a regression test.

### DIFF
--- a/src/libs/bitmap/src/lib.rs
+++ b/src/libs/bitmap/src/lib.rs
@@ -181,17 +181,34 @@ impl Bitmap {
             "bitmap length must match the number of bits"
         );
 
-        let mut start: usize = self.next_free;
+        let initial_start: usize = self.next_free;
+        let mut start: usize = initial_start;
+        let mut wrapped: bool = false;
 
-        // Traverse the bitmap until the last possible starting bit.
-        while start <= self.number_of_bits - size {
-            // Check for fast skip/ path.
+        // Traverse the bitmap, wrapping around once if needed.
+        loop {
+            // Stop condition: exceeded the last valid starting position.
+            if start > self.number_of_bits - size {
+                // If we haven't wrapped yet and started past 0, retry from beginning.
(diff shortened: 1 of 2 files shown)


src/libs/bitmap/src/lib.rs
```diff
@@ -219,6 +238,7 @@ impl Bitmap {
                 self.next_free = start + size;
                 return Ok(start);
             }
+            } // if !done
```
ppenna: Can you rollback this line? to keep coding standards.

src/libs/bitmap/src/lib.rs
```diff
@@ -181,17 +181,36 @@ impl Bitmap {
...
+                    wrapped = true;
+                } else {
+                    done = true;
+                }
+            }
+
+            // After wrap-around, stop if we've reached the initial position.
+            if !done && wrapped && start >= initial_start {
+                done = true;
+            }
```
ppenna, line 204: Could we use `break` and void use of `done` and the extra test condition?

tangruize: Sure, changing this to loop + break here is no problem. Note that the Verus verified version uses while !done because Verus loop doesn't propagate proof state across break, but that's a Verus-side concern and won't affect this fix. I'll push the update.

PR #1499, opened 2026-03-01
[kernel] Spin-then-halt idle for IKC

ppenna:
## Summary

Reduce PIT timer frequency from 10 kHz to 100 Hz and add a TSC-bounded spin-poll loop in the kernel handler that checks IKC credits before falling through to HLT.

This supersedes the approach in #1474 which reduced the timer frequency without decoupling IKC polling, causing ~10ms latency floors on all IKC-dependent benchmarks.

## Problem

PR #1474 reduced the timer from 10 kHz to 100 Hz to cut VM-exit overhead, but the kernel idle loop still uses HLT. Since IKC uses memory-mapped credits polling (no interrupt injection), the guest must wait up to 10ms (1/100 Hz) for the next timer interrupt before noticing new credits. This caused:
- `round-trip-latency` 32B: 1,867 µs → 8,880 µs (+375%)
- `warm-start-vmm` p50: 124 µs → 8,915 µs (+7089%)
- `echo-breakdown` `vmexit()` p50: 9,885 µs ≈ exactly 10ms

## Solution: Spin-Then-Halt

When the kernel has no work (no kcalls, no IKC messages, no zombies), instead of immediately yielding to HLT via `giveup()`, it now:

1. **Spin-polls** IKC credits using `PAUSE` + TSC for a configurable budget (`idle_spin_budget_us`, default 200 µs)
2. If credits arrive during the spin window → breaks back to the main loop to handle messages
3. If the budget expires → falls through to `giveup()` → HLT as before

This decouples IKC responsiveness from timer frequency: messages arriving within 200 µs are handled immediately, while truly idle periods still use power-efficient HLT.

## Changes

- **`build/kernel_config.toml`**: `timer_freq` 10000 → 100, `scheduler_freq` 1000 → 10 (maintaining 100ms quantum), add `idle_spin_budget_us = 200`
- **`src/libs/config/build.rs`**: Parse `idle_spin_budget_us` and generate `IDLE_SPIN_BUDGET_US` constant
- **`src/kernel/src/stdio.rs`**: Add `has_pending_messages()` — lightweight credits-only check without reading messages
- **`src/kernel/src/kcall/handler.rs`**: Spin-poll credits for budget duration before `giveup()` → HLT

## Expected Impact

| Metric | Before (10 kHz) | PR 1474 (100 Hz + HLT) | This PR (100 Hz + spin-then-halt) |
|---|---|---|---|
| Timer VM-exits/sec | 10,000 | 100 | 100 |
| IKC worst-case latency | 100 µs | 10 ms | ~200 µs (spin budget) |
| Host CPU during idle | Minimal (HLT) | Minimal (HLT) | 200 µs spin then HLT |

## Configuration

Set `idle_spin_budget_us = 0` in `kernel_config.toml` to disable spin-polling and revert to HLT-only behavior.

---
*Supersedes #1474.*

### DIFF
--- a/build/kernel_config.toml
+++ b/build/kernel_config.toml
@@ -65,10 +65,13 @@ kredzone_size = 128
 #==================================================================================================
 
 # Timer frequency (in Hz).
-timer_freq = 10000
+# Lower values reduce KVM VM-exit overhead from timer interrupts.
+# 100 Hz provides 10ms resolution with ~100x fewer VM exits than 10 kHz.
+timer_freq = 100
 
 # Scheduler frequency (in ticks).
-scheduler_freq = 1000
+# At 100 Hz timer, 10 ticks = 100ms preemption quantum.
+scheduler_freq = 10
 
 #==================================================================================================
 # Inter-Kernel / Inter-Process Communication
--- a/src/kernel/src/hal/platform/hyperlight/mod.rs
+++ b/src/kernel/src/hal/platform/hyperlight/mod.rs
@@ -137,7 +137,7 @@ pub(super) unsafe fn enable_interrupts() {
 /// It is safe to call this function only when the CPU is able to receive interrupts.
 ///
 pub(super) unsafe fn wait_for_interrupt() {
-    ::arch::cpu::halt();
+    ::arch::cpu::pause();
 }
 
 ///
--- a/src/kernel/src/hal/platform/microvm/mod.rs
+++ b/src/kernel/src/hal/platform/microvm/mod.rs
@@ -114,7 +114,7 @@ pub(super) unsafe fn enable_interrupts() {
 /// It is safe to call this function only when the CPU is able to receive interrupts.
 ///
 pub(super) unsafe fn wait_for_interrupt() {
-    ::arch::cpu::halt();
+    ::arch::cpu::pause();
 }
 
 ///
--- a/src/kernel/src/hal/platform/pc/mod.rs
+++ b/src/kernel/src/hal/platform/pc/mod.rs
@@ -152,7 +152,7 @@ pub(super) unsafe fn enable_interrupts() {
 /// It is safe to call this function only when the CPU is able to receive interrupts.
 ///
 pub(super) unsafe fn wait_for_interrupt() {
-    ::arch::cpu::halt();
+    ::arch::cpu::pause();
 }
 
 #[cfg(feature = "bios")]


ppenna:
## Benchmark Results (microvm, local virtualized env)

### Approach Comparison

I tested three approaches on this virtualized environment:

| Approach | warm-start-vmm p50 |
|---|---|
| dev baseline (10 kHz + HLT) | 285 µs |
| 100 Hz + HLT only (PR #1474) | ~8915 µs |
| 100 Hz + handler spin-then-halt (5ms budget) | 8422 µs |
| **100 Hz + PAUSE idle loop (this PR)** | **224 µs** |

### Why spin-then-halt failed

The handler-level spin-poll checked IKC credits via `read_volatile` + PAUSE for up to 5ms before falling through to HLT. Despite correct logic, it showed **zero improvement** over HLT-only.

Root cause: the VMM serializes guest execution with message delivery through shared `guest`/`vmem` mutex locks. While `KVM_RUN` is active, these locks are held, blocking the input thread from writing credits to guest memory. Credits only become visible after a VM exit releases the locks.

The PAUSE approach works because KVM's Pause-Loop Exiting (PLE) triggers frequent VM exits during the idle loop (`giveup()` → enable interrupts → PAUSE → return → check credits), giving the VMM opportunities to deliver messages between iterations.

### Full Results (dev vs this PR)

```
Benchmark          | dev p50 (µs) | PR p50 (µs) | Δ
-------------------+--------------+--------------+------
warm-start-vmm     |          285 |          224 | -21%
warm-start          |          657 |          739 | +12%
round-trip 32B     |          787 |          648 | -18%
round-trip 64B     |          634 |          664 |  +5%
round-trip 128B    |         1438 |         1747 | +21%
round-trip 512B    |         1583 |         1534 |  -3%
round-trip 4KiB    |         1579 |         1579 |   0%
boot-time           |        48396 |        48171 |   0%
cold-start          |        10883 |        11247 |  +3%
cold-start-uvm      |         8067 |         9202 | +14%
concurrent (4 VMs) |         9839 |        13791 | +40%
```

Note: local virtualized environment results are noisy. CI (bare-metal hosted) will provide more reliable comparison.

ppenna:
Closing — this PR is stale and the underlying problem has been solved differently on `dev`:

- **Timer reduction** (10 kHz → 100 Hz) landed via PR #1474 (fd52712cd).
- **IKC latency** was resolved by interrupt-driven wake-up (IRQ 9) in 281a977a3, which is strictly better than the spin-poll approach proposed here — instant wake-up with zero CPU overhead vs. 200µs busy-wait.

The spin-then-halt approach is no longer needed.

PR #1510, opened 2026-03-03
[kernel] Rollback in `mmio_alloc()` when MMIO page mapping fails

### LINKED ISSUE #1481: [kernel] Add rollback in mmio_alloc() when MMIO page mapping fails
## Summary
`ProcessManager::mmio_alloc()` maps MMIO pages one by one and carries a FIXME:
- "If we fail, we need to revert operation."

## Location
- `src/kernel/src/pm/process/manager/mod.rs`
- `mmio_alloc()`

## Problem
If mapping fails after partial progress, previously mapped pages are left modified, producing partial state and potential leaks/inconsistency.

## Proposed Fix
- Track successfully mapped MMIO pages during allocation.
- On failure, unmap/revert all pages mapped in the current call.
- Ensure `state.add_mmio(region)` is committed only after full success.
- Add tests for partial-failure rollback behavior.

## Acceptance Criteria
- `mmio_alloc()` is atomic (all-or-nothing).
- No partial MMIO mappings remain after an error.
- Regression test covers failure in the middle of the mapping loop.

cachebag:
`mmio_alloc()` previously applied `kctrl()` page-by-page without any pre-validation, so a failure mid-loop left earlier pages with modified permissions while `add_mmio()` was never called, producing partial state. This follows the dry-run pattern: a first pass validates that every page can be mapped without modifying any page table entries, then a second pass that actually applies the changes.

Closes #1481

### DIFF
--- a/src/kernel/src/mm/virt/vmem.rs
+++ b/src/kernel/src/mm/virt/vmem.rs
@@ -1177,6 +1177,15 @@ impl Vmem {
         vaddr: PageAligned<VirtualAddress>,
         access: AccessPermission,
     ) -> Result<(), Error> {
+        self.kctrl_impl(vaddr, access, false)
+    }
+
+    fn kctrl_impl(
+        &mut self,
+        vaddr: PageAligned<VirtualAddress>,
+        access: AccessPermission,
+        dry_run: bool,
+    ) -> Result<(), Error> {
         trace!("{vaddr:?}");
 
         // Check if the provided address lies outside the kernel space.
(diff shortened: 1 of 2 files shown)


ppenna:
@cachebag thanks for submitting this PR. The fix is logically correct, but it’s a bit fragile. I’d suggest following the dry‑run pattern we already use elsewhere in the codebase. The idea is straightforward: execute the operation once while disabling all side‑effecting actions, then run it again for real.

cachebag:
> @cachebag thanks for submitting this PR. The fix is logically correct, but it’s a bit fragile. I’d suggest following the dry‑run pattern we already use elsewhere in the codebase. The idea is straightforward: execute the operation once while disabling all side‑effecting actions, then run it again for real.

Understood. I followed the pattern I found in `copy_from_user_unaligned` and `copy_user_to_user`. 

Would there be no risk of TOCTOU gap here?

ppenna:
@cachebag thanks! Good question! TOCTOU cannot happen here because the kernel runs on uninterruptible mode.

PR #1532, opened 2026-03-04
fix(syscall): return ENOENT/EBADF instead of ENOTSUP in standalone mode

ppenna:
Path-based syscalls now return ENOENT and fd-based syscalls return EBADF in standalone mode, matching POSIX semantics Python expects.


ppenna:
Closing as stale. The `dev` branch restructured standalone syscall routing: standalone blocks now forward directly to VFS functions (e.g. `vfs_open`, `vfs_renameat`, `vfs_stat`) which return correct POSIX error codes, eliminating the separate `memfs`+`standalone` fallback pattern this PR modified. Rebasing would cause merge conflicts across all 10 files and effectively require a rewrite.

The one remaining fix (`readlinkat.rs` standalone error code: `OperationNotSupported` → `NoSuchEntry`) will be submitted as a fresh, minimal PR.

PR #1533, opened 2026-03-04
fix(nanvixd): route guest output to host stderr in single-process mode

ppenna:
Default console_file to None in single-process mode so guest output goes to host stderr instead of a log file.


ppenna:
Closing as stale. This draft PR has been open since March 4 and has diverged from `dev`. The changes can be re-proposed in a fresh PR if still needed.

PR #1536, opened 2026-03-04
fix(kernel): increase kernel heap size and fix slab allocator gap

ppenna:
Increase SLAB_COUNT from 32 to 128 (1MB to 4MB heap) to prevent OOM. Fix slab allocator to handle 513-4095 byte allocations that previously fell through.


ppenna:
Superseded by #1591

PR #1544, opened 2026-03-04
[nanvix-sandbox] Restore L2 VMs With Different Snapshot Dirs

### LINKED ISSUE #1203: [libs] Enhance L2 VM Logging Capabilities
## Context

Right now, whenever linuxd is spawned inside an L2 VM, it is always spawned from the same snapshot. This includes, among other things, the exact same console file, which is where the `linuxd` logs will go. This has two unfortunate consequences:
1. If you are using L2 mode, then you need to fetch linuxd's logs from `/tmp/clh-console` not the regular `./logs` directory, which is very unintuitive (hardcoded [here](https://github.com/nanvix/nanvix/blob/dev/scripts/generate-l2-snapshot.sh#L55)).
2. Even worse, when you are deploying multiple linuxd instances, all in L2 mode, they will all concurrently write to the same file.

Fixing this is conceptually easy, but will require a bit of work. As discussed in the past, each L2 VM will need to get its own sym-linked version of the snapshot directory, where the `config.json` file is actually modified and the console file points to a sound, unique, file.

csegarragonz:
## Context

A cloud-hypervisor snapshot consists of two config JSON files, and a heavyweight file with the state of the memory.

## Problem

All L2 VMs are restored from the same snapshot, which means that they log to the same file.
under high concurrency when restoring L2 VMs quasi in parallel this causes crashes (this is a hypothesis but it happens without L2 VMs)

The log file is specified in the console-file entry of the config.json file of the snapshot bundle

## Solution

Each L2 VM gets its own snapshot directory where two of the three files are a symbolic link to the parent one, and the third one is a copy with a different console-file entry.

The idea is to have a SnapshotDirHandle RAII handle that we give to each linuxd instance

Closes #1203

### DIFF
--- a/src/libs/nanvix-sandbox-cache/src/config.rs
+++ b/src/libs/nanvix-sandbox-cache/src/config.rs
@@ -69,7 +69,7 @@ pub struct SandboxCacheConfig<T> {
     log_directory: String,
     /// Flag indicating whether to deploy linuxd inside an L2 VM (using cloud-hypervisor).
     l2: bool,
-    /// Path to the snapshot file in an L2 deployment.
+    /// Path to the base snapshot file in an L2 deployment.
     l2_snapshot_path: String,
     /// Path to the temporary directory for Unix sockets and transient files.
     tmp_directory: String,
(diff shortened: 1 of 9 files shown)


ppenna:
@csegarragonz I am moving this to draft because I am orchestrating the CI due to the high pressure on it.

PR #1558, opened 2026-03-06
[daemons] Add Timeout To User VM Close Task

csegarragonz:
In this commit I modify the reaper task in linuxd that takes care of closing a user VM connection so that it does not block indefinitely.

Right now, an unresponsive or slow user VM can block the main linuxd loop indefintately, starving other resources. I guard against this behaviour in two different ways. First I add a timeout to the close_connection async. Second, when possible, I spawn a throwaway task that will do the (bounded-time) cleanup task in the background without affecting the main linuxd behaviour.


ppenna:
Tracked the unbounded `cmd_tx.send(Shutdown).await` blocking issue flagged in review as #1561.

PR #1560, opened 2026-03-06
[linuxd] Replace probe with GatewayReady

ppenna:
This pull request implements a more robust and explicit handshake protocol between the Linux daemon (`linuxd`) and the Nanvix daemon (`nanvixd`) to coordinate the readiness of gateway sockets for User VMs. The changes introduce a new control-plane message (`GatewayReady`) from `linuxd` to `nanvixd` and refactor the code to use this notification instead of relying on connection attempts to check readiness. The update also improves code structure and clarity by removing retry logic and making the communication protocol more explicit.


ppenna:
Adding @csegarragonz for visibility.

PR #1578, opened 2026-03-08
[nanvix-test] Graceful shutdown on interrupt

### LINKED ISSUE #1235: [nanvix-test] Gracefully Handle Signals to Prevent Host Dangling States
## Context

The `nanvix-test` utility currently lacks proper signal handling. When it receives a signal (e.g., SIGINT, SIGTERM), the application terminates abruptly. This can leave the host environment in an inconsistent or dangling state, especially if cleanup routines or resource deallocation steps are skipped.

## Problem

Without a signal-handling mechanism, unexpected interruptions during test execution may cause:

- Incomplete cleanup of temporary resources
- Stale processes or files on the host
- Potential interference with subsequent test runs or CI workflows

## Proposed Solution

Introduce a signal handler within `nanvix-test` to catch common termination signals and perform a best-effort graceful shutdown. At minimum, the handler should:

- Intercept relevant POSIX signals
- Trigger cleanup routines
- Terminate the program in an orderly manner

This enhancement ensures that test execution remains robust and that the host environment is left in a clean, well-defined state under all termination conditions.

## Source Code References

- [`nanvix-test`](https://github.com/nanvix/nanvix/tree/3851d0863479dcfbe1846095dd0b98ab0c304a00/src/utils/nanvix-test)

msalman-abid:
Closes #1235. This PR introduces graceful shutdown for `nanvix-test`, with the following changes:
- Use `CancellationToken` from `tokio-util` (new dependency) to propagate signals down to executors
- Wrap executors in http.rs and terminal.rs within a `tokio::select!` block to race against the cancellation token. Triggers `Drop` on cancellation.
- Wrap the body in `main()` to setup background listeners for  SIGINT and SIGTERM before starting to execute tests, On cancellation, return 128+signum. On repeat signal, terminate immediately.

AI Disclaimer:
I used Copilot AI assistance specifically in two instances:
- To draw up an execution plan for the issue to identify places where a `tokio::select!` block would be required.
- Setting up listeners for signals in `main()`, specifically lower level code like:
```Rust
    sig_flag.store(signum, Ordering::SeqCst);
    // ...
    let signum: i32 = received_signal.load(Ordering::SeqCst);
    if signum != 0 {
        process::exit(128i32.saturating_add(signum));
    }
```
This seemed like a particularly good use of Copilot because of the safeguard against instruction re-ordering in the CPU, guaranteeing that the `load()` sees the `store()`.

### DIFF
--- a/src/utils/nanvix-test/Cargo.toml
+++ b/src/utils/nanvix-test/Cargo.toml
@@ -17,6 +17,7 @@ nanvixd = { workspace = true, default-features = false }
 nanvix = { workspace = true, features = ["internal-api"] }
 no_fail = { workspace = true, features = ["std"] }
 tokio = { workspace = true, features = ["full"] }
+tokio-util = { workspace = true }
 chrono = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 toml = { workspace = true }
--- a/src/utils/nanvix-test/src/executor/empty.rs
+++ b/src/utils/nanvix-test/src/executor/empty.rs
@@ -123,27 +123,36 @@ pub(crate) async fn empty(
     iterations: usize,
     log_layout: &TestLogLayout,
     _extra_nanvixd_args: &[String],
+    cancellation_token: CancellationToken,
 ) -> Result<()> {
-    let log_root: &Path = Path::new(runner_config.log_directory.as_str());
-    let guest_log_tracker: GuestLogTracker = GuestLogTracker::capture(log_root)?;
+    tokio::select! {
+        result = async {
+            let log_root: &Path = Path::new(runner_config.log_directory.as_str());
+            let guest_log_tracker: GuestLogTracker = GuestLogTracker::capture(log_root)?;
 
-    for iteration in 0..iterations {
-        let RunnerLogPaths {
-            stdout: _stdout_file_path,
-            stderr: _stderr_file_path,
-        } = log_layout.allocate_runner_logs(Some(iteration));
+            for iteration in 0..iterations {
+                let RunnerLogPaths {
+                    stdout: _stdout_file_path,
+                    stderr: _stderr_file_path,
+                } = log_layout.allocate_runner_logs(Some(iteration));
 
-        ::log::info!("empty(): no-op iteration on non-Unix platform (iteration={iteration})");
+                ::log::info!("empty(): no-op iteration on non-Unix platform (iteration={iteration})");
 
-        guest_log_tracker.move_new_logs(log_layout.test_directory())?;
-        log_layout.normalize_component_logs(iteration)?;
+                guest_log_tracker.move_new_logs(log_layout.test_directory())?;
+                log_layout.normalize_component_logs(iteration)?;
+            }
+
+            guest_log_tracker.move_new_logs(log_layout.test_directory())?;
+            if iterations > 0 {
+                let last_iteration: usize = iterations - 1;
+                log_layout.normalize_component_logs(last_iteration)?;
+            }
+
+            Ok(())
+    } => result,
+        _ = cancellation_token.cancelled() => {
+            error!("empty(): cancellation requested");
+            Err(::anyhow::anyhow!("cancelled"))
+        },
     }
-
-    guest_log_tracker.move_new_logs(log_layout.test_directory())?;
-    if iterations > 0 {
-        let last_iteration: usize = iterations - 1;
-        log_layout.normalize_component_logs(last_iteration)?;
-    }
-
-    Ok(())
 }
--- a/src/utils/nanvix-test/src/executor/empty.rs
+++ b/src/utils/nanvix-test/src/executor/empty.rs
@@ -19,7 +19,9 @@ use crate::{
     },
 };
 use ::anyhow::Result;
+use ::log::error;
 use ::std::path::Path;
+use ::tokio_util::sync::CancellationToken;
 
 //==================================================================================================
 // Standalone Functions
@@ -49,44 +51,53 @@ pub(crate) async fn empty(
     iterations: usize,
     log_layout: &TestLogLayout,
     extra_nanvixd_args: &[String],
+    cancellation_token: CancellationToken,
 ) -> Result<()> {
-    let l2_enabled: bool = runner_config.l2_enabled;
-    let hwloc_file_path: Option<String> = runner_config.hwloc_file_path.clone();
-    let log_root: &Path = Path::new(runner_config.log_directory.as_str());
-    let guest_log_tracker: GuestLogTracker = GuestLogTracker::capture(log_root)?;
+    tokio::select! {
+        result = async {
+            let l2_enabled: bool = runner_config.l2_enabled;
+            let hwloc_file_path: Option<String> = runner_config.hwloc_file_path.clone();
+            let log_root: &Path = Path::new(runner_config.log_directory.as_str());
+            let guest_log_tracker: GuestLogTracker = GuestLogTracker::capture(log_root)?;
 
-    for iteration in 0..iterations {
-        let RunnerLogPaths {
-            stdout: stdout_file_path,
-            stderr: stderr_file_path,
-        } = log_layout.allocate_runner_logs(Some(iteration));
+            for iteration in 0..iterations {
+                let RunnerLogPaths {
+                    stdout: stdout_file_path,
+                    stderr: stderr_file_path,
+                } = log_layout.allocate_runner_logs(Some(iteration));
 
-        let nanvixd_http_args: NanvixdHttpArgs = NanvixdHttpArgs::new(
-            (stdout_file_path.as_path(), stderr_file_path.as_path()),
-            (runner_config.ipv4_addr.as_str(), runner_config.port_num),
-            hwloc_file_path.clone(),
-            l2_enabled,
-            runner_config.netns_pool_size,
-            log_layout.test_directory(),
-            extra_nanvixd_args,
-        )?;
+                let nanvixd_http_args: NanvixdHttpArgs = NanvixdHttpArgs::new(
+                    (stdout_file_path.as_path(), stderr_file_path.as_path()),
+                    (runner_config.ipv4_addr.as_str(), runner_config.port_num),
+                    hwloc_file_path.clone(),
+                    l2_enabled,
+                    runner_config.netns_pool_size,
+                    log_layout.test_directory(),
+                    extra_nanvixd_args,
+                )?;
 
-        {
-            let _nanvixd_handle: NanvixdHttp =
-                NanvixdHttp::spawn(runner_config, &nanvixd_http_args).await?;
-        }
+                {
+                    let _nanvixd_handle: NanvixdHttp =
+                        NanvixdHttp::spawn(runner_config, &nanvixd_http_args).await?;
+                }
 
-        guest_log_tracker.move_new_logs(log_layout.test_directory())?;
-        log_layout.normalize_component_logs(iteration)?;
+                guest_log_tracker.move_new_logs(log_layout.test_directory())?;
+                log_layout.normalize_component_logs(iteration)?;
+            }
+
+            guest_log_tracker.move_new_logs(log_layout.test_directory())?;
+            if iterations > 0 {
+                let last_iteration: usize = iterations - 1;
+                log_layout.normalize_component_logs(last_iteration)?;
+            }
+
+            Ok(())
+    } => result,
+        _ = cancellation_token.cancelled() => {
+            error!("empty(): cancellation requested");
+            Err(::anyhow::anyhow!("cancelled"))
+        },
     }
-
-    guest_log_tracker.move_new_logs(log_layout.test_directory())?;
-    if iterations > 0 {
-        let last_iteration: usize = iterations - 1;
-        log_layout.normalize_component_logs(last_iteration)?;
-    }
-
-    Ok(())
 }
 
 ///
--- a/src/utils/nanvix-test/src/executor/http.rs
+++ b/src/utils/nanvix-test/src/executor/http.rs
@@ -44,6 +44,7 @@ use ::std::{
     },
 };
 use ::tokio::time::timeout;
+use ::tokio_util::sync::CancellationToken;
 
 //==================================================================================================
 // Standalone Functions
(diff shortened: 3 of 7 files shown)


src/utils/nanvix-test/src/main.rs
```diff
@@ -5,22 +5,22 @@
...
+#![deny(clippy::char_lit_as_u8)]
+#![deny(clippy::fn_to_numeric_cast)]
+#![deny(clippy::fn_to_numeric_cast_with_truncation)]
+#![deny(clippy::ptr_as_ptr)]
+#![deny(clippy::unnecessary_cast)]
+#![deny(invalid_reference_casting)]
+#![deny(clippy::panic)]
+#![deny(clippy::unimplemented)]
+#![deny(clippy::todo)]
+#![deny(clippy::unreachable)]
```
ppenna: Please revert this. If you had to do it because of `#[tokio]` annotation in the `main()` function, perform a manual initialization of the runtime instead.

src/utils/nanvix-test/src/main.rs
```diff
@@ -118,7 +131,49 @@ fn main() -> Result<()> {
...
+    // Spawn a background task that listens for SIGINT and SIGTERM.
+    // First signal: cancel the token so in-flight work drains gracefully.
+    // Second signal: force-exit immediately.
+    let sig_flag: Arc<AtomicI32> = Arc::clone(&received_signal);
+    let sig_token: CancellationToken = cancellation_token.clone();
+    runtime.spawn(async move {
+        let mut sigint =
+            signal(SignalKind::interrupt()).expect("failed to register SIGINT handler");
+        let mut sigterm =
+            signal(SignalKind::terminate()).expect("failed to register SIGTERM handler");
```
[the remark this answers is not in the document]
ppenna: +1 no use of expect.

ppenna (changes requested):

src/utils/nanvix-test/Cargo.toml
```diff
@@ -17,6 +17,7 @@ nanvixd = { workspace = true }
 nanvix = { workspace = true, features = ["internal-api"] }
 no_fail = { workspace = true, features = ["std"] }
 tokio = { workspace = true, features = ["full"] }
+tokio-util = { workspace = true }
```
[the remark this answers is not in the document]
msalman-abid: The sync module in tokio-util 0.7.18 is unconditionally available (no feature gate). This is not needed.

src/utils/nanvix-test/src/main.rs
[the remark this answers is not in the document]
msalman-abid: Addressed

src/utils/nanvix-test/src/main.rs
msalman-abid: Reverted, sorry about that.
I had trouble when I was thinking of refactoring the test run loop to have less levels of indentations, and wrapping it in a `tokio::select!` block, and this fixed some of the compilation errors. Since I do not do that, we don't need this change anymore.

Although I am curious: the `.github/skills/coding-standards/SKILL.md` file says:
```md
- Prefer `#![deny(...)]` over `#![forbid(...)]` for unwrap/expect lints to allow scoped test
```
Are there certain places where `forbid(..)` is preferred over `deny(...)`?

src/utils/nanvix-test/src/executor/empty.rs
```diff
@@ -70,8 +78,13 @@ pub(crate) async fn empty(
         )?;
 
         {
-            let _nanvixd_handle: NanvixdHttp =
-                NanvixdHttp::spawn(runner_config, &nanvixd_http_args).await?;
+            let _nanvixd_handle: NanvixdHttp = tokio::select! {
+                result = NanvixdHttp::spawn(runner_config, &nanvixd_http_args) => result?,
+                _ = cancellation_token.cancelled() => {
+                    info!("empty(): cancelled during nanvixd spawn (iteration={})", iteration);
```
ppenna: Could you make this an error? As per project guidelines we always must log an error before returning `Err`

src/utils/nanvix-test/src/executor/http.rs
```diff
@@ -129,9 +132,24 @@ pub(crate) async fn test_with_http_executor(
 
     // Run tests within a scoped block to ensure logs are captured before moving them.
     {
-        let _nanvixd_handle: NanvixdHttp = NanvixdHttp::spawn(runner_config, &nanvixd_args).await?;
+        let _nanvixd_handle: NanvixdHttp = tokio::select! {
+            result = NanvixdHttp::spawn(runner_config, &nanvixd_args) => result?,
+            _ = cancellation_token.cancelled() => {
+                info!("test_with_http_executor(): cancelled during nanvixd spawn");
```
ppenna: Could you make this an error? As per project guidelines we always must log an error before returning `Err`

src/utils/nanvix-test/src/executor/http.rs
```diff
@@ -141,7 +159,13 @@ pub(crate) async fn test_with_http_executor(
                 l2_enabled,
             )?;
 
-            let mut user_vm: UserVm = UserVm::spawn(runner_config, &uservm_args).await?;
+            let mut user_vm: UserVm = tokio::select! {
+                result = UserVm::spawn(runner_config, &uservm_args) => result?,
+                _ = cancellation_token.cancelled() => {
+                    info!("test_with_http_executor(): cancelled during uservm spawn");
```
ppenna: Could you make this an error? As per project guidelines we always must log an error before returning `Err`

src/utils/nanvix-test/src/executor/terminal.rs
```diff
@@ -154,8 +166,13 @@ pub async fn test_with_terminal_executor(
             Duration::from_millis(runner_config.stream_collection_timeout_ms);
 
         let (mut nanvixd, stream_collectors) = {
-            let mut nanvixd: NanvixdTerminal =
-                NanvixdTerminal::spawn(runner_config, &nanvixd_terminal_args).await?;
+            let mut nanvixd: NanvixdTerminal = tokio::select! {
+                result = NanvixdTerminal::spawn(runner_config, &nanvixd_terminal_args) => result?,
+                _ = cancellation_token.cancelled() => {
+                    info!("test_with_terminal_executor(): cancelled during nanvixd spawn");
```
ppenna: Could you make this an error? As per project guidelines we always must log an error before returning `Err`

src/utils/nanvix-test/src/executor/terminal.rs
```diff
@@ -195,28 +212,52 @@ pub async fn test_with_terminal_executor(
...
+        let stdout_bytes: Vec<u8> = tokio::select! {
+            result = wait_stream_collector(
+                stdout_handle,
+                Arc::clone(&stdout_buffer),
+                collection_timeout,
+                "stdout",
+                iteration,
+            ) => result?,
+            _ = cancellation_token.cancelled() => {
+                info!(
```
ppenna: Could you make this an error? As per project guidelines we always must log an error before returning `Err`

src/utils/nanvix-test/src/executor/terminal.rs
```diff
@@ -195,28 +212,52 @@ pub async fn test_with_terminal_executor(
...
+        let stderr_bytes: Vec<u8> = tokio::select! {
+            result = wait_stream_collector(
+                stderr_handle,
+                Arc::clone(&stderr_buffer),
+                collection_timeout,
+                "stderr",
+                iteration,
+            ) => result?,
+            _ = cancellation_token.cancelled() => {
+                info!(
```
ppenna: Could you make this an error? As per project guidelines we always must log an error before returning `Err`

src/utils/nanvix-test/src/executor/terminal.rs
```diff
@@ -195,28 +212,52 @@ pub async fn test_with_terminal_executor(
...
 
         // Wait for the nanvixd process to exit and get its exit code.
         // FIXME (#1010): On hyperlight, SIGKILL terminates nanvixd without a clean exit, causing
         // wait_exit_code() to fail. When skip_exit_code_validation is true, we tolerate this
         // failure since we cannot reliably get the exit code anyway.
-        let exit_code: i32 = match nanvixd.wait_exit_code().await {
+        let exit_code: i32 = match tokio::select! {
+            result = nanvixd.wait_exit_code() => result,
+            _ = cancellation_token.cancelled() => {
+                info!("test_with_terminal_executor(): cancelled during wait_exit_code");
```
ppenna: Could you make this an error? As per project guidelines we always must log an error before returning `Err`

src/utils/nanvix-test/src/main.rs
```diff
@@ -118,7 +131,49 @@ fn main() -> Result<()> {
...
+        sig_flag.store(signum, Ordering::SeqCst);
+        sig_token.cancel();
+
+        // Wait for a second signal — force-exit.
+        tokio::select! {
+            _ = sigint.recv() => {},
+            _ = sigterm.recv() => {},
+        }
+        error!("signal_listener(): received second signal, forcing immediate exit");
+        process::exit(128i32.saturating_add(signum));
```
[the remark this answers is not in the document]
ppenna: +1 and we should replace `128`  with a constant for improve redability.

ppenna (changes requested):

ppenna (commented):
## PR Review: `[nanvix-test] Graceful shutdown on interrupt`

**Single commit:** `55f1cd2a5` — 93 files changed, +1161 / −1585

### Summary

This PR bundles **three unrelated changes** into one commit. They should ideally be separate PRs or at least separate commits:

1. **Graceful signal handling in `nanvix-test`** (the titular feature)
2. **Syscall library refactoring** (feature-gate restructuring across ~40 files)
3. **CI/release simplification** (remove memory-size matrix, manifest generation, benchmark data)

---

### 1. Graceful Shutdown in `nanvix-test` — Generally Good

**Approach:** Uses `tokio_util::CancellationToken` + a background signal listener task. First SIGINT/SIGTERM cancels the token for graceful drain; second signal calls `process::exit(128+signum)`.

**Positives:**
- Clean two-phase signal handling (graceful → forced).
- Token is properly threaded through all executor entrypoints (`empty`, `http`, `terminal`).
- Exit code follows Unix convention (`128 + signum`).

**Issues:**

- **`process::exit()` in the signal listener task** (`src/utils/nanvix-test/src/main.rs`): Calling `process::exit()` from a spawned task skips destructors and drop guards. If any executor holds resources (temp files, child processes, sockets), they won't be cleaned up on a second signal. Consider using `std::process::abort()` or propagating the exit back through the runtime instead.

- **`return Err(anyhow!("cancelled"))` vs `return Ok(())`** — inconsistent cancellation semantics. In `src/utils/nanvix-test/src/executor/empty.rs` and the HTTP/terminal executors, cancellation during spawn returns `Err("cancelled")`. But in `src/utils/nanvix-test/src/main.rs`, cancellation during `prepare_runner_environment` returns `Ok(())`. The caller in `main()` then calls `process::exit(128+signum)` anyway, so the distinction doesn't matter in practice, but the inconsistency is confusing.

- **Comment style** (`src/utils/nanvix-test/src/main.rs`): `// run the main test loop` — should be capitalized and end with a period per coding standards.

---

### 2. Syscall Feature-Gate Refactoring — Mostly Correct

The pattern across ~40 syscall files is:
1. Move VFS routing under `#[cfg(feature = "memfs")]` (was `#[cfg(feature = "standalone")]`)
2. Add standalone-mode early return with `OperationNotSupported`
3. Remove the separate `_linuxd` helper function, inlining the IPC code
4. Imports are unconditional (no more `#[cfg(not(feature = "standalone"))]` on import blocks)

**Issues:**

- **`compile_error!` guard removed** (`src/libs/syscall/src/lib.rs`): The old `#[cfg(all(feature = "standalone", not(feature = "memfs")))] compile_error!(...)` ensured standalone always has memfs. Removing it without explanation means invalid feature combinations silently compile. If the refactored code no longer requires this invariant, it should be documented; otherwise this looks like an accidental deletion.

- **Missing hex validation in `extract_commit_id`** (`src/libs/nanvix-registry/src/lib.rs`): The old implementation validated that the commit ID was a non-empty hex string. The new version accepts any characters between `"release-"` and the first dot. For example, `release-not-valid.tar.bz2` would return `"not-valid"` as a commit ID. Consider re-adding: `if !commit_id.chars().all(|c| c.is_ascii_hexdigit()) { return None; }`.

---

### 3. CI/Release Simplification

Removes the memory-size matrix dimension from CI builds, the manifest generation script, and benchmark baseline CSVs.

**Issues:**

- **Version downgrade: `0.12.170` → `0.12.163`** (`Cargo.toml`): The dev branch is at `0.12.170`. This PR sets the version back to `0.12.163`, which is a regression. This will cause conflicts on merge and breaks semantic versioning. This looks like an accidental rebase artifact — the branch was likely forked from an older dev commit.

- **Removing `DEBIAN_FRONTEND=noninteractive`** (`scripts/setup/ubuntu.sh`): This was there to prevent interactive prompts from blocking CI. Removing it risks hanging CI runs if a package prompts for input.

- **Removing `bzip2` from setup** (`scripts/setup/ubuntu.sh`): The Makefile still uses `tar -cjf` (bzip2 compression) for release archives. If `bzip2` isn't pre-installed on the runner image, release builds will fail.

---

### 4. linuxd Change

Removes a `write_all` of the IKC frame header before the error bytes in `src/daemons/linuxd/src/lib.rs`. This looks like a bug fix (double frame header write), but there's no explanation of why it's part of this PR or what bug it addresses.

---

### Verdict

The graceful-shutdown feature is well-implemented. However, this PR needs attention on:

1. **Rebase onto current `dev`** — version `0.12.163` → `0.12.170` regression
2. **Split into separate commits** — at minimum: signal handling, syscall refactoring, CI cleanup, linuxd fix
3. **Re-evaluate `compile_error!` removal** — is standalone without memfs now valid?
4. **Re-evaluate removal of `DEBIAN_FRONTEND` and `bzip2`** from setup script
5. **Re-add hex validation** in `extract_commit_id`

msalman-abid:
1. Replaced `info!` with `error!` where requested, along with a couple of other places where an `Err` is returned.
2. Track and return the second signal on force exit, Also use a constant instead of a magic number (128)

I'm not sure why the PR review comment describes out of scope changes -- this PR should only have 1 commit and a +217 -41 diff.

src/utils/nanvix-test/src/main.rs
```diff
@@ -118,7 +133,61 @@ fn main() -> Result<()> {
...
+        // Wait for a second signal — force-exit.
+        let (second_signum, second_signame): (i32, &str) = tokio::select! {
+            _ = sigint.recv() => (libc::SIGINT, "SIGINT"),
+            _ = sigterm.recv() => (libc::SIGTERM, "SIGTERM"),
+        };
+        error!(
+            "signal_listener(): received second signal {second_signame}, forcing immediate exit"
+        );
+        process::exit(BASE_RETURN_CODE.saturating_add(second_signum));
+    });
```
ppenna: `process::exit()` from a spawned task skips destructors and drop guards. If any executor holds resources (temp files, child processes, sockets), they will not be cleaned up on a second signal. Consider using `std::process::abort()` or propagating the exit back through the runtime instead.

src/utils/nanvix-test/src/main.rs
```diff
@@ -118,7 +133,61 @@ fn main() -> Result<()> {
...
+
+    // run the main test loop
+    let result: Result<()> = runtime.block_on(run(cancellation_token));
```
ppenna: Comment should be capitalized and end with a period per coding standards.
```suggestion
    // Run the main test loop.
```

src/utils/nanvix-test/src/main.rs
```diff
@@ -229,14 +298,19 @@ async fn run() -> Result<()> {
...
+        _ = prepare_runner_environment(
+            runner_config.l2_enabled,
+            runner_config.port_num,
+            Path::new(runner_config.tmp_directory.as_str()),
+            runner_config.tcp_cleanup_max_wait_seconds,
+            runner_config.tcp_cleanup_poll_interval_seconds,
+        ) => {},
+        _ = cancellation_token.cancelled() => {
+            info!("main(): cancelled during environment preparation");
+            return Ok(());
```
ppenna: Inconsistent cancellation semantics: this returns `Ok(())` on cancellation, while all executor cancellations return `Err(anyhow!("cancelled"))`. Consider making this consistent — either all cancellation paths return `Err` or all return `Ok`.

src/utils/nanvix-test/src/main.rs
msalman-abid: It looks like `process::abort()` behaves similarly; the program is aborted immediately and does not call destructors.

As an alternative. does a new `force_exit` cancellation token sound viable? It could race against `runtime.block_on(run())` and drop the runtime on a second signal. It could look like:

```diff
// ...
        // Wait for a second signal — record it and cancel the force-exit token.
        let (second_signum, second_signame): (i32, &str) = tokio::select! {
            _ = sigint.recv() => (libc::SIGINT, "SIGINT"),
            _ = sigterm.recv() => (libc::SIGTERM, "SIGTERM"),
        };
        error!(
            "signal_listener(): received second signal {second_signame}, forcing immediate exit"
        );
-       process::exit(BASE_RETURN_CODE.saturating_add(second_signum));
+       sig_flag.store(second_signum, Ordering::SeqCst);
+       sig_force_token.cancel();
    });

    // Run the main test loop; abort early if a second signal forces exit.
-   let result: Result<()> = runtime.block_on(run(cancellation_token));
+    let result: Result<()> = runtime.block_on(async {
+        tokio::select! {
+            result = run(cancellation_token) => result,
+            _ = force_exit_token.cancelled() => {
+             error!("main(): forcing exit on repeat signal...");
+             return Err(::anyhow::anyhow!("cancelled"));
+            },
+        }
+    });

    // If a signal was caught, override the exit code to 128+signum.
    let signum: i32 = received_signal.load(Ordering::SeqCst);
    if signum != 0 {
+       drop(runtime);
        process::exit(BASE_RETURN_CODE.saturating_add(signum));
    }
```

src/utils/nanvix-test/src/main.rs
msalman-abid: Addressed!

src/utils/nanvix-test/src/main.rs
msalman-abid: Addressed!

src/utils/nanvix-test/src/executor/http.rs
```diff
@@ -129,9 +131,24 @@ pub(crate) async fn test_with_http_executor(
...
     // Run tests within a scoped block to ensure logs are captured before moving them.
     {
-        let _nanvixd_handle: NanvixdHttp = NanvixdHttp::spawn(runner_config, &nanvixd_args).await?;
+        let _nanvixd_handle: NanvixdHttp = tokio::select! {
+            result = NanvixdHttp::spawn(runner_config, &nanvixd_args) => result?,
+            _ = cancellation_token.cancelled() => {
+                error!("test_with_http_executor(): cancelled during nanvixd spawn");
+                return Err(::anyhow::anyhow!("cancelled"));
+            },
+        };
```
ppenna: This is counter intuitive. Cancellation strategy should wrap the entire execution, not only the spawn of `nanvixd`, no?

src/utils/nanvix-test/src/executor/terminal.rs
```diff
@@ -154,8 +165,13 @@ pub async fn test_with_terminal_executor(
             Duration::from_millis(runner_config.stream_collection_timeout_ms);
 
         let (mut nanvixd, stream_collectors) = {
-            let mut nanvixd: NanvixdTerminal =
-                NanvixdTerminal::spawn(runner_config, &nanvixd_terminal_args).await?;
+            let mut nanvixd: NanvixdTerminal = tokio::select! {
```
ppenna: Same comment as before.

src/utils/nanvix-test/src/executor/terminal.rs
```diff
@@ -195,28 +211,52 @@ pub async fn test_with_terminal_executor(
...
+            ) => result?,
+            _ = cancellation_token.cancelled() => {
+                error!(
+                    "test_with_terminal_executor(): cancelled during stderr collection \
+                     (iteration={})",
+                    iteration
+                );
+                return Err(::anyhow::anyhow!("cancelled"));
+            },
+        };
```
ppenna: If we wrap the entire test case on select, this pattern goes away.

ppenna (changes requested):

src/utils/nanvix-test/src/executor/http.rs
msalman-abid: Addressed. I'm inclined to agree.

It seemed like a good idea at the time to keep error messages specific and only wrap longer running functions (such as spawn or receive_payload, and rely on the `if()` at the start of the loop to catch any signals in-between, but this may be an anti-pattern. I have wrapped all of the execution in the `select!` block now.

msalman-abid:
I've changed the cancellation to wrap the entire testcase instead of specific functions internally. It introduces additional levels of indentation to each executor file, which may be undesirable, but the linter doesn't seem to complain. Let me know if we want to refactor this.

src/utils/nanvix-test/src/executor/terminal.rs
msalman-abid: Addressed! I wrapped all of the execution to check for cancellation. The logs and return codes look ok when I test this with different signals.

msalman-abid:
@ppenna I'm marking this as a draft PR -- I need to rebase after new changes for windows support (#1783) and resolve conflicts.

msalman-abid:
Changes since the last review:
1. Enclose all of the executor logic within a `tokio::select!` block
2. Add `#[cfg(unix)]` to linux-specific signals and imports
3. Add a `#[cfg(not(unix))]` block in `main()` to listen for Windows' ctrl+c signal

Sorry for the delay. I can reiterate on this sooner now if we have more changes to make.

src/utils/nanvix-test/src/main.rs
```diff
@@ -122,7 +138,102 @@ fn main() -> Result<()> {
...
+        });
+    }
+
+    // Run the main test loop. Abort early if a second signal forces exit.
+    let result: Result<()> = runtime.block_on(async {
+        tokio::select! {
+            result = run(cancellation_token) => result,
+            _ = force_exit_token.cancelled() => {
+                error!("main(): forcing exit on repeat signal...");
+                Err(::anyhow::anyhow!("force exit"))
```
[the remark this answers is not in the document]
msalman-abid: We said that a `process::exit()` wouldn't work for us in an earlier review (https://github.com/nanvix/nanvix/pull/1578#discussion_r2936955735). Let me know if we want to keep the `drop()` or go with this suggestion.

ppenna:
Queueing this for review.

msalman-abid:
bump. Rebased with the new cargo.toml.

src/utils/nanvix-test/src/main.rs
```diff
@@ -85,12 +85,28 @@ use ::log::{
...
+    SignalKind,
+    signal,
 };
+use ::tokio_util::sync::CancellationToken;
 
 //==================================================================================================
 // Constants
 //==================================================================================================
 
+// Base return code for test interrupts; actual exit code will be 128 + signal number.
```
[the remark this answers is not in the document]
ppenna: +1

src/utils/nanvix-test/src/main.rs
```diff
@@ -122,7 +138,102 @@ fn main() -> Result<()> {
...
+            sig_flag.store(2, Ordering::SeqCst);
+            sig_token.cancel();
+
+            // Second Ctrl+C: force immediate exit.
+            if tokio::signal::ctrl_c().await.is_err() {
+                return;
+            }
+            error!("signal_listener(): received second Ctrl+C, forcing immediate exit");
+            sig_flag.store(2, Ordering::SeqCst);
+            sig_force_token.cancel();
```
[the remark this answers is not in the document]
ppenna: +1

msalman-abid:
Changes:
1. Fix comment prefix to `///`
2. Move Windows exit code magic number to named constant: `WINDOWS_CTRL_C_SIGNUM`
3. Log error in branch `if tokio::signal::ctrl_c().await.is_err()`

PR #1585, opened 2026-03-09
Build x86_64 C cross-compiler in Docker toolchain image

rfonseca:
## Summary
- extend `scripts/setup/toolchain.sh` so `build_tool()` can forward an optional `--target`
- build binutils, GCC stage0/stage1, and newlib for both `i686` and `x86_64`
- install both cross-compiler families in the same Docker toolchain image
- keep LLVM, Rust, Python, and Cloud Hypervisor as shared host tools
- copy the `x86_64-nanvix` sysroot into `Dockerfile.optimized` and verify both targets there

## Notes
- This PR is a direct cherry-pick of local commit `54e4cfbd7b9833b0995d5464150ea4f62927ea47` onto `origin/dev`.


rfonseca:
Closing this incorrect dev-based PR; replacing it with a PR based on feature-kernel-x64 that includes the full stack through 54e4cfbd7.

PR #1595, opened 2026-03-10
[nanvix-sandbox-cache] Split Locking Around Cache

### LINKED ISSUE #1293: [nanvixd] Sandboxes are never cached in `SandboxCache::get()`
## Description

The `SandboxCache::get()` method takes in the arguments `tenant_id`, `program`, `app_name`, and `program_args`. These arguments are then passed to `SandboxTag::new()` to create a tag. This constructor always creates a random `sandbox_id`, a `UserVmIdentifier`. Thus, even if every argument was the same, due to the nature of random `UserVmIdentifier`s, the tag key would (almost) never match an existing key.

This has become more obvious now that the hashmap key has been substituted from `SandboxTag` to `UserVmIdentifier`, but the bug predates this change. It was never detected earlier, since `SandboxCache::get()` has always been used for creating new sandboxes, and never for actually retrieving an existing sandbox.

## Expected Behavior

The `SandboxCache::get()` method should take the entire tag as an argument, or at least receive another `UserVmIdentifier` optional argument. If the `UserVmIdentifier` is `Some` and it matches an existing sandbox, then the sandbox should be returned from the cache, otherwise it should be created.

csegarragonz:
In this PR I address a serialization issue that affects concurrent creation of user VMs under load.

The main issue is that each request to `get()` a sandbox from `SandboxCache` acquires a full, global, lock on the cache, effectively serializing all requests. In reality, the only requests that need to be serialized are requests belonging to the same tenant whose linuxd instance has not yet been initialized. This can be achieved by removing the global lock from the `SandboxCache` and adding individual read-write locks to the `running_sandboxes` and `tenants` maps, indexing running user VMs and linuxd instances, respectively.

The only caveat to the above is that, even after this PR, requests to create new user VMs are still, effectively, serialized due to the global lock on the control-plane listener socket to await for the `GatewayReady` notifications. In a follow-up PR we should also get rid of this global lock, and instead implement a background task that owns the listener socket and that can interact with the other tasks spawning user VMs and linuxd instances. See tracking issue #1610.

In the process I address #1293 by adding the option to retrieve a running user VM by id.

Closes #1293

### DIFF
--- a/src/libs/nanvix-http/src/client/multi_process.rs
+++ b/src/libs/nanvix-http/src/client/multi_process.rs
@@ -40,7 +40,6 @@ use ::std::{
     sync::Arc,
 };
 use ::syscomm::SocketType;
-use ::tokio::sync::Mutex;
 use ::user_vm_api::UserVmIdentifier;
 
 //==================================================================================================
@@ -63,7 +62,7 @@ use ::user_vm_api::UserVmIdentifier;
 ///
 pub(crate) struct HttpClient<T> {
     /// Shared handle to the sandbox cache for managing sandboxes.
-    sandbox_cache: Arc<Mutex<SandboxCache<T>>>,
+    sandbox_cache: Arc<SandboxCache<T>>,
 }
 
 //==================================================================================================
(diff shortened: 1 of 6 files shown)


ppenna:
@csegarragonz you may want to check latest commits I pushed here.

PR #1599, opened 2026-03-10
[build] Use attributes instead of `verus!`

jaylorch:
This pull request enhances the `bitmap` library with improved support for Verus verification attributes and updates the code to use the latest conventions for specifying verification conditions. The most significant changes include the addition of new nightly Rust features for Verus, refactoring of specification annotations for better clarity and maintainability, and updating dependencies.

### DIFF
--- a/src/libs/bitmap/src/lib.spec.rs
+++ b/src/libs/bitmap/src/lib.spec.rs
@@ -338,4 +338,57 @@ impl BitmapView {
     }
 }
 
+//==================================================================================================
+// View Implementation for Bitmap
+//==================================================================================================
+
+impl View for Bitmap {
+    type V = BitmapView;
+
+    closed spec fn view(&self) -> BitmapView {
+        BitmapView {
+            num_bits: self.number_of_bits as int,
+            set_bits: Set::new(
+                |i: int| 0 <= i < self.number_of_bits as int && Self::bit_at(self.bits@, i),
+            ),
+        }
+    }
+}
+
+//==================================================================================================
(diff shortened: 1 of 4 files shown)


src/libs/bitmap/src/lib.spec.rs
```diff
@@ -338,4 +338,58 @@ impl BitmapView {
     }
 }
 
+//==================================================================================================
+// View Implementation for Bitmap
+//==================================================================================================
+
+#[cfg(verus_keep_ghost)]
```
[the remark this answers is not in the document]
jaylorch: Good idea, done.

src/libs/raw-array/src/lib.rs
```diff
@@ -196,50 +196,33 @@ impl<T> RawArrayStorage<T> {
...
-
 // External type specification for RawArrayStorage.
-#[verifier::reject_recursive_types(T)]
-#[verifier::external_type_specification]
-#[verifier::external_body]
+#[cfg(verus_keep_ghost)]
+#[verus_verify(reject_recursive_types(T))]
+#[verus_verify(external_type_specification)]
+#[verus_verify(external_body)]
 pub struct ExRawArrayStorage<T>(RawArrayStorage<T>);
```
ppenna, line 205: Are we missing `#[verus_verify]` here?

jaylorch: The other `#[verus_verify]` attributes cause it to be visible to the verifier, so it's not needed. But it wouldn't hurt to add it.

PR #1600, opened 2026-03-10
[scripts] Rename VERUS_DIR to VERUS_HOME

ppenna:
## Summary

Rename `VERUS_DIR` to `VERUS_HOME` and simplify custom Verus installation support. Instead of building Verus from source inside the setup script, users point `VERUS_HOME` at a directory containing pre-built or locally compiled Verus binaries.

## Changes

- **`Makefile`** — renamed `VERUS_DIR` → `VERUS_HOME`; `ensure-verus` validates the binary exists when a custom path is set, skips the automatic download.
- **`doc/build.md`** — updated usage examples to reference `VERUS_HOME`.
- **`doc/setup.md`** — added "Using a Custom Verus Installation" section.
- **`.github/skills/build/SKILL.md`** — updated override reference.

## Usage

```bash
# Default (prebuilt download, unchanged)
./z build -- verify

# Custom Verus installation (pre-built or source-built)
./z build -- verify VERUS_HOME=~/verus/target-verus/release
```

## Validation

- `make help` shows `VERUS_HOME` correctly.
- Custom `VERUS_HOME` with missing binary: clear error message.
- Custom `VERUS_HOME` with source-built Verus: **69 verified, 0 errors** on bitmap crate.
- Default prebuilt path: unchanged behavior.

### DIFF
--- a/.github/skills/build/SKILL.md
+++ b/.github/skills/build/SKILL.md
@@ -118,7 +118,7 @@ Nanvix uses Verus for formal verification of selected kernel crates. The expecte
 ```
 
 - The `ensure-verus` prerequisite downloads the correct Verus release automatically.
-- Override the install location with `VERUS_DIR=/path/to/verus`.
+- Override the install location with `VERUS_EXECUTABLE_DIR=/path/to/verus`.
 - The `vstd` crate version in `Cargo.toml` is exact-pinned (`=`) to match the Verus binary.
 
 ## Cleaning
--- a/Makefile
+++ b/Makefile
@@ -339,8 +339,12 @@ export SETCAP_CMD := setcap
 # Verus Formal Verification
 #===================================================================================================
 
-# Path to the Verus installation directory.
-export VERUS_DIR ?= $(TOOLCHAIN_DIR)/verus
+# Path to the directory containing the Verus executable.
+export VERUS_EXECUTABLE_DIR ?= $(TOOLCHAIN_DIR)/verus
+
+# Default Verus directory (used to detect custom VERUS_EXECUTABLE_DIR overrides).
+# patsubst strips trailing slashes so `toolchain/verus/` is not treated as custom.
+VERUS_DEFAULT_EXECUTABLE_DIR := $(patsubst %/,%,$(TOOLCHAIN_DIR)/verus)
 
 # List of crates to verify with Verus.
 VERUS_CRATES := bitmap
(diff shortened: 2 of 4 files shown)


jaylorch:
The name `VERUS_HOME` conveys (even more strongly than `VERUS_DIR` does) the incorrect impression that it should be set to the top-level directory of the Verus repo. How about `VERUS_EXECUTABLE_DIR` instead? Also, it would be good to make sure the script never does any writes or updates to the contents of that directory.

jaylorch:
It doesn't seem to be invoking the Verus executables in `VERUS_HOME`. I see it adds `VERUS_HOME` to the path, but that doesn't seem sufficient.

ppenna:
I validated this locally by pointing `VERUS_EXECUTABLE_DIR` at a source-built Verus (`~/src/verus/source/target-verus/release/`). The `PATH` prepend mechanism does work correctly — `cargo verus` discovers `cargo-verus` via `PATH`, and `cargo-verus` in turn invokes the `verus` rustc wrapper from the same directory.

Evidence from the build output:

```
Using custom Verus installation at /home/ppenna/src/verus/source/target-verus/release.
```

The executed process shows the custom binary was used:

```
/home/ppenna/src/verus/source/target-verus/release/verus /home/ppenna/.cargo/bin/rustc --crate-name vstd ...
```

Both binaries resolve to the custom directory:

```
$ PATH="…/target-verus/release:$PATH" which cargo-verus
/home/ppenna/src/verus/source/target-verus/release/cargo-verus
$ PATH="…/target-verus/release:$PATH" which verus
/home/ppenna/src/verus/source/target-verus/release/verus
```

The verification did fail, but due to a **version mismatch** (local Verus is `0.2026.03.10.13c14a1` while the project's `vstd` crate is pinned to `0.0.0-2026-02-08-0120`), not a path issue. The error `unexpected verus diagnostic item verus::verus_builtin::fin` is the newer Verus binary rejecting symbols from the older vstd library — expected behavior when versions diverge.

ppenna:
Follow-up: after updating the `vstd` crate version in `Cargo.toml` to match the local Verus build (`0.0.0-2026-03-08-0103`), verification passes successfully with the custom `VERUS_EXECUTABLE_DIR`:

```
$ make verify VERUS_EXECUTABLE_DIR=~/src/verus/source/target-verus/release

Using custom Verus installation at /home/ppenna/src/verus/source/target-verus/release.
verification results:: 69 verified, 0 errors
```

This confirms the `PATH` prepend mechanism works correctly for custom Verus installations. The earlier failure was purely a version mismatch between the Verus binary and the pinned `vstd` crate — not a path resolution issue.

ppenna:
Also verified on the `attributes-for-verus` branch (rebased on top of `enhancement-scripts-verus`). Verification passes with the custom Verus build:

```
$ make verify VERUS_EXECUTABLE_DIR=~/src/verus/source/target-verus/release

Using custom Verus installation at /home/ppenna/src/verus/source/target-verus/release.
verification results:: 69 verified, 0 errors
Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.71s
```

The `VERUS_EXECUTABLE_DIR` override works correctly across both branches.

jaylorch:
> It doesn't seem to be invoking the Verus executables in `VERUS_HOME`. I see it adds `VERUS_HOME` to the path, but that doesn't seem sufficient.

Oops, never mind. I see my issue. I had made changes not only to the Verus executable but also to the vstd library, and I wasn't seeing the latter because I hadn't updated Cargo.toml. (I foolishly thought I didn't need to update Cargo.toml since I was only changing the Verus executable.)

It is indeed using the Verus executable from that path.

jaylorch (approved):

jaylorch:
I suggest updating the title of this PR before merging.

PR #1640, opened 2026-03-14
[kernel] Prevent triple fault on stack overflow

ppenna:
## Summary

Fixes intermittent triple faults during `dlfcn-pie-c.elf` execution in standalone debug mode (~2-3% failure rate).

### Kernel (`src/kernel/src/pm/process/manager/mod.rs`)
- **Remove guard check from `do_exit()` and `do_exit_thread()`** — prevents the exception handler exit path from re-triggering the overflow detection. The check remains in `do_schedule()` and `do_sleep()`, which catch overflow at every context-switch point.

### UserVM (`src/uservm/src/vmm/microvm/`)
- **Map `VcpuExit::Shutdown` to `Halt`** instead of `Unknown` — ensures nanvixd terminates promptly on guest triple faults instead of hanging for 300s.
- **Log registers (ESP, EIP, CR2, CR3)** on shutdown for post-mortem diagnosis.

## Verification

- **Before fix**: ~2-3% triple fault rate over 500 runs of `dlfcn-pie-c.elf` in standalone debug mode.
- **After fix**: 0 triple faults in 500 runs. The ~2.6% remaining failures are clean graceful shutdowns (guard check detects overflow, logs error, halts VM).

### DIFF
--- a/src/kernel/src/hal/arch/x86/hooks.S
+++ b/src/kernel/src/hal/arch/x86/hooks.S
@@ -179,6 +179,44 @@
 
 .endm
 
+/*
+ * Stack overflow guard threshold.
+ *
+ * If ESP falls below this value when entering an exception handler, the
+ * kernel stack has overflowed into the kpool metadata region (page tables
+ * and page directory).  Continuing would corrupt the page directory and
+ * trigger an unrecoverable recursive exception cascade (triple fault).
+ *
+ * Instead we shut the VM down immediately via the VMM I/O port so the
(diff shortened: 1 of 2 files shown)


ppenna:
Closing — all proposed changes are superseded:
- Guard check removal: root cause fixed by #1655 (`cld`); proper dynamic guard added in #1663.
- Shutdown mapping: already merged to dev.
- Register dump: merged via #1646.

Issue #1641, opened 2026-03-14
[kernel] Missing cld in exception entry causes triple fault


ppenna:
## Summary

Intermittent kernel crash (triple fault) during the `dlfcn-pie-c.elf` test, occurring in ~2–4%
of debug-mode runs across all deployment modes. The root cause is a **missing `cld` instruction**
at kernel exception entry: user-mode code may set the x86 direction flag (DF=1 via `std`), and
the exception handler inherits it. Compiler-generated `rep movsl` instructions in the Rust handler
path then copy memory **backwards**, corrupting the kpool metadata region and triggering a triple fault.

## Environment

- **Deployment mode**: all (standalone, single-process, multi-process)
- **Build mode**: debug (`RELEASE=no`)
- **Machine**: microvm (also observed on hyperlight)
- **Failing test**: `dlfcn-pie-c.elf`

## Symptoms

- Test produces no output and the VM shuts down abruptly.
- Failure rate: ~2–4% in debug builds; not observed in release builds.
- Crash register state (100% consistent):
  - `CR2=0x007FFFFC` — write to unmapped page just below `KPOOL_BASE` (`0x800000 - 4`).
  - `ESI=0x008001A0` — reading from the page directory (should be a heap address).
  - `EDI=0x007FFFFC` — destination of the backwards `rep movsl`.
  - `ECX=0x0080DAD5` — garbage iteration count (should be `0x20`).
  - `fault_eip=0x0019c2c8` — `rep movsl` inside `Box::new_in<Node<RunnableProcess>>`.
  - `ESP=0x00812058` — only ~4 KiB used (NOT a deep stack overflow).

### Missing `cld` at Exception Entry

The x86 System V ABI requires the direction flag (DF, bit 10 of EFLAGS) to be **0** at every
function call boundary. All compiler-generated code — including Rust's `core::ptr::write`,
`memcpy`, and `Box::new` — assumes DF=0 and uses `rep movsl` / `rep movsb` for bulk copies.

The kernel exception handler saves EFLAGS in the hardware exception frame (preserving DF for
restoration on `iret`), but **never clears DF** before calling Rust code. If user-mode code
has executed `std` (setting DF=1) before the fault, every `rep` instruction in the handler
path operates in reverse.

### The Cascade

1. **User-space fault**: `dlfcn-pie-c.elf` intermittently page-faults with DF=1 in EFLAGS.
[...]

PR #1649, opened 2026-03-15
[kernel] Replace panic with graceful shutdown on overflow

ppenna:
Replace panic with error+shutdown in check_running_stack_guard(). Remove guard check from do_exit() and do_exit_thread(). Part of #1641.


ppenna:
Superseded by #1640, which includes the manager hardening changes (panic→shutdown, guard check removal) alongside the root cause fix (missing cld in exception entry).

PR #1648, opened 2026-03-15
[uservm] Log registers on triple fault

ppenna:
Dump ESP, EIP, CR2, and CR3 when the guest triggers a VcpuExit::Shutdown (triple fault). These values are essential for post-mortem diagnosis of page-directory corruption and stack overflow scenarios.

Part of #1641.

PR #1650, opened 2026-03-15
[kernel] Add exception recursion guard

ppenna:
Add _in_exception flag checked at _do_excp entry. On re-entry, halt via port I/O instead of cascading to triple fault. Part of #1641.


ppenna:
Superseded by #1640 and #1648. The BSS-based _in_exception flag approach was disproven during investigation: with corrupted page tables, the BSS variable itself (at 0x3D9080) becomes unmapped and causes the guard to fault. The ESP guard (register-only, zero memory accesses) in #1640 and the kernel-mode fault check in #1648 replace this approach.

PR #1651, opened 2026-03-15
[kernel] Increase kernel stack to 256 KB

ppenna:
Raise kstack_size from 64 KB to 256 KB. Debug-mode frames exceed 64 KB when timer interrupts fire during deep paths. Part of #1641.


ppenna:
No longer needed. Root cause of #1641 is a missing cld instruction in the exception handler entry, not insufficient stack size. The direction flag bug causes rep movsl to copy backwards through kpool (74 KiB), which was misdiagnosed as deep stack overflow. With the cld fix in #1640, the 64 KB kernel stack is sufficient.

PR #1666, opened 2026-03-16
[slab] Don't waste index bits on index

jaylorch:
This pull request refactors the slab allocator implementation to improve correctness and simplify block indexing logic. The main changes involve removing the explicit tracking of index blocks, updating how the number of index and data blocks is calculated, and refining allocation and deallocation logic to avoid off-by-one errors and ensure proper bounds checking.

### DIFF
--- a/src/libs/slab/src/lib.rs
+++ b/src/libs/slab/src/lib.rs
@@ -113,7 +113,6 @@ impl Slab {
 
         // Compute layout of the slab allocator.
         let total_num_blocks: usize = len / block_size;
-        // info!("total number of blocks: {:?}", total_num_blocks);
 
         // The number of index blocks (`num_index_blocks`) we need is
         //  `ceil(total_num_blocks / (block_size * u8::BITS + 1))`
@@ -133,7 +132,6 @@ impl Slab {
             } else {
                 1
             };
-        // info!("number of index blocks: {:?}", num_index_blocks);
         if num_index_blocks >= total_num_blocks {
             return Err(Error::new(ErrorCode::InvalidArgument, "insufficient blocks for index"));
         }
@@ -141,14 +139,12 @@ impl Slab {
         let data_addr: *mut u8 = addr.add(num_index_blocks * block_size);
 
         let num_data_blocks: usize = total_num_blocks - num_index_blocks;
-        // info!("number of data blocks: {:?}", num_data_blocks);
         let index_len = (num_data_blocks / U8_BITS)
             + if num_data_blocks.is_multiple_of(U8_BITS) {
                 0
             } else {
                 1
             };
-        // info!("length of index in bytes: {:?}", index_len);
 
         // Instantiate index.
         let storage: RawArray<u8> = RawArray::from_raw_parts(addr, index_len)?;
--- a/src/libs/slab/src/lib.rs
+++ b/src/libs/slab/src/lib.rs
@@ -125,8 +125,8 @@ impl Slab {
         // `num_index_blocks` blocks contain. The right-hand side of this inequality
         // is the number of blocks that aren't index blocks. So, a bitmap occupying
         // `num_index_blocks` blocks can address all the blocks outside of that bitmap.
-        let u8_bits: usize = u8::BITS as usize;
-        let divisor: usize = block_size * u8_bits + 1;
+        const U8_BITS: usize = u8::BITS as usize;
+        let divisor: usize = block_size * U8_BITS + 1;
         let num_index_blocks: usize = (total_num_blocks / divisor)
             + if total_num_blocks.is_multiple_of(divisor) {
                 0
@@ -142,8 +142,8 @@ impl Slab {
 
         let num_data_blocks: usize = total_num_blocks - num_index_blocks;
         // info!("number of data blocks: {:?}", num_data_blocks);
-        let index_len = (num_data_blocks / u8_bits)
-            + if num_data_blocks.is_multiple_of(u8_bits) {
+        let index_len = (num_data_blocks / U8_BITS)
+            + if num_data_blocks.is_multiple_of(U8_BITS) {
                 0
             } else {
                 1
@@ -164,7 +164,7 @@ impl Slab {
         // them "in use" and thereby prevent them from being
         // allocated. Note that there are at most 7 such bits we need
         // to set.
-        for i in num_data_blocks..(index_len * u8_bits) {
+        for i in num_data_blocks..(index_len * U8_BITS) {
             index.set(i)?;
         }
 
--- a/src/libs/slab/src/lib.rs
+++ b/src/libs/slab/src/lib.rs
@@ -101,11 +101,6 @@ impl Slab {
             return Err(Error::new(ErrorCode::InvalidArgument, "invalid block size"));
         }
 
-        // Check if the `block_size` is a power of two.
-        if block_size & (block_size - 1) != 0 {
-            return Err(Error::new(ErrorCode::InvalidArgument, "block size is not a power of two"));
-        }
-
         // Check if `addr` is aligned to `block_size`.
         if !(addr as usize).is_multiple_of(block_size) {
             return Err(Error::new(ErrorCode::InvalidArgument, "unaligned start address"));
--- a/src/libs/slab/src/lib.rs
+++ b/src/libs/slab/src/lib.rs
@@ -98,7 +98,7 @@ impl Slab {
             return Err(Error::new(ErrorCode::InvalidArgument, "wrapping memory region"));
         }
 
-        // Check if block size is valid.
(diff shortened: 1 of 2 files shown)


src/libs/slab/src/lib.rs
```diff
@@ -108,37 +106,49 @@ impl Slab {
...
+        // for the following reason. This condition implies:
+        //  `num_index_blocks * (block_size * u8::BITS + 1) >= total_num_blocks`
+        // This, in turn, implies that:
+        //  `num_index_blocks * block_size * u8::BITS  >= total_num_blocks - num_index_blocks`
+        // The left-hand side of this inequality is the number of bits that
+        // `num_index_blocks` blocks contain. The right-hand side of this inequality
+        // is the number of blocks that aren't index blocks. So, a bitmap occupying
+        // `num_index_blocks` blocks can address all the blocks outside of that bitmap.
+        let u8_bits: usize = u8::BITS as usize;
+        let divisor: usize = block_size * u8_bits + 1;
```
ppenna: Can these be `const` ?

src/libs/slab/src/lib.rs
```diff
@@ -108,37 +106,49 @@ impl Slab {
...
-        // info!("number of data blocks: {:?}", num_data_blocks);
+
         let data_addr: *mut u8 = addr.add(num_index_blocks * block_size);
 
-        // Check if `data_addr` is aligned to `block_size`.
-        if !(data_addr as usize).is_multiple_of(block_size) {
-            return Err(Error::new(ErrorCode::InvalidArgument, "unaligned data address"));
-        }
+        let num_data_blocks: usize = total_num_blocks - num_index_blocks;
+        // info!("number of data blocks: {:?}", num_data_blocks);
```
ppenna: This is legacy dead code and can be removed.

src/libs/slab/src/lib.rs
```diff
@@ -108,37 +106,49 @@ impl Slab {
...
+        let index_len = (num_data_blocks / u8_bits)
+            + if num_data_blocks.is_multiple_of(u8_bits) {
+                0
+            } else {
+                1
+            };
+        // info!("length of index in bytes: {:?}", index_len);
```
ppenna: This is legacy dead code and can be removed.

src/libs/slab/src/lib.rs
```diff
@@ -108,37 +106,49 @@ impl Slab {
...
+
         let data_addr: *mut u8 = addr.add(num_index_blocks * block_size);
 
-        // Check if `data_addr` is aligned to `block_size`.
-        if !(data_addr as usize).is_multiple_of(block_size) {
-            return Err(Error::new(ErrorCode::InvalidArgument, "unaligned data address"));
-        }
+        let num_data_blocks: usize = total_num_blocks - num_index_blocks;
+        // info!("number of data blocks: {:?}", num_data_blocks);
+        let index_len = (num_data_blocks / u8_bits)
```
ppenna: Missing type annotation.

src/libs/slab/src/lib.rs
```diff
@@ -108,37 +106,49 @@ impl Slab {
...
+        // `num_index_blocks` blocks can address all the blocks outside of that bitmap.
+        let u8_bits: usize = u8::BITS as usize;
+        let divisor: usize = block_size * u8_bits + 1;
+        let num_index_blocks: usize = (total_num_blocks / divisor)
+            + if total_num_blocks.is_multiple_of(divisor) {
                 0
             } else {
                 1
             };
         // info!("number of index blocks: {:?}", num_index_blocks);
```
ppenna: We are now removing the restriction that total blocks don't have to be a multiple of 8.

ppenna (commented):
**Minor issues found during review:**

---

**1. Stale test comment (test.rs, line 62–63)**

The comment in `test_slab_creation_minimal_valid_blocks` still says _"the guard against \`num_index_blocks > total_num_blocks\`"_ but the guard is now `>=`. Should be updated to match.

---

**2. Missing test for non-multiple-of-8 data blocks**

The PR removes the restriction that total blocks must be a multiple of 8, but no test exercises the trailing-bit-marking path (where `num_data_blocks % 8 != 0`). Consider adding a test, e.g.:

```rust
// 11 blocks of size 1: total=11, divisor=9, num_index=2, data=9
// → index_len = ceil(9/8) = 2, trailing bits = 16 - 9 = 7
```

This would confirm correctness of the trailing-bit initialization logic.

---

**3. `test_slab_creation` passes `memory.len()` (element count) as byte length**

Pre-existing issue not introduced by this PR, but worth noting: `Vec<u32>::len()` returns 1024 (element count), not 4096 (byte count). The test happens to work because `from_raw_parts` divides `len / block_size = 1024 / 4 = 256`, but conceptually `len` is documented as _"length of the memory region in bytes"_. `test_allocate_out_of_bounds` has the same issue.

src/libs/slab/src/lib.rs
jaylorch: Yes, that's one of the nice side effects of this update.

src/libs/slab/src/lib.rs
jaylorch: The `u8_bits` one can, indeed! Done.

src/libs/slab/src/lib.rs
jaylorch: Removed!

src/libs/slab/src/lib.rs
jaylorch: Removed!

src/libs/slab/src/lib.rs
jaylorch: Added!

src/libs/slab/src/lib.rs
```diff
@@ -108,37 +106,45 @@ impl Slab {
...
         if !(addr as usize).is_multiple_of(block_size) {
             return Err(Error::new(ErrorCode::InvalidArgument, "unaligned start address"));
         }
 
         // Compute layout of the slab allocator.
         let total_num_blocks: usize = len / block_size;
-        // info!("total number of blocks: {:?}", total_num_blocks);
-        if !total_num_blocks.is_multiple_of(u8::BITS as usize) {
-            return Err(Error::new(ErrorCode::InvalidArgument, "invalid number of blocks"));
-        }
```
ppenna, line 121: Are we intentionally dropping constraint of having this be a multiple of 8? why?

jaylorch: It doesn't help for the region to be a multiple of 8 blocks anymore, because the index doesn't cover the entire region. Even if the region is a multiple of 8 blocks, that doesn't mean the part of the region the index covers (which excludes the index) will be a multiple of 8 blocks.

ppenna: My concern is not to introduce dead bits in the map.

jaylorch: It would be better not to have dead bits in the map, but, alas, I don't think there's a way. There are either going to be dead bits at the beginning (like the current approach) or the end (like this new approach). At least this new approach (a) has the bits at the end, where they won't have to be scanned through on every allocation, and (b) has a bound on the number of dead bits (it's at most seven).

jaylorch:
> **Minor issues found during review:**
> 
> **1. Stale test comment (test.rs, line 62–63)**
> 
> The comment in `test_slab_creation_minimal_valid_blocks` still says _"the guard against `num_index_blocks > total_num_blocks`"_ but the guard is now `>=`. Should be updated to match.

Fixed!

> **2. Missing test for non-multiple-of-8 data blocks**
> 
> The PR removes the restriction that total blocks must be a multiple of 8, but no test exercises the trailing-bit-marking path (where `num_data_blocks % 8 != 0`). Consider adding a test, e.g.:
> 
> ```rust
> // 11 blocks of size 1: total=11, divisor=9, num_index=2, data=9
> // → index_len = ceil(9/8) = 2, trailing bits = 16 - 9 = 7
> ```
> 
> This would confirm correctness of the trailing-bit initialization logic.

OK, I've added more code in `test_slab_creation_minimal_valid_blocks` to test for this.

> **3. `test_slab_creation` passes `memory.len()` (element count) as byte length**
> 
> Pre-existing issue not introduced by this PR, but worth noting: `Vec<u32>::len()` returns 1024 (element count), not 4096 (byte count). The test happens to work because `from_raw_parts` divides `len / block_size = 1024 / 4 = 256`, but conceptually `len` is documented as _"length of the memory region in bytes"_. `test_allocate_out_of_bounds` has the same issue.

Good find! I updated all the tests to use `Vec<u8>` instead of `Vec<u32>` to make sure that using `memory.len()` properly uses the length in bytes.

ppenna (dismissed):
LGTM. Please rebase with latest `origin/dev`

PR #1692, opened 2026-03-18
[libs] Use stored field in `index()` check for bitmap

tangruize:
The `index()` function recomputed the total bit count as `bits.len() * u8::BITS` for its bounds check. Replace this with the pre-stored `number_of_bits` field, which already holds the same value and avoids a redundant multiplication.


ppenna (approved):
LGTM

PR #1693, opened 2026-03-18
[kernel] Fix `Condvar::notify_process()` doc to match code

tangruize:
The documentation for `Condvar::notify_process()` claimed it "wakes up all threads of a process" but the implementation only wakes the first matching thread via `position().remove()`. This doc fix accurately describes the existing single-thread-wake behavior.


ppenna (approved):
LGTM

PR #1729, opened 2026-03-19
[build] Use upstream hyperlight v0.13.1

danbugs:
## Summary

- Switch hyperlight dependencies from nanvix fork to upstream crates.io v0.13.1
- Adapt kernel platform layer to upstream API (PV timer, memory layout, scratch I/O, RAMFS file mappings, credit management via scratch memory)
- Adapt uservm to upstream API (GuestCounter, StderrRedirect, map_file_cow for RAMFS, simplified credit management)
- Adapt bulk output function to read inline payload from combined buffer
- Fix VFS to reject read on directory FDs and update stat epoch


src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -180,34 +180,44 @@ pub unsafe fn puts(message: &str) {
...
+            fn __phys_memcpy(dst: *mut u8, src: *const u8, size: usize);
+        }
+        __phys_memcpy(payload.as_mut_ptr(), data_gpa as *const u8, data_len);
+
+        // Send header + payload together through VmbusWrite.
+        let mut combined: alloc::vec::Vec<u8> =
+            alloc::vec::Vec::with_capacity(::sys::ipc::DataChunkHeader::SIZE + data_len);
+        combined.extend_from_slice(&header_bytes);
+        combined.extend_from_slice(&payload);
+        let _ = ProcessEnvironmentBlock::vmbus_bulk_write(&combined);
```
ppenna: Can we open an issue to keep track of this workaround?

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -398,14 +407,23 @@ fn register_pit(ioports: &mut IoPortAllocator) -> Result<Pit, Error> {
     ioports.register_read_write(::arch::cpu::pit::PIT_CTRL)?;
     ioports.register_read_write(::arch::cpu::pit::PIT_DATA)?;
 
+    // Configure the PV timer: writing the period (in microseconds) to the
+    // PvTimerConfig port tells the hypervisor to inject periodic timer interrupts at that rate.
+    // This must match OutBAction::PvTimerConfig (107) from hyperlight-common.
+    const PV_TIMER_PORT: u16 = 107;
+    let period_us: u32 = 1_000_000 / ::config::kernel::TIMER_FREQ;
```
ppenna: Please create a constant in `config/constants.rs` and use it here.

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -398,14 +407,23 @@ fn register_pit(ioports: &mut IoPortAllocator) -> Result<Pit, Error> {
     ioports.register_read_write(::arch::cpu::pit::PIT_CTRL)?;
     ioports.register_read_write(::arch::cpu::pit::PIT_DATA)?;
 
+    // Configure the PV timer: writing the period (in microseconds) to the
+    // PvTimerConfig port tells the hypervisor to inject periodic timer interrupts at that rate.
+    // This must match OutBAction::PvTimerConfig (107) from hyperlight-common.
+    const PV_TIMER_PORT: u16 = 107;
```
ppenna: Move it `config` crate.

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -552,6 +553,75 @@ pub fn init(
...
+    let peb_ptr: *const HyperlightPEB = peb_base as *const HyperlightPEB;
+    let input_gva = unsafe { (*peb_ptr).input_stack.ptr } as usize;
+
+    if input_gva >= config::kernel::MEMORY_SIZE {
+        // Extend the scratch I/O region to cover all the way up to (but not including)
+        // the last page of the address space. This includes the I/O buffers AND the
+        // bookkeeping area (scratch size, allocator, exception stack, guest counter).
+        // We stop one page short of MAX_GVA because the last page (frame 0xfffff)
+        // exceeds FrameNumber::MAX and cannot be booked by the frame allocator.
+        use ::hyperlight_common::layout::MAX_GVA;
```
ppenna: Place imports at file-level scope.

src/libs/vfs/src/fd.rs
```diff
@@ -445,6 +445,10 @@ pub fn vfs_read(fd: c_int, buf: &mut [u8]) -> Result<c_size_t, Fat32Error> {
     let mut slot: spin::MutexGuard<'_, Option<VfsEntry>> = FD_TABLE.lock(fd)?;
     let entry: &mut VfsEntry = slot.as_mut().ok_or(Fat32Error::InvalidFd)?;
 
+    if entry.handle.is_dir() {
+        return Err(Fat32Error::NotSupported);
+    }
```
ppenna, line 450: Whys is this change required?

src/libs/vfs/src/fd.rs
```diff
@@ -527,8 +531,8 @@ pub fn vfs_lseek(fd: c_int, offset: off_t, whence: c_int) -> Result<off_t, Fat32
 /// - Timestamps set to a fixed epoch value (FAT has no sub-second precision).
 /// - Permissions: owner read+write for files, owner rwx for directories.
 fn populate_stat_fields(buf: &mut ::sysapi::sys_stat::stat, size: u64, is_dir: bool) {
-    // Fixed epoch timestamp: 2020-01-01T00:00:00Z (1577836800).
-    const FIXED_EPOCH: i64 = 1_577_836_800;
+    // Fixed epoch timestamp: 2024-01-01T00:00:00Z (1704067200).
+    const FIXED_EPOCH: i64 = 1_704_067_200;
```
ppenna, line 535: Why is this change required?

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -148,72 +144,22 @@ impl Drop for StderrRedirect {
...
-        })?;
-        self.manager
-            .get_shared_mem_mut()
-            .copy_to_slice(data, offset)
-            .map_err(|e| {
-                let reason: String = format!("read_bytes failed at {addr:#x}: {e}");
-                error!("{reason}");
-                anyhow::anyhow!(reason)
-            })
+    /// TODO
```
ppenna: What is todo? Please open a github issue and link it here.

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -357,48 +231,10 @@ impl Vmm {
...
-                        let reason: &str = "not enough memory";
-                        error!(
-                            "new(): {reason} ({required_memory} bytes required, {memory_size} \
-                             bytes total)"
-                        );
-                        return Err(anyhow::anyhow!(reason));
-                    }
-
-                    // Round up to page boundary for KVM compatibility.
-                    // FIXME (#1307): fix calculation for required_memory.
```
ppenna, line 381: Did we fix this?

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -538,32 +407,71 @@ impl Vmm {
...
-                        Ok(code as u16)
-                    },
-                    _ => {
-                        error!("run(): vmm aborted (error={error:?})");
-                        Ok(ErrorCode::ConnectionReset.into())
-                    },
+                            String::new()
+                        };
+                        return Some((code, message));
+                    }
```
ppenna: This level of nesting is hard to read, can we breakdown into functions? When people see code like that, they just skip. Also, what are the hardcode values `6` and `10` about?

src/utils/nanvixd/Cargo.toml
```diff
@@ -25,8 +25,8 @@ default = ["microvm", "multi-process"]
 # into Nanvix Daemon (nanvixd).
 single-process = ["nanvix/single-process"]
 multi-process = ["nanvix/multi-process"]
-standalone = ["gdb", "nanvix/standalone"]
-gdb = ["microvm", "nanvix/gdb"]
+standalone = ["nanvix/standalone"]
+gdb = ["nanvix/gdb"]
```
ppenna, line 29: Please extract on different PR.

ppenna (changes requested):
Left some minor comments.

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -152,14 +152,18 @@ pub struct VirtualMemory {
 //==================================================================================================
 
 impl VirtualMemory {
-    /// TODO
+    /// Writes a sequence of bytes into guest memory at the given address.
+    ///
+    /// TODO (#1731): implement using upstream hyperlight host shared-memory API.
     pub fn write_bytes(&mut self, _addr: u64, _data: &[u8]) -> ::anyhow::Result<()> {
-        unimplemented!();
+        Err(anyhow::anyhow!("write_bytes not implemented for hyperlight VMM"))
```
ppenna, line 167: Please log an error before failing.

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -152,14 +152,18 @@ pub struct VirtualMemory {
...
     }
 
-    /// TODO
+    /// Reads a sequence of bytes from guest memory at the given address.
+    ///
+    /// TODO (#1731): implement using upstream hyperlight host shared-memory API.
     pub fn read_bytes(&mut self, _addr: u64, _data: &mut [u8]) -> ::anyhow::Result<()> {
-        unimplemented!();
+        Err(anyhow::anyhow!("read_bytes not implemented for hyperlight VMM"))
```
ppenna, line 175: Please log an error before failing.

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -448,30 +452,31 @@ impl Vmm {
...
+                let msg_str = &rest[pos + MESSAGE_PREFIX.len()..];
+                msg_str
+                    .find('"')
+                    .map(|end| msg_str[..end].to_string())
+                    .unwrap_or_default()
+            })
+            .unwrap_or_default();
+
+        Some((code, message))
     }
```
ppenna, line 512: Can he have a unit test for this?

src/kernel/src/hal/platform/hyperlight/mod.rs
danbugs: Done. Opened #1730 to track the `__phys_memcpy` workaround. Also added a `FIXME (#1730)` comment in the code. Additionally, per Copilot's suggestion, optimized the bulk path to use a single allocation instead of two (copy header into the first bytes, then `__phys_memcpy` payload directly into the tail).

ppenna (dismissed):
Left some final comments please address them, rebase with latest `origin/dev` and make PR as ready.

src/kernel/src/hal/platform/hyperlight/mod.rs
danbugs: Done. Added `TIMER_PERIOD_US` to the hyperlight config template alongside `PV_TIMER_PORT`. It mirrors the existing `microvm::TIMER_PERIOD_US` pattern. The kernel code now uses `::config::hyperlight::TIMER_PERIOD_US` and `::config::hyperlight::PV_TIMER_PORT`.

src/kernel/src/hal/platform/hyperlight/mod.rs
danbugs: Done (same commit as above). Both `PV_TIMER_PORT` and `TIMER_PERIOD_US` now live in the hyperlight config template.

src/kernel/src/hal/platform/hyperlight/mod.rs
danbugs: Done. Moved `MAX_GVA`, `FileMappingInfo`, `RAMFS_REGION_NAME`, and `RAMFS_MMIO_TAG` from block-scoped imports to file-level imports.

src/uservm/src/vmm/hyperlight/mod.rs
danbugs: Done. Opened #1731 to track implementing `read_bytes`/`write_bytes` using the upstream hyperlight shared-memory API. The `unimplemented!()` calls have been replaced with `Err(anyhow!(...))` stubs to satisfy `clippy::unimplemented`, and the doc comments now reference the tracking issue.

src/uservm/src/vmm/hyperlight/mod.rs
danbugs: Yes, this is effectively fixed by the migration to upstream hyperlight v0.13.1. The old code had a complex `required_memory` calculation that tried to predict the exact guest memory layout. The new code simply sets `heap_size = memory_size` and lets hyperlight manage the internal allocation. Since hyperlight now handles the layout internally, the FIXME is no longer applicable.

src/uservm/src/vmm/hyperlight/mod.rs
danbugs: Done. Extracted the Debug-string parsing into a separate `parse_guest_aborted_from_debug()` method. The magic numbers `6` and `10` were the lengths of the string prefixes `"code: "` (6 chars) and `"message: \""` (10 chars) — replaced them with named `const` strings (`CODE_PREFIX`, `MESSAGE_PREFIX`) and use `.len()` instead. The nesting is also reduced since the helper uses early-return via `?` on `Option`.

src/utils/nanvixd/Cargo.toml
danbugs: Done. Reverted the `standalone`/`gdb` feature decoupling — it's back to the original `standalone = ["gdb", "nanvix/standalone"]` and `gdb = ["microvm", "nanvix/gdb"]`. Will open a separate PR for that change.

src/uservm/src/lib.rs
```diff
@@ -954,30 +955,13 @@ pub fn bulk_output_fn(
...
-                hyperlight_host::HyperlightError::AnyhowError(anyhow::Error::msg(reason))
-            })?;
+        // Extract the inline payload that follows the header.
+        let mut payload_data: Vec<u8> = data[DataChunkHeader::SIZE..].to_vec();
 
         // Label: uservm::lib::vm_output::send()
         profiler::timestamp_message!(&mut payload_data, 0);
 
+        let data_len: usize = payload_data.len();
         let bulk: DataChunk = DataChunk::new(header, payload_data);
```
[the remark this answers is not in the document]
danbugs: Done. Added validation that compares `actual_len` against `header.data_len()`. Returns an error if the payload is truncated, and slices to exactly `expected_len` bytes to avoid forwarding extra data.

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -552,6 +553,75 @@ pub fn init(
...
+                        TruncatedMemoryRegion::from_memory_region(MemoryRegion::new(
+                            RAMFS_REGION_NAME,
+                            VirtualAddress::from_raw_value(ramfs_base),
+                            ramfs_size,
+                            MemoryRegionType::Mmio,
+                            AccessPermission::RDWR,
+                        )?)?;
+                    ioaddresses.register(RAMFS_MMIO_TAG, ramfs_region.clone())?;
+                    mmio_regions.push_back(ramfs_region);
+                }
```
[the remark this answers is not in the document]
danbugs: Done. The loop now reads the entry's `label` field and only registers entries whose label is `"ramfs"`, skipping any other file mappings.

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -180,34 +180,44 @@ pub unsafe fn puts(message: &str) {
...
+        extern "C" {
+            fn __phys_memcpy(dst: *mut u8, src: *const u8, size: usize);
+        }
+        __phys_memcpy(payload.as_mut_ptr(), data_gpa as *const u8, data_len);
+
+        // Send header + payload together through VmbusWrite.
+        let mut combined: alloc::vec::Vec<u8> =
+            alloc::vec::Vec::with_capacity(::sys::ipc::DataChunkHeader::SIZE + data_len);
+        combined.extend_from_slice(&header_bytes);
+        combined.extend_from_slice(&payload);
```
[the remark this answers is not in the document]
danbugs: Done. Replaced the two-allocation pattern with a single `combined` buffer. Header is copied into the first bytes, then `__phys_memcpy` writes the payload directly into the tail.

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -538,32 +407,71 @@ impl Vmm {
...
+
+        // For Initialize path, parse the Debug representation.
+        // Format: ...GuestAborted { code: N, message: "..." }...
+        let debug_str = format!("{error:?}");
+        if let Some(start) = debug_str.find("GuestAborted") {
+            let rest = &debug_str[start..];
+            // Extract code value after "code: "
+            if let Some(code_start) = rest.find("code: ") {
+                let code_str = &rest[code_start + 6..];
+                if let Some(code_end) = code_str.find([',', ' ', '}']) {
```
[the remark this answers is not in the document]
danbugs: Actually, `str::find([char; N])` compiles fine in stable Rust (the `Pattern` trait is implemented for arrays of `char`). No change needed.

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -148,72 +144,22 @@ impl Drop for StderrRedirect {
...
-            .copy_to_slice(data, offset)
-            .map_err(|e| {
-                let reason: String = format!("read_bytes failed at {addr:#x}: {e}");
-                error!("{reason}");
-                anyhow::anyhow!(reason)
-            })
+    /// TODO
+    pub fn read_bytes(&mut self, _addr: u64, _data: &mut [u8]) -> ::anyhow::Result<()> {
+        unimplemented!();
     }
```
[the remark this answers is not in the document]
danbugs: Replaced `unimplemented!()` with `Err(anyhow!(...))` stubs. Tracked in #1731.

src/kernel/src/hal/platform/hyperlight/mod.rs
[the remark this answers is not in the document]
danbugs: Done. The loop now reads the entry's `label` field and only registers entries whose label is `"ramfs"`, skipping any other file mappings.

src/libs/vfs/src/fd.rs
```diff
@@ -445,6 +445,10 @@ pub fn vfs_read(fd: c_int, buf: &mut [u8]) -> Result<c_size_t, Fat32Error> {
     let mut slot: spin::MutexGuard<'_, Option<VfsEntry>> = FD_TABLE.lock(fd)?;
     let entry: &mut VfsEntry = slot.as_mut().ok_or(Fat32Error::InvalidFd)?;
 
+    if entry.handle.is_dir() {
+        return Err(Fat32Error::NotSupported);
```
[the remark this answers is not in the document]
danbugs: Kept as `NotSupported` since that's the consistent pattern used throughout the VFS for all unsupported operations on directory FDs (`write`, `seek`, `ftruncate`, `fallocate` all return `NotSupported`). Switching just `read` to `NotAFile` would be inconsistent. If we want strict POSIX `EISDIR` semantics, we should do it for all directory operations in a separate change.

src/libs/vfs/src/fd.rs
danbugs: Tested without it — `test_open_directory` in the VFS test suite crashes the kernel. Without this guard, `vfs_read()` is called on a directory file descriptor and falls through to the FAT32 read path, which can't handle directory handles and causes the kernel to abort. The guard returns `NotSupported` early, which is the same pattern used by `write()`, `seek()`, etc. on directory FDs throughout the VFS.

src/libs/vfs/src/fd.rs
danbugs: Also tested without it — `test_fstat_directory_and_timestamps` aborts the kernel when the epoch is 2020. The VFS test binary was compiled expecting timestamps matching `1704067200` (2024-01-01). The epoch value in `populate_stat_fields` must match what the test binary asserts against, so the update from 2020 to 2024 is required for the test suite to pass.

src/uservm/src/vmm/hyperlight/mod.rs
danbugs: Done

src/uservm/src/vmm/hyperlight/mod.rs
danbugs: Done

src/uservm/src/vmm/hyperlight/mod.rs
danbugs: Done

src/libs/vfs/src/fd.rs
ppenna: Is the test for `microvm` machine also failing?

src/libs/vfs/src/fd.rs
ppenna: Is the test for microvm machine also failing?

src/libs/vfs/src/fd.rs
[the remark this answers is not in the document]
ppenna: Is the test for microvm machine also failing?

test/test-l2.toml
```diff
@@ -61,18 +61,21 @@ executor = "http"
 program = "./bin/echo-c.elf"
 input = "hello world!"
 expected_output = "hello world!"
+runs_on = ["microvm", "qemu-pc"]
```
ppenna: `qemu-pc` is invalid

test/test-l2.toml
ppenna, line 1: So, are we just disabling all tests for hyperlight? If so, let's remove it for once the CI and save CI time. Also, for each test that we disable, please create an issue so we can track this work item.

.github/workflows/ci.yml
ppenna, line 1: If we are disabling most benchmarks, shouldn't we simply disable all benchmarks for hyperlight and thus save CI time? Also, for each benchmark we disable, please create a tracking issue.

ppenna (dismissed):

ppenna:
Converted this PR to draft, as work is still required and we cannot afford use the CI for change-debug-test.

test/test-l2.toml
danbugs: Yes, removed.

test/test-l2.toml
danbugs: No, all tests are passing. That was just part of debugging.

.github/workflows/ci.yml
danbugs: Yes, benchmarks were disabled in this PR.

src/libs/vfs/src/fd.rs
danbugs: Will create an issue to investigate.

src/libs/vfs/src/fd.rs
danbugs: Will create an issue to investigate.

src/libs/vfs/src/fd.rs
ppenna, line 450: Calling read on a directory is a posix-compliant operation. Why does it pass for the `microvm` machine type and not for `hyperlight`? Can you create an integration test for this? Anyways, this change must be submitted as another PR.

src/libs/vfs/src/fd.rs
ppenna, line 535: Why does it pass for the `microvm` machine type and not for `hyperlight`? Can you create an integration test for this? Anyways, this change must be submitted as another PR.

src/uservm/src/vmm/hyperlight/guest.rs
```diff
@@ -47,21 +39,9 @@ impl Guest {
...
-
-        let credits_offset: usize = manager.get_guest_credits_offset();
-        manager
-            .get_shared_mem_mut()
-            .write::<u64>(credits_offset, self.credits)?;
-
-        Ok(())
+        vmem.counter
+            .increment()
+            .map_err(|e| anyhow::anyhow!("{e:?}"))
```
ppenna: Please log errors before failing.

src/uservm/src/vmm/hyperlight/guest.rs
```diff
@@ -78,21 +58,9 @@ impl Guest {
...
-
-        let credits_offset: usize = manager.get_guest_credits_offset();
-        manager
-            .get_shared_mem_mut()
-            .write::<u64>(credits_offset, self.credits)?;
-
-        Ok(())
+        vmem.counter
+            .decrement()
+            .map_err(|e| anyhow::anyhow!("{e:?}"))
```
ppenna: Please log errors before failing.

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -420,27 +265,57 @@ impl Vmm {
...
+        config.set_heap_size(heap_size as u64);
 
         // Creates Hyperlight sandbox.
-        let mut sandbox: UninitializedSandbox = UninitializedSandbox::new(guest_env, Some(config))?;
-        let manager: SandboxMemoryManager<ExclusiveSharedMemory> = sandbox.mgr.clone();
-        let vmem: Arc<Mutex<VirtualMemory>> = Arc::new(Mutex::new(VirtualMemory {
-            manager: manager.clone(),
-        }));
+        let mut sandbox: UninitializedSandbox = UninitializedSandbox::new(guest_env, Some(config))
+            .map_err(|e| anyhow::anyhow!("{e:?}"))?;
```
ppenna: Please log errors before failing.

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -420,27 +265,57 @@ impl Vmm {
...
+        // shared memory slot. With nanvix-unstable, BASE_ADDRESS is 0x0 so the
+        // shared memory occupies GPA [0, shared_mem_size). The file mapping
+        // metadata is automatically written to the PEB during evolve(), so the
+        // guest kernel can discover and identity-map the RAMFS region during boot.
+        if let Some(ramfs_filename) = &args.ramfs_filename {
+            let ramfs_path: &Path = Path::new(ramfs_filename);
+            let ramfs_gpa: u64 = sandbox.shared_mem_size() as u64;
+            let ramfs_size: u64 = sandbox
+                .map_file_cow(ramfs_path, ramfs_gpa, Some("ramfs"))
+                .map_err(|e| anyhow::anyhow!("failed to map ramfs file: {e:?}"))?;
```
ppenna: Please log errors before failing.

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -420,27 +265,57 @@ impl Vmm {
...
+                ramfs_path, ramfs_size, ramfs_gpa
+            );
+        }
+
+        // Create a guest counter backed by a fixed offset in scratch memory.
+        // The counter holds its own Arc clones of the mapping handle and RwLock,
+        // so it remains valid across evolve() and for the sandbox's entire lifetime.
+        let counter: GuestCounter = sandbox
+            .guest_counter()
+            .map_err(|e| anyhow::anyhow!("{e:?}"))?;
```
ppenna: Please log errors before failing.

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -420,27 +265,57 @@ impl Vmm {
...
+        // The file is mapped copy-on-write at the first GPA after the sandbox's
+        // shared memory slot. With nanvix-unstable, BASE_ADDRESS is 0x0 so the
+        // shared memory occupies GPA [0, shared_mem_size). The file mapping
+        // metadata is automatically written to the PEB during evolve(), so the
+        // guest kernel can discover and identity-map the RAMFS region during boot.
+        if let Some(ramfs_filename) = &args.ramfs_filename {
+            let ramfs_path: &Path = Path::new(ramfs_filename);
+            let ramfs_gpa: u64 = sandbox.shared_mem_size() as u64;
+            let ramfs_size: u64 = sandbox
+                .map_file_cow(ramfs_path, ramfs_gpa, Some("ramfs"))
```
ppenna: Can we have a file-scope constant for `"ramfs"`?

src/uservm/src/lib.rs
```diff
@@ -762,6 +785,38 @@ pub fn build_input_fn(
...
+
+        let chunk_len: usize = buf.len().min(MAX_CHUNK);
+        let chunk: Vec<u8> = buf.drain(..chunk_len).collect();
+        trace!("VmbusBulkRead: returning {} bytes ({} remaining)", chunk_len, buf.len());
+        Ok(chunk)
+    };
+
+    Box::new(bulk_input)
+}
+
```
ppenna, line 820: The number of 512-byte buffer is twice as low as the number of 256-bytes buffer, which is in turn tice as low as the number of 128-bytes buffer.... Relying on having the biggest buffer is the most fragile decision.

Cargo.toml
```diff
@@ -199,7 +205,7 @@ vfs = { path = "src/libs/vfs", default-features = false }
 vstd = { version = "=0.0.0-2026-03-15-0111", default-features = false }
 vmm-sys-util = { version = "0.15.0", default-features = false }
 wasmi = { version = "1.0.9", default-features = false }
-windows = { version = "0.62", features = [
+windows = { version = "0.61", features = [
```
ppenna: Please revert this change.

src/kernel/src/mm/kheap.rs
```diff
@@ -143,6 +158,24 @@ impl Kheap {
...
+                heap_start_addr.add(8 * slab_size),
+                slab_size,
+                SlabSize::Slab2048 as usize,
+            )?,
+            #[cfg(feature = "hyperlight")]
+            slab_4096_bytes: Slab::from_raw_parts(
+                heap_start_addr.add(9 * slab_size),
+                slab_size,
+                SlabSize::Slab4096 as usize,
+            )?,
```
ppenna, line 180: The kernel is not designed for supporting large allocations (this is the reason why I dropped recently the 4096-entry size). Do you know what is causing such large allocators? In any case we need to create an issue and link it here.

src/libs/nvx/src/lib.rs
```diff
@@ -408,13 +408,32 @@ fn vfs_init_ramfs() {
...
+        // The VFS needs read-write access, so copy the FAT image into a heap buffer.
+        #[cfg(feature = "hyperlight")]
+        let (mount_ptr, free_mmio_early) = {
+            let base_ptr: *const u8 = info.base().into_raw_value() as *const u8;
+            let mut buf: alloc::vec::Vec<u8> = alloc::vec![0u8; total_size];
+            unsafe { core::ptr::copy_nonoverlapping(base_ptr, buf.as_mut_ptr(), total_size) };
+            let rw_ptr: *mut u8 = buf.as_mut_ptr();
+            core::mem::forget(buf);
             let _ = ::sys::kcall::mm::mmio_free(RAMFS_MMIO_TAG);
+            (rw_ptr, true)
```
ppenna, line 423: This will hurt performance significantly, especially for large file systems. Please create an issue and link it here.

ppenna (dismissed):

ppenna:
@danbugs move this back to draft so you can address comments.

src/libs/vfs/src/fd.rs
danbugs: Updated the issue tracking this work item (#1761), but, tldr, this fix is also needed in the microvm. Not sure why CI wasn't catching the problem. Will extract to a separate PR and perhaps add an integration test for this specifically.

src/libs/vfs/src/fd.rs
danbugs: Updated the issue tracking this work item (#1761), but, tldr, this fix is also needed in the microvm. Not sure why CI wasn't catching the problem. Will extract to a separate PR and perhaps add an integration test for this specifically.

src/uservm/src/vmm/hyperlight/guest.rs
danbugs: Added error!() logging before returning the error in both add_credit() and consume_credit().

src/uservm/src/vmm/hyperlight/guest.rs
danbugs: Added error!() logging before returning the error in both add_credit() and consume_credit().

src/uservm/src/vmm/hyperlight/mod.rs
danbugs: Added error!() logging before failing in UninitializedSandbox::new.

src/uservm/src/vmm/hyperlight/mod.rs
danbugs: Done

src/uservm/src/vmm/hyperlight/mod.rs
danbugs: Done

src/uservm/src/vmm/hyperlight/mod.rs
danbugs: Added const RAMFS_LABEL: &str = "ramfs" at file scope and updated the map_file_cow call to use it.

Cargo.toml
danbugs: Reverted windows crate from 0.61 back to 0.62.

src/uservm/src/lib.rs
danbugs: Created #1779 to track evaluating a smaller chunk size (targeting the 256-byte slab) at the cost of more VM exits. Added FIXME comment in code.

src/kernel/src/mm/kheap.rs
danbugs: Created #1780 to investigate chunking the output path (same pattern as the input path) so we can remove the extended slab tiers entirely. Added FIXME comment in code.

src/libs/nvx/src/lib.rs
danbugs: Linked existing issue #1760 with a FIXME comment.

PR #1763, opened 2026-03-21
[build] Fix Stale-Artifact Bugs

ppenna:
## Summary

Fixes two stale-artifact bugs that affect the Windows build-run-debug cycle
with `z.ps1`.

### Pre-clean guest artifacts before Docker build

Docker's `--output type=local,dest=.` is additive: it writes new files but
never deletes files that no longer exist in the build output. Without cleanup,
stale guest binaries (e.g., a removed `.elf`) persist across builds in `bin/`
and `lib/`.

`Build-Guest` now removes guest-produced files (`*.elf`, `*.wasm`, `*.a`,
`*.so`, `*.img`) and ephemeral sysroot directories before invoking
`Invoke-DockerBuild`. Host-built `.exe` files are preserved.

### Sync stale host binaries before run

`cargo build` writes to `target/{debug,release}/` but `z.ps1 run` launches
from `bin/`. When a developer builds with `cargo build` directly (e.g., during
an IDE debug iteration) and then runs `z.ps1 run`, the executed binary is
stale.

`Invoke-Run` now calls `Sync-HostBinaries`, which compares timestamps of
`nanvixd.exe` and `uservm.exe` between `target/` and `bin/`. If `target/`
holds a newer copy, it is automatically copied to `bin/` with a warning.

## Testing

- Reproduced Bug 1: confirmed 28 s timestamp gap between `target/debug/uservm.exe`
  and `bin/uservm.exe` after a direct `cargo build`.
- Verified `Sync-HostBinaries` detects and resolves the gap.
- Verified pre-clean removes stale `.elf`/`.wasm` files.
- Syntax-validated with `[System.Management.Automation.PSParser]`.

### DIFF
--- a/z.ps1
+++ b/z.ps1
@@ -425,7 +425,33 @@ function Build-UserVm {
 function Build-Guest {
     param([bool]$IsRelease, [bool]$UseMinimal = $true, [string[]]$ExtraMakeParams = @())
 
+    # Prevent cleanup errors from aborting the build (consistent with other functions).
+    $ErrorActionPreference = 'Continue'
+
     Write-Info "Building guest components (Docker)..."
+
+    # Remove guest artifacts from previous Docker exports. Docker's
+    # --output type=local is additive: it writes new files but never deletes
+    # files that no longer exist in the output. Without this cleanup, stale
+    # guest binaries (e.g., a removed .elf) persist across builds.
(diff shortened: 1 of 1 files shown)


ppenna:
Supersede by #1770

PR #1798, opened 2026-03-24
[slab] Prevent overflow

jaylorch:
The code for `Slab::from_raw_parts` has the following line, which might overflow:
```
        let divisor: usize = block_size * U8_BITS + 1;
```
As I started verifying it, Verus pointed out that this might overflow. So I'm submitting this PR to ensure it can't.


ppenna:
Moved to draft to coordinate merges.

ppenna (approved):

PR #1828, opened 2026-03-26
Store versioned releases and add retention policy

ada-x64:
- [x] Fix 1: Verify existing tag points to same commit instead of force-updating (lines 121-128) — fetches existing ref, fails if SHA differs, reuses if same
- [x] Fix 2: Only set `make_latest: "true"` when commit is current dev HEAD (lines 138-148) — computes `isDevHead` once at script start
- [x] Fix 3: Gate `latest` git tag update on commit being current dev HEAD (lines 279-301) — skips with informative message on older re-runs
- [x] Additional: added try-catch around `getRef` call during tag verification for clearer error context
- [x] Code review and security scan

---

⚡ Quickly spin up Copilot coding agent tasks from anywhere on your macOS or Windows machine with [Raycast](https://gh.io/cca-raycast-docs).

### DIFF
--- a/.github/actions/publish-release/action.yml
+++ b/.github/actions/publish-release/action.yml
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 name: "Publish Release"
-description: "Publish the latest release to GitHub and trigger downstream dispatches"
+description: "Publish a versioned release to GitHub and trigger downstream dispatches"
 
 inputs:
   github-token:
@@ -33,47 +33,23 @@ runs:
         pattern: release-*
         path: release-artifacts
 
-    - name: "Reset Latest Release"
+    - name: "Clean Up Stale Releases"
       uses: actions/github-script@v8
       with:
         github-token: ${{ inputs.github-token }}
         script: |
-          const tag = 'latest';
           const owner = context.repo.owner;
           const repo = context.repo.repo;
 
-          async function deleteReleaseByTag() {
-            try {
-              const { data } = await github.rest.repos.getReleaseByTag({ owner, repo, tag });
-              core.info(`Deleting release ${data.id} (${tag}).`);
-              await github.rest.repos.deleteRelease({ owner, repo, release_id: data.id });
-            } catch (error) {
-              if (error.status === 404) {
-                core.info('No existing latest release to delete.');
-                return;
-              }
-              throw error;
-            }
-          }
-
-          // Clean up orphaned draft releases left behind by prior runs that
-          // failed after creating a draft but before publishing it.
-          // getReleaseByTag() only finds published releases, so draft orphans
-          // are invisible to the deleteReleaseByTag() cleanup above.
-          // Also cleans up "untagged-*" draft releases created when the GitHub
-          // API fails to associate a tag during createRelease (e.g., due to a
-          // tag deletion/creation race condition).
+          // Clean up orphaned draft releases left behind by prior runs.
+          // All intentional releases are created with draft: false, so any
+          // draft release is stale by definition.
           async function deleteStaleDraftReleases() {
             const releases = await github.paginate(
               github.rest.repos.listReleases,
               { owner, repo, per_page: 100 }
             );
-            const stale = releases.filter(r =>
-              r.draft && (
-                r.tag_name === tag ||
-                r.tag_name.startsWith('untagged-')
-              )
-            );
+            const stale = releases.filter(r => r.draft);
             for (const rel of stale) {
               core.info(`Deleting stale draft release ${rel.id} (tag="${rel.tag_name}", name="${rel.name}").`);
               await github.rest.repos.deleteRelease({ owner, repo, release_id: rel.id });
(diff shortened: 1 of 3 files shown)


.github/actions/publish-release/action.yml
```diff
@@ -112,93 +73,104 @@ runs:
...
+            } catch (error) {
+              if (error.status === 422) {
+                core.info(`Tag ${tag} already exists; updating to ${commitSha}.`);
+                await github.rest.git.updateRef({
+                  owner,
+                  repo,
+                  ref: `tags/${tag}`,
+                  sha: commitSha,
+                  force: true,
+                });
```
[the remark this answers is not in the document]
ada-x64: good catch

.github/actions/publish-release/action.yml
```diff
@@ -112,93 +73,104 @@ runs:
...
+                owner,
+                repo,
+                tag_name: tag,
+                target_commitish: commitSha,
+                name: `Nanvix ${version}`,
+                body: `Automated release for Nanvix ${version}.`,
+                draft: false,
+                prerelease: false,
+                make_latest: "true",
+              });
```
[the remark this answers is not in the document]
ada-x64: good catch

.github/actions/publish-release/action.yml
```diff
@@ -311,6 +274,33 @@ runs:
...
+                owner,
+                repo,
+                ref: 'refs/tags/latest',
+                sha: commitSha,
+              });
+              core.info('Created refs/tags/latest.');
+            } else {
+              // Non-critical — warn but don't fail the release.
+              core.warning(`Failed to update latest tag: ${error.message}`);
+            }
```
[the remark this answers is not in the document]
ada-x64: same issue as above, right?

ada-x64 (commented):
@copilot apply the changes :)

scripts/get-nanvix.sh
ppenna, line 1: Why are we removing this?

.github/workflows/prune-releases.yml
ppenna, line 1: Could you add a note about the mechanics of how long we keep releases around here: https://github.com/nanvix/nanvix/blob/dev/.github/skills/ci-cd-pipeline/SKILL.md ?

ppenna (changes requested):

ppenna (commented):
## Review

**Overall:** Good PR. The architecture shift from "overwrite latest" to "versioned + prune" is well-motivated and the implementation handles concurrency and idempotency thoughtfully.

### Issues to address before merge

1. **`deleteStaleDraftReleases` is overly broad** (`.github/actions/publish-release/action.yml`): The filter now deletes *all* draft releases (`releases.filter(r => r.draft)`). If someone ever creates a legitimate draft release for a different purpose, this will destroy it. Consider narrowing the filter to drafts whose `tag_name` matches the `v{semver}` pattern, or at minimum drafts older than some threshold.

2. **`createRef` fallback in `latest` tag update may fail silently**: When `updateRef` returns 404/422 for `refs/tags/latest`, you call `createRef` — but if `createRef` itself fails (permissions, another race), that error is swallowed by the outer catch's `core.warning`. Consider wrapping the `createRef` fallback in its own try/catch with a more specific warning message.

### Minor / informational

3. **No comment after concurrent-release 422 recovery with matching SHA**: After retrieving the concurrent release and confirming the SHA matches, execution falls through to asset upload. A brief `// fall through to upload` comment would improve readability.

4. **`target_commitish` comparison for idempotency**: `getReleaseByTag` returns `target_commitish` as the value passed at creation time. If a release were ever created with a branch name instead of a SHA, the comparison would fail incorrectly. In practice the code always creates with a SHA so this is fine, but a defensive comment would help future readers.

5. **Pruning could dangle `refs/tags/latest`**: If the `latest` tag happens to point to a release that gets pruned (unlikely since it should be recent), the tag would dangle. Low priority but worth a guard to skip tag deletion when it matches the `latest` pointer.

### Positives

- Immutable versioned tags with SHA verification on conflict — eliminates the tag-delete-then-recreate race condition. Major simplification.
- `isDevHead` guard prevents `make_latest` and `refs/tags/latest` regression on stale re-runs.
- Idempotency handled correctly: existing release with same SHA reuses, different SHA skips gracefully.
- Prune workflow is clean: semver regex, skip patch==0 milestones, age-based cutoff, best-effort tag cleanup, repo guard for forks.
- No stale references to `get-nanvix.sh` or `publish-latest` remain in the codebase.
- Security looks good: no injection risks, proper token scoping, tag immutability enforcement.

scripts/get-nanvix.sh
ada-x64: This has been implemented as part of zutils.
https://github.com/nanvix/zutils/blob/dev/src/nanvix_zutil/info.py

ada-x64: I didn't think about it, but it's definitely still in use. Let's restore it.

PR #1862, opened 2026-03-29
[uservm/nanvixd] E: VMM startup optimizations

ppenna:
## Summary

Reduce VMM and host startup overhead for WHP/Windows standalone mode.

## Changes

- **[uservm] E: Defer pvclock timer start** — delay the 100 Hz pvclock timer until the first application-level I/O exit, avoiding timer interrupts during the costly first `WHvRunVirtualProcessor` call (~6 ms saving on boot-time).
- **[nanvixd] E: Reduce standalone startup cost** — use single-threaded tokio runtime and skip temporary directory creation in standalone mode.

## Split from

Split from PR #1852 ("Improve cold-start performance on WHP/Windows").

PR #1868, opened 2026-03-29
[nanvixd] Use single-thread runtime in standalone mode

ppenna:
## Summary

Use a single-threaded tokio runtime in standalone mode to avoid the overhead of spawning worker threads at startup.

## Changes

- Add `#[cfg(feature = "standalone")]` block using `new_current_thread()` in `main()`.
- Gate the existing `new_multi_thread()` builder with `#[cfg(not(feature = "standalone"))]`.

## Split from

Split from PR #1862.

### DIFF
--- a/src/utils/nanvixd/src/main.rs
+++ b/src/utils/nanvixd/src/main.rs
@@ -137,6 +137,15 @@ macro_rules! log_info {
 //   `#![forbid(clippy::expect_used)]` lint configuration.
 ///
 pub fn main() -> Result<ExitCode> {
+    // Standalone mode uses a single-thread runtime to avoid the overhead of
+    // spawning worker threads at startup.  Other deployment modes keep the
+    // multi-thread runtime for higher throughput.
+    #[cfg(feature = "standalone")]
+    let rt: ::tokio::runtime::Runtime = ::tokio::runtime::Builder::new_current_thread()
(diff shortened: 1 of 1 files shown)


ppenna:
Closing this PR: switching to a single-thread Tokio runtime (`new_current_thread()`) is incompatible with the current VMM bridging pattern.

The standalone code path in `nanvix-sandbox/standalone/uservm.rs` uses `spawn_blocking` + `Handle::current().block_on()` to bridge synchronous VMM code into the async runtime. On a current-thread runtime this creates a circular dependency:

- The **runtime thread** (the only one) blocks waiting for the `spawn_blocking` task to complete.
- The **blocking-pool thread** calls `Handle::current().block_on()`, which needs the runtime thread to poll futures.

Neither thread can make progress → **deadlock**.

The same pattern repeats inside the WHP VMM layer (`vmm/whp/mod.rs: Vmm::spawn()`), which also calls `spawn_blocking` → `block_on` for thread-ID reporting and shutdown handling.

### Alternatives considered
- `new_multi_thread().worker_threads(1)` — avoids the deadlock but still spins up the multi-thread scheduler infrastructure, negating the startup-overhead benefit.
- Refactoring the VMM layer to remove `spawn_blocking` → `block_on` — too invasive for the gain.

PR #1861, opened 2026-03-29
[kernel] E: Kernel boot optimizations for WHP

ppenna:
## Summary

Optimize the kernel boot path for WHP/Windows, reducing cold-start latency.

### LAPIC Calibration
- **[kernel] E: Reduce LAPIC calibration delay** — reduce PIT-based calibration window from 10 ms to 1 ms (~9 ms saving).
- **[kernel] E: RDTSC LAPIC calibration on WHP** — replace PIT-based LAPIC timer calibration with an RDTSC spin loop, eliminating ~100 PIT-polling VM exits during boot.

### Memory Initialization
- **[kernel] E: Skip BSS clear on WHP** — WHP already zeroes guest memory via VirtualAlloc; skip redundant guest-side `rep stosb` to avoid EPT-violation faults (~2 ms saving).
- **[kernel] E: Bulk identity page table fill** — add `fill_identity()` to `PageTable` for batch PTE writes in `virt::init()`.

## Split from

Split from PR #1852 ("Improve cold-start performance on WHP/Windows").

PR #1877, opened 2026-03-30
[kernel] LAPIC calibration optimizations for WHP

ppenna:
## Summary
Optimize LAPIC timer calibration on WHP to reduce cold-start latency.

## Changes
- **Reduce LAPIC calibration delay** — reduce PIT-based calibration window from 10 ms to 1 ms (~9 ms saving).
- **RDTSC LAPIC calibration on WHP** — replace PIT-based calibration with an RDTSC spin loop, eliminating ~100 PIT-polling VM exits during boot. Uses project cpuid helper (`get_base_frequency_mhz()`) for leaf 0x16.

## Split from
Split from PR #1861.


ppenna:
@danbugs FYI

PR #1916, opened 2026-04-01
[slab] Prove liveness of Slab constructor

jaylorch:
This PR states a stronger liveness property about `RawArray::from_raw_parts` and uses it to prove that the constructor `Slab::from_raw_parts` is live, i.e., that it only fails if it's passed unreasonable parameters.

### DIFF
--- a/src/libs/slab/src/lib.proof.rs
+++ b/src/libs/slab/src/lib.proof.rs
@@ -48,6 +48,70 @@ impl Slab {
                 ==> self.index@.is_bit_set(i)
     }
 
+    proof fn lemma_wrapping_add_consequences(
+        addr: *mut u8,
+        len: usize
+    )
+        ensures
+            if ((addr as usize) + (len * size_of::<u8>())) % (usize::MAX + 1) < (addr as usize) {
+                (addr as usize) + len > usize::MAX
+            }
+            else {
+                (addr as usize) + len <= usize::MAX
+            },
+    {
+        assert(
+            if ((addr as usize) + (len * size_of::<u8>())) % (usize::MAX + 1) < (addr as usize) {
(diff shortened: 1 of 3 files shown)


src/libs/slab/src/lib.proof.rs
```diff
@@ -48,6 +48,70 @@ impl Slab {
...
+        total_num_blocks: usize,
+        num_index_blocks: usize,
+        divisor: usize,
+    )
+        requires
+            len <= isize::MAX,
+            block_size > 0,
+            total_num_blocks == len / block_size,
+            divisor == block_size * u8::BITS + 1,
+            num_index_blocks == (total_num_blocks / divisor)
```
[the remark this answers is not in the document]
jaylorch: OK, done.

src/libs/slab/src/lib.rs
```diff
@@ -104,7 +104,21 @@ impl Slab {
...
+                     &&& {
+                         ||| addr as usize == 0
+                         ||| len == 0
+                         ||| len >= i32::MAX
+                         ||| len > isize::MAX
+                         ||| addr as usize + len > usize::MAX
+                         ||| block_size == 0
+                         ||| block_size >= i32::MAX
+                         ||| block_size > (usize::MAX - 1) / (u8::BITS as int)
+                         ||| len < block_size * 2
```
ppenna, line 118: Why that?

ppenna (approved):
LGTM

src/libs/slab/src/lib.rs
jaylorch: Avoiding those conditions is what's necessary to (a) not overflow calculations, and (b) have enough room for index information. They should all be easy to avoid.

PR #1936, opened 2026-04-02
[pal][windows] Use memory-mapped I/O for file loading

esaurez:
Replace fs::read() (which copies the entire file into a Vec<u8>) with Windows CreateFileMapping/MapViewOfFile for zero-copy memory-mapped file access. This avoids heap allocation and copying for kernel ELF and initrd loading.

The FileMapping struct now holds OS handles and an mmap pointer, with proper cleanup via Drop. The public API (ptr(), size()) is unchanged.

**Benchmark (p50, 50 iterations, cold-start):**
| Metric | Before | After | Change |
|--------|--------|-------|--------|
| Total | 60,814 us | 60,713 us | -0.2% |

The improvement is marginal because kernel/initrd files are small, but this aligns the Windows PAL with the Linux mmap-based implementation and avoids unnecessary heap allocations.

Depends on #1919 (CPUID frequency override).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -220,6 +220,9 @@ windows = { version = "0.62", features = [
         "Win32_System_Hypervisor",
         "Win32_System_Memory",
         "Win32_System_Threading",
+        "Win32_Storage_FileSystem",
+        "Win32_Foundation",
+        "Win32_Security",
 ] }
 
 [profile.dev]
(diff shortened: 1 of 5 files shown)


ppenna:
Closing as stale. Two of the three proposed changes are already in dev:

- **LAPIC RDTSC-based calibration** — implemented via shared-memory frequency passing.
- **Memory-mapped I/O for file loading** — FileMapping with CreateFileMappingW/MapViewOfFile already landed.

The remaining piece (CPUID leaf 0x16 override via WHvGetCapability) is tracked in PR #1919.

PR #1957, opened 2026-04-04
[ci] Add 256MB image support for Windows

ppenna:
## Summary

Add `memory-size` matrix dimension to Windows CI build and release archive jobs so that 256MB standalone images are built and published for Windows, matching the existing Linux pipeline behavior.

## Problem

The Linux CI pipeline builds and publishes both 128MB and 256MB standalone images via its `memory-size: [128, 256]` matrix. The Windows pipeline lacked this dimension entirely — it only built with the default 128MB config and hardcoded `128mb` in release archive names.

### `windows-ci-build`
- Add `memory-size: [128, 256]` to matrix (256MB excluded from debug builds, matching Linux).
- Add `Patch memory_size in kernel config` step identical to the Linux `ci-build` job.
- Update artifact naming to include memory size suffix.

### `windows-integration-test`
- Pin artifact download to `128mb` (matching Linux `ci-test`, which only tests the default 128MB).

### `windows-release-archive`
- Add `memory-size: [128, 256]` to matrix.
- Replace hardcoded `128mb` in archive and artifact names with `${{ matrix.memory-size }}mb`.

### No changes needed
- **`publish-release`** action already discovers all `release-*` artifacts via glob.
- **Windows benchmark jobs** build from source and are unaffected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -264,24 +264,9 @@ jobs:
         uses: ./.github/actions/native-setup
 
       - name: "Patch memory_size in kernel config"
-        shell: bash
-        env:
-          MEMORY_SIZE_MB: ${{ matrix.memory-size }}
-        run: |
-          memory_size_bytes=$(( MEMORY_SIZE_MB * 1048576 ))
-          sed -i "s/^memory_size = .*/memory_size = ${memory_size_bytes}/" build/kernel_config.toml
-          # Verify that exactly one memory_size line exists after patching
-          matches=$(grep -c "^memory_size = " build/kernel_config.toml || true)
-          if [ "${matches}" -ne 1 ]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -624,6 +629,26 @@ jobs:
...
+            echo "Error: Expected exactly one 'memory_size' entry in build/kernel_config.toml, found ${matches}." >&2
+            exit 1
+          fi
+          # Verify that the memory_size line has the expected value
+          if ! grep -q "^memory_size = ${memory_size_bytes}$" build/kernel_config.toml; then
+            echo "Error: Failed to set 'memory_size' to ${memory_size_bytes} bytes in build/kernel_config.toml." >&2
+            exit 1
+          fi
+          echo "Set memory_size to ${memory_size_bytes} bytes (${MEMORY_SIZE_MB} MB)"
+
```
ppenna, line 621: This should be shared with Linux jobs.

PR #1981, opened 2026-04-07
whp: pre-populate EPT entries for kernel regions before guest execution

esaurez:
## Summary

Pre-populate the hypervisor's second-level address translation (EPT) entries using `WHvAdviseGpaRange(Populate, Write)` for targeted guest physical address ranges before the first `WHvRunVirtualProcessor` call.

## Problem

During cold-start boot, the Nanvix kernel zeroes kernel stack pages (64 KiB via `alloc_kpages`) and builds page tables. On WHP, each first guest write to a previously-untouched page triggers an EPT violation handled by the hypervisor at ~620 µs per page. This makes `forge_user_context` 76.8× slower on WHP vs KVM (9,914 µs vs 129 µs), accounting for 68% of the total performance gap.

## Approach

Call `WHvAdviseGpaRange` with `WHvAdviseGpaRangeCodePopulate` and `WHvMemoryAccessWrite` to proactively fault in EPT entries from the host side. The API internally calls `VidExoAccessVaFault()` to resolve SLAT entries without requiring guest execution.

### Targeted ranges (8 MiB total):
1. **Kernel code/data**: GPA 0..4 MiB — kernel ELF loaded at ~0x100000, guest page tables, low-memory structures
2. **Kernel pool (kpool)**: GPA 8..12 MiB (`config::kernel::KPOOL_BASE_RAW`) — where kernel stacks are allocated via `alloc_kpages`

### Why targeted instead of full memory:
After testing three configurations, the targeted 8 MiB approach provides the best trade-off:

| Config | Host cost | Guest savings | Net at p50 |
|--------|-----------|---------------|------------|
| 128 MiB full | ~14 ms (32K pages × 0.5 µs) | ~1.8 ms | **-12 ms worse** |
| 16 MiB | ~2 ms | ~1.4 ms | ~neutral |
| **8 MiB targeted** | **~1 ms** | **~1.3 ms** | **+300 µs better** |

The guest only touches ~200 pages during cold-start boot. Populating all 32K pages in 128 MiB wastes 99.4% of the effort.

## Benchmark Results

50 iterations, high-performance power mode, CPU affinity pinned:

| Metric | Baseline | With EPT populate | Delta |
|--------|----------|-------------------|-------|
| **total p50** | 60,632 µs | **60,324 µs** | -0.5% |
| **total p95** | 63,151 µs | **61,698 µs** | **-2.3%** |
| guest_exec p50 | 46,255 µs | 44,850 µs | -3.0% |
| exit_handling p50 | 7,309 µs | 7,612 µs | +4.1% |

Key observations:
- **p95 improvement is the primary benefit** (-1.5 ms) — reduced variance from EPT fault storms
- guest_exec consistently faster (-1.4 ms) as fewer in-flight EPT faults during kernel boot
- Modest p50 improvement as host populate cost partially offsets guest savings

## Implementation Details

- `vmem.rs`: Added `populate_ept()` method to `VirtualMemory` struct
- `mod.rs`: Called after `setup_pvclock()` and before wrapping vmem in `Arc<Mutex<>>`
- Uses `config::kernel::KPOOL_BASE_RAW` and `config::kernel::KPOOL_SIZE` for kpool range
- Available in `windows` crate v0.62 (`Win32_System_Hypervisor` feature)

### DIFF
--- a/src/uservm/src/vmm/microvm/whp/mod.rs
+++ b/src/uservm/src/vmm/microvm/whp/mod.rs
@@ -512,6 +512,13 @@ impl Vmm {
             Arc::new(Mutex::new(guest))
         };
 
+        // Pre-populate the hypervisor's EPT (second-level address translation) entries for
+        // the guest physical address range that the kernel will touch during boot. This avoids
+        // costly per-page EPT violations when the kernel zeroes stacks and builds page tables.
+        // Only the kernel code region (first 4 MiB) and kernel pool are populated to minimise
+        // the host-side overhead of faulting in pages the guest never touches.
+        vmem.populate_ept(partition.handle())?;
+
         let partition_handle = partition.handle();
         let timer: Arc<std::sync::Mutex<timer::Timer>> =
             Arc::new(std::sync::Mutex::new(timer::Timer::new(partition_handle)));
--- a/src/uservm/src/vmm/microvm/whp/vmem.rs
+++ b/src/uservm/src/vmm/microvm/whp/vmem.rs
@@ -24,8 +24,15 @@ use ::std::{
 };
 use windows::Win32::System::{
     Hypervisor::{
+        WHV_ADVISE_GPA_RANGE_POPULATE,
+        WHV_ADVISE_GPA_RANGE_POPULATE_FLAGS,
         WHV_MAP_GPA_RANGE_FLAGS,
+        WHV_MEMORY_RANGE_ENTRY,
+        WHV_PARTITION_HANDLE,
+        WHvAdviseGpaRange,
+        WHvAdviseGpaRangeCodePopulate,
         WHvMapGpaRange,
+        WHvMemoryAccessWrite,
     },
     Memory::{
         MEM_COMMIT,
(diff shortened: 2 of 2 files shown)


ppenna:
Closing in favor of #2009, which supersedes this PR with a broader and improved implementation of the same EPT pre-population optimization.

**Key improvements in #2009 over this PR:**

1. **Dynamic initrd coverage** — Instead of a static 8 MiB range (kernel 0..4M + kpool 8..12M), #2009 computes `[0..initrd_end)` dynamically from the loaded ELF, covering the full initrd. For large guests like Python (14 MB ELF), this populates ~26 MB of kernel+initrd pages vs only 4 MB here — yielding a **3.1% total latency reduction (-42 ms)** on Python cold-start.

2. **KVM parity** — Adds `populate_ept()` for the KVM backend using `madvise(MADV_POPULATE_WRITE)`, so both WHP and KVM benefit.

3. **Shared architecture** — Refactors the populate logic into a `Guest::ept_populate_ranges()` method in `guest.rs`, making the GPA range list explicit and backend-agnostic. This prevents silent regressions if the memory layout changes.

4. **Performance instrumentation** — Adds an `ept_populate_us` field to `PerfTimings` for tracking the populate cost in benchmarks.

5. **Broader file scope** — Changes span 6 files (`perf.rs`, `guest.rs`, `kvm/vmem.rs`, `whp/vmem.rs`, `whp/mod.rs`, `mod.rs`) vs 2 files here.

The core idea (calling `WHvAdviseGpaRange(Populate)` before guest execution) is identical — #2009 simply generalizes and extends it.

PR #1999, opened 2026-04-09
[nanvix-sandbox] Fix acceptor startup race

ppenna:
Replace the single `yield_now()` in the `Arc::new_cyclic` bootstrap with a bounded retry loop. In a multi-threaded runtime a different worker can re-poll the future before `new_cyclic` sets the strong count, causing `upgrade()` to return `None` and the accept loop to silently never start.

This fixes intermittent first-invocation linuxd control-plane timeouts observed in CI (e.g. nanvix/sqlite#24199812920).

### DIFF
--- a/src/libs/nanvix-sandbox/src/control_plane_acceptor.rs
+++ b/src/libs/nanvix-sandbox/src/control_plane_acceptor.rs
@@ -14,6 +14,7 @@
 use crate::config::{
     CONTROL_PLANE_ACCEPT_TIMEOUT,
     MAX_EARLY_CONTROL_PLANE_CONNECTIONS,
+    MAX_WEAK_UPGRADE_RETRIES,
 };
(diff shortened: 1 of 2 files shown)


src/libs/nanvix-sandbox/src/control_plane_acceptor.rs
```diff
@@ -126,14 +126,21 @@ impl ControlPlaneAcceptor {
...
+                    tokio::task::yield_now().await;
+                    if let Some(acceptor) = weak_self.upgrade() {
+                        acceptor.run().await;
+                        return;
+                    }
                 }
+                error!(
+                    "accept loop failed to start: could not upgrade weak reference after 100 \
+                     yield cycles"
+                );
```
ppenna: Replace hardcode value `100` with a file-scope constant.

PR #2033, opened 2026-04-13
[sparse-bitmap] New SparseBitmap crate

danbugs:
## Summary

Adds a new no_std `sparse-bitmap` library crate for tracking a (potentially sparse) set of bit indices as a sorted, non-overlapping `Vec<Chunk>`, where each chunk pairs a global offset with a dense backing `Bitmap`.

Memory use is proportional to the number of chunks and their sizes, not to the index space covered — suitable for the case where indices come from a wide address range (e.g. a 32-bit physical address space) but only a small fraction is tracked, and those tracked indices fall in a small number of clusters.

The dense `Bitmap` is a degenerate single-chunk `SparseBitmap` (`SparseBitmap::from_bitmap(0, ..)`), so callers can start dense and grow sparse on demand without changing the tracking type.

## Key changes

- New `src/libs/sparse-bitmap/` crate: `Vec<Chunk>` structure, O(log n) lookup via `partition_point`.
- Public API: `new` / `from_bitmap` / `add_chunk` / `set` / `clear` / `test` / `alloc` / `alloc_range` / `capacity` / `chunk_count` / `is_empty` / `covers`.
- Error semantics mirror dense `Bitmap`: `set` returns `ResourceBusy` on an already-set bit, `clear` returns `BadAddress` on an unset or uncovered bit, `test` returns `Ok(false)` on uncovered indices.
- `set_or_add_chunk(index, chunk_bits)` convenience method: sets a bit and auto-allocates an aligned chunk of `chunk_bits` bits if no existing chunk covers the index.
- Registered in the workspace and in `ALL_GUEST_RUST_LIBS_TEST_LIST`.
- 33 unit tests covering construction, chunk management, error semantics, alloc walking, high-address edge cases, and equivalence with a dense `Bitmap` for the single-chunk configuration.

## Follow-up

A separate PR (to be linked here shortly) uses this crate to refactor the kernel's `FrameAllocator` and `Kpool` around `SparseBitmap`, lifting platform-specific inner state into `hal/platform/{microvm,hyperlight}/` and removing cfg sprawl from `mm/phys/`. This split keeps the library and its kernel consumer reviewable independently.

### DIFF
--- a/src/libs/sparse-bitmap/src/lib.rs
+++ b/src/libs/sparse-bitmap/src/lib.rs
@@ -1,0 +1,847 @@
+// Copyright(c) The Maintainers of Nanvix.
+// Licensed under the MIT License.
+
+//! # Sparse Bitmap
+//!
+//! A bitmap that represents a (potentially sparse) set of bit indices as a
+//! vector of disjoint dense [`Bitmap`] chunks, each paired with the global
+//! offset it covers. Suitable when a small number of non-contiguous ranges
+//! needs to be tracked across an arbitrarily-wide index space.
+//!
+//! The shape — chunk layout, chunk sizes, chunk boundaries — is fixed at
+//! construction time. Callers must declare the full set of tracked
+//! ranges up front by passing them to [`SparseBitmap::new`]. This keeps
+//! the backing storage, the global capacity, and the per-bit validation
+//! surface predictable: there is no way to grow or shrink the collection
+//! after construction.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+//==================================================================================================
+// Extern Crates
+//==================================================================================================
+
+extern crate alloc;
+
+//==================================================================================================
+// Imports
+//==================================================================================================
+
+use ::alloc::vec::Vec;
+use ::bitmap::Bitmap;
+use ::sys::error::{
+    Error,
+    ErrorCode,
+};
+
+//==================================================================================================
+// Structures
+//==================================================================================================
+
+/// A single dense chunk in a sparse bitmap. Covers the global index range
+/// `[offset, offset + bitmap.number_of_bits())`.
+///
+/// The dense bitmap field is intentionally private: [`SparseBitmap`] owns
+/// the per-bit state. Callers that reach for a chunk (via
+/// [`SparseBitmap::find_chunk`]) only need its boundaries; they go through
+/// [`SparseBitmap::set`] / [`SparseBitmap::clear`] / [`SparseBitmap::test`]
+/// for the actual bits.
+#[derive(Debug)]
+pub struct Chunk {
+    offset: usize,
+    bitmap: Bitmap,
+}
+
+impl Chunk {
+    /// Global index of the first bit covered by this chunk.
+    pub fn offset(&self) -> usize {
+        self.offset
+    }
+
+    /// Number of bits in this chunk.
+    pub fn num_bits(&self) -> usize {
+        self.bitmap.number_of_bits()
+    }
+
+    /// End of the chunk's covered range (exclusive).
+    pub fn end(&self) -> usize {
+        self.offset + self.bitmap.number_of_bits()
+    }
+
+    /// Whether the chunk covers `index`.
+    fn covers(&self, index: usize) -> bool {
+        index >= self.offset && index < self.end()
+    }
+}
+
+///
+/// # Description
+///
+/// A sparse bitmap: a bitmap whose storage is distributed across a fixed
+/// number of dense [`Bitmap`] chunks declared at construction time. Memory
+/// use is proportional to the number of chunks and their sizes, not to
+/// the index space covered.
+///
+/// Each chunk is identified by its *global offset* — the starting index it
+/// covers in the "global view." Chunks are kept sorted by offset and are
+/// guaranteed to be non-overlapping. The set of chunks does not change
+/// after construction.
+///
+/// ## When to use
+///
+/// Prefer [`SparseBitmap`] over [`Bitmap`] when:
+///
+/// - Indices come from a wide address space (e.g. the full 32-bit physical
+///   address space) but only a small fraction is actually tracked, and
+/// - Those tracked indices fall in a small number of clusters, so per-chunk
+///   fixed overhead is amortized across many indices.
+///
+/// A dense [`Bitmap`] is strictly more memory-efficient when nearly every
+/// index in its range is active. A [`SparseBitmap`] with a single chunk
+/// at offset 0 is semantically equivalent to a [`Bitmap`].
+///
+/// ## Error semantics
+///
+/// The API mirrors [`Bitmap`]'s: [`Self::set`] returns
+/// [`ErrorCode::ResourceBusy`] when the bit is already set, [`Self::clear`]
+/// returns [`ErrorCode::BadAddress`] when the bit is not set (including
+/// when no chunk covers the index), and [`Self::test`] returns `Ok(false)`
+/// for uncovered indices rather than an error.
+///
+#[derive(Debug)]
+pub struct SparseBitmap {
+    /// Chunks, sorted by `offset` and non-overlapping. Populated at
+    /// construction time by [`Self::new`]; never mutated afterwards.
+    chunks: Vec<Chunk>,
+    /// Total capacity across all chunks, in bits. Computed at
+    /// construction time so [`Self::capacity`] is O(1).
+    capacity_bits: usize,
+    /// Index into `chunks` to try first on the next [`Self::alloc`] or
+    /// [`Self::alloc_range`] call. Amortizes searches when earlier
+    /// chunks are fully allocated: a successful alloc updates this
+    /// hint to the satisfying chunk, so subsequent allocs resume there
+    /// rather than re-scanning from the front.
+    next_chunk_hint: usize,
+}
+
+//==================================================================================================
+// Implementations
+//==================================================================================================
+
+impl SparseBitmap {
+    ///
+    /// # Description
+    ///
+    /// Creates a sparse bitmap from the full, pre-provisioned set of
+    /// `(offset, bitmap)` chunks. The shape is fixed at construction:
+    /// there is no later grow/shrink. Callers that need several
+    /// disjoint tracked ranges must declare them all here.
+    ///
+    /// Each bitmap's backing storage is chosen by the caller — heap via
+    /// [`Bitmap::new`], or a static BSS buffer via [`Bitmap::from_raw_array`]
+    /// when no allocator is available.
+    ///
+    /// # Parameters
+    ///
+    /// - `chunks`: Pairs `(offset, bitmap)` where `offset` is the global
+    ///   index of bit 0 of `bitmap`. Order doesn't matter — the
+    ///   constructor sorts by offset. Ranges must not overlap.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, the constructed [`SparseBitmap`]. Upon failure:
+    /// - [`ErrorCode::InvalidArgument`] if `chunks` is empty, if a
+    ///   chunk's range would overflow `usize`, or if two chunks'
+    ///   ranges overlap.
+    ///
+    pub fn new(chunks: Vec<(usize, Bitmap)>) -> Result<Self, Error> {
+        if chunks.is_empty() {
+            return Err(Error::new(
+                ErrorCode::InvalidArgument,
+                "sparse bitmap requires at least one chunk",
+            ));
+        }
+        // Validate each individual chunk's overflow guard and compute
+        // total capacity.
+        let mut capacity_bits: usize = 0;
+        let mut built: Vec<Chunk> = Vec::with_capacity(chunks.len());
+        for (offset, bitmap) in chunks {
+            let num_bits = bitmap.number_of_bits();
+            offset.checked_add(num_bits).ok_or_else(|| {
+                Error::new(ErrorCode::InvalidArgument, "chunk range overflows usize")
+            })?;
+            capacity_bits = capacity_bits.checked_add(num_bits).ok_or_else(|| {
+                Error::new(ErrorCode::InvalidArgument, "capacity overflows usize")
+            })?;
+            built.push(Chunk { offset, bitmap });
+        }
+
+        // Sort by offset, then reject overlaps by walking the sorted
+        // list pairwise.
+        built.sort_by_key(|c| c.offset);
+        for pair in built.windows(2) {
+            if pair[0].end() > pair[1].offset {
+                return Err(Error::new(ErrorCode::InvalidArgument, "chunk overlaps another chunk"));
+            }
+        }
+
+        Ok(Self {
+            chunks: built,
+            capacity_bits,
+            next_chunk_hint: 0,
+        })
+    }
+
+    ///
+    /// # Description
+    ///
+    /// Sets the bit at the given global `index`.
+    ///
+    /// # Parameters
+    ///
+    /// - `index`: Global index of the bit to set.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, `Ok(())`. Upon failure:
+    /// - [`ErrorCode::InvalidArgument`] if no chunk covers `index`.
+    /// - [`ErrorCode::ResourceBusy`] if the bit is already set.
+    /// - Any error propagated from [`Bitmap::set`].
+    ///
+    pub fn set(&mut self, index: usize) -> Result<(), Error> {
+        match self.find_chunk_mut(index) {
+            Some(chunk) => {
+                let local = index - chunk.offset;
+                chunk.bitmap.set(local)
+            },
+            None => Err(Error::new(ErrorCode::InvalidArgument, "no chunk covers index")),
+        }
+    }
+
+    ///
+    /// # Description
+    ///
+    /// Clears the bit at the given global `index`.
+    ///
+    /// # Parameters
+    ///
+    /// - `index`: Global index of the bit to clear.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, `Ok(())`. Upon failure:
+    /// - [`ErrorCode::BadAddress`] if no chunk covers `index` (treated as
+    ///   "bit is not set," matching [`Bitmap::clear`]'s semantics for a
+    ///   cleared bit), or if the bit is already cleared.
+    /// - Any error propagated from [`Bitmap::clear`].
+    ///
+    pub fn clear(&mut self, index: usize) -> Result<(), Error> {
+        match self.find_chunk_mut(index) {
+            Some(chunk) => {
+                let local = index - chunk.offset;
+                chunk.bitmap.clear(local)
+            },
+            None => Err(Error::new(ErrorCode::BadAddress, "bit is not set")),
+        }
+    }
+
+    ///
+    /// # Description
+    ///
+    /// Returns whether the bit at the given global `index` is set.
+    ///
+    /// # Parameters
+    ///
+    /// - `index`: Global index of the bit to test.
+    ///
+    /// # Returns
+    ///
+    /// `Ok(true)` if the bit is set; `Ok(false)` if it is cleared, *or* if
+    /// no chunk covers `index`. Propagates errors from [`Bitmap::test`].
+    ///
+    pub fn test(&self, index: usize) -> Result<bool, Error> {
+        match self.find_chunk(index) {
+            Some(chunk) => {
+                let local = index - chunk.offset;
+                chunk.bitmap.test(local)
+            },
+            None => Ok(false),
+        }
+    }
+
+    ///
+    /// # Description
+    ///
+    /// Allocates a single free bit, sets it, and returns its global
+    /// index.
+    ///
+    /// Thin wrapper over [`Self::alloc_range`] with `count = 1`. Lets
+    /// single-bit and range allocation share the same hint-maintenance
+    /// and cross-chunk logic.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, the global index of the allocated bit. Upon failure,
+    /// [`ErrorCode::OutOfMemory`] if no chunk has a free bit.
+    ///
+    pub fn alloc(&mut self) -> Result<usize, Error> {
+        self.alloc_range(1)
+    }
+
+    ///
+    /// # Description
+    ///
+    /// Allocates a contiguous range of `count` free bits, sets them,
+    /// and returns the global index of the first bit.
+    ///
+    /// ## Search order
+    ///
+    /// 1. **Single-chunk pass.** Starting from the cached
+    ///    [`Self::next_chunk_hint`] and wrapping around, try
+    ///    [`Bitmap::alloc_range`] on each chunk whose capacity is at
+    ///    least `count`. Fast path.
+    ///
+    /// 2. **Cross-chunk pass.** If no single chunk fits, walk chunk
+    ///    pairs looking for a run that starts inside one chunk, spans
+    ///    its suffix, and continues into the prefix of the next chunk
+    ///    (and possibly further chunks beyond that). Only *touching*
+    ///    chunks — where `chunk[i].end() == chunk[i+1].offset` — are
+    ///    eligible, since a gap between chunks means the global indices
+    ///    aren't actually contiguous.
+    ///
+    /// The cross-chunk pass preserves the invariant that every returned
+    /// range is genuinely contiguous in the global index space. Callers
+    /// that don't want cross-chunk allocations can pre-provision non-
+    /// touching chunks.
+    ///
+    /// # Parameters
+    ///
+    /// - `count`: Number of contiguous bits to allocate. Must be non-zero.
+    ///
+    /// # Returns
+    ///
+    /// Upon success, the global index of the first bit in the allocated
+    /// range. Upon failure:
+    /// - [`ErrorCode::InvalidArgument`] if `count == 0`.
+    /// - [`ErrorCode::OutOfMemory`] if neither pass can satisfy the
+    ///   request.
+    ///
+    pub fn alloc_range(&mut self, count: usize) -> Result<usize, Error> {
+        if count == 0 {
+            return Err(Error::new(ErrorCode::InvalidArgument, "count must be non-zero"));
+        }
+        // By construction (`Self::new` rejects empty input), `chunks`
+        // has at least one entry.
+        let n = self.chunks.len();
+
+        // Pass 1: single chunk. `next_chunk_hint` is always a valid
+        // index into `chunks` — initialised to 0 at construction and
+        // only ever updated to an index that satisfied a prior
+        // allocation, and the chunk set is fixed, so no clamp is
+        // needed.
+        let start = self.next_chunk_hint;
+        for step in 0..n {
+            let idx = (start + step) % n;
+            let chunk = &mut self.chunks[idx];
+            if count > chunk.bitmap.number_of_bits() {
+                continue;
+            }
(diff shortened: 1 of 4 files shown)


src/libs/bitmap/src/sparse.rs
```diff
@@ -0,0 +1,239 @@
...
+/// allocated frame numbers at arbitrary 32-bit addresses, where a dense
+/// bitmap would need to be sized for the entire address space.
+///
+/// Operations are `O(log n)` in the number of currently-set bits.
+///
+#[derive(Debug, Default)]
+pub struct SparseBitmap {
+    /// Set of indices whose bit is currently set.
+    set: BTreeSet<usize>,
+}
```
ppenna: We should rework this abstraction. The current implementation is extremely memory intensive.

We should have a `Vec` (expecting a small about of non-contiguous ranges) that points to several bitmap structures, each of which with an additional `usize`, indicating its offset in the "global view".

Furthermore, this should be turned into a crate itself that depends on bitmap.

ppenna (changes requested):

src/libs/bitmap/src/sparse.rs
danbugs: Reworked. The crate is now a standalone `sparse-bitmap` library using a `Vec<Chunk>` where each `Chunk { offset: usize, bitmap: Bitmap }` stitches a dense bitmap into the sparse view at a global offset. The dense `Bitmap` is now a single-chunk `SparseBitmap` (`SparseBitmap::from_bitmap(0, ..)`), which is what `FrameAllocator` and `Kpool` actually wrap–no separate `BTreeSet` side-pool.

src/libs/bitmap/src/sparse.rs
```diff
@@ -0,0 +1,239 @@
...
+    ///
+    /// # Returns
+    ///
+    /// Upon success, `Ok(())` is returned. Returns
+    /// [`ErrorCode::InvalidArgument`] if the bit is already set.
+    ///
+    pub fn set(&mut self, index: usize) -> Result<(), Error> {
+        if !self.set.insert(index) {
+            return Err(Error::new(
+                ErrorCode::InvalidArgument,
```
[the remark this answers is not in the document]
danbugs: Fixed. `SparseBitmap::set()` now returns `ErrorCode::ResourceBusy` on a bit that's already set, matching dense `Bitmap::set()`. The error code distinction is load-bearing for `FrameAllocator::free`'s silent-on-`BadAddress` path on Hyperlight, so keeping the codes parallel matters.

src/libs/bitmap/src/sparse.rs
```diff
@@ -0,0 +1,239 @@
...
+    ///
+    /// # Returns
+    ///
+    /// Upon success, `Ok(())` is returned. Returns
+    /// [`ErrorCode::InvalidArgument`] if the bit was not set.
+    ///
+    pub fn clear(&mut self, index: usize) -> Result<(), Error> {
+        if !self.set.remove(&index) {
+            return Err(Error::new(
+                ErrorCode::InvalidArgument,
```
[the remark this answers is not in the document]
danbugs: Fixed. `SparseBitmap::clear()` now returns `ErrorCode::BadAddress` both when the bit is already clear and when no chunk covers the index, matching dense `Bitmap::clear()`.

src/kernel/src/mm/phys/frame.rs
```diff
@@ -115,25 +138,74 @@ impl FrameAllocator {
...
+    /// Upon success, `Ok(())` is returned. Upon failure, an error is
+    /// returned instead. It is an error to free a frame that was never
+    /// registered with this allocator.
     ///
     pub fn free(&mut self, frame: FrameAddress) -> Result<(), Error> {
-        let frame_number: usize = frame.into_frame_number().into_raw_value();
-        match self.bitmap.clear(frame_number) {
-            Ok(()) => Ok(()),
-            Err(error) => {
+        let frame_number: usize = frame.into_raw_value() / mem::FRAME_SIZE;
```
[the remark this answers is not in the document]
danbugs: Fixed. `FrameAllocator::free` now uses `frame.into_frame_number().into_raw_value()`–drops the implicit `FRAME_SIZE == page_size` assumption.

src/kernel/src/mm/phys/frame.rs
```diff
@@ -115,25 +138,74 @@ impl FrameAllocator {
...
+        if frame_number < self.bitmap.number_of_bits() {
+            if let Err(error) = self.bitmap.clear(frame_number) {
                 error!("{error:?} (frame={frame:?})");
-                Err(error)
-            },
+                return Err(error);
+            }
+            Ok(())
+        } else if self.foreign.test(frame_number) {
+            // Safe to unwrap: we just observed the bit is set.
```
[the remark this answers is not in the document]
danbugs: Fixed. The comment is gone–after the rework, free is a single dispatch into `SparseBitmap::clear()` with no test-then-clear pattern to need a comment.

src/kernel/src/mm/phys/frame.rs
```diff
@@ -115,25 +138,74 @@ impl FrameAllocator {
...
+    /// or if the frame has already been registered.
+    ///
+    #[allow(dead_code)]
+    pub fn alloc_foreign(&mut self, frame: FrameAddress) -> Result<(), Error> {
+        let frame_number: usize = frame.into_raw_value() / mem::FRAME_SIZE;
+        if frame_number < self.bitmap.number_of_bits() {
+            return Err(Error::new(
+                ErrorCode::InvalidArgument,
+                "alloc_foreign: frame is inside bitmap range",
+            ));
```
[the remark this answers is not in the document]
danbugs: Renamed to `book_foreign`. Aligns with the existing book verb in the same file (used for marking dense-range frames as allocated).

danbugs:
A note for reviewing what* this PR sets up that isn't visible from the diff alone.

The platform-isolation in `Kpool`/`FrameAllocator` inner state being moved under hal/platform/microvm/ plus the HL stubs are scaffolding for the in-progress Nanvix-on-Hyperlight i686 guest with CoW snapshot/restore (development branch [hyperlight-no-unstable-dev-v2](https://github.com/nanvix/nanvix/tree/hyperlight-no-unstable-dev-v2)). On that branch the HL Inners diverge:
- `Kpool` draws from a scratch bump cursor instead of a fixed pool region. The returned frames are EPT-writable and survive snapshot restore.
- `FrameAllocator` `alloc()` does the same on the user-frame path. Both use `SparseBitmap::set_or_add_chunk` to grow chunks dynamically as scratch frames are booked.
- `adjust_kernel_frame_to_scratch` goes away–the platform module owns scratch semantics directly.
- Because the `hal::platform::{microvm,hyperlight}::*;` dispatch is pub use'd at the platform layer, `mm/phys/{kpool,frame}.rs` doesn't need any `cfg` gates: the same `Kpool`/`FrameAllocator` shape is wrapped over a different `Inner` per platform. When the HL feature lands upstream, only the contents of `hal/platform/hyperlight/{kpool,frame}.rs` change–the FIXME-marked stubs are wholesale replaced.

Net effect this PR is angling toward: one `SparseBitmap` abstraction, one set of `mm/phys` facades, two platform `Inner`s that encode genuinely different allocation strategies (dense bitmap vs. scratch bump), zero `cfg` in `mm/phys/`.

ppenna (changes requested):
Please make this PR only about `libs/sparse-bitmap` for now. Remove from PR all changes made to the kernel.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+    /// # Parameters
+    ///
+    /// - `offset`: Global index of bit 0 of `bitmap`.
+    /// - `bitmap`: Dense bitmap to wrap.
+    ///
+    pub fn from_bitmap(offset: usize, bitmap: Bitmap) -> Self {
+        let mut sparse = Self::new();
+        sparse.chunks.push(Chunk { offset, bitmap });
+        sparse
+    }
```
ppenna: We should enable this to be an array of pre-provisioned bitmaps with offsets, potentially to enable static allocation from BSS memory.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+
+        let bitmap = Bitmap::new(num_bits)?;
+        let insert_at = self
+            .chunks
+            .iter()
+            .position(|c| c.offset > offset)
+            .unwrap_or(self.chunks.len());
+        self.chunks.insert(insert_at, Chunk { offset, bitmap });
+        Ok(())
+    }
```
ppenna: This will use managed memory and it's likely what we dont want. Probably we want to have `add_chunk` where we add a pre-provisioned bitmap as well? Also, this encourage people to dynamically grow the size of the structure, which will make consumption and reasoning harder.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+            match chunk.bitmap.alloc_range(count) {
+                Ok(local) => return Ok(chunk.offset + local),
+                Err(e) if e.code == ErrorCode::OutOfMemory => continue,
+                Err(e) => return Err(e),
+            }
+        }
+        Err(Error::new(
+            ErrorCode::OutOfMemory,
+            "no chunk has a contiguous free range of the requested size",
+        ))
```
ppenna: What about allocations that live between frames?

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+    /// # Description
+    ///
+    /// Returns the total capacity across all chunks, in bits.
+    ///
+    pub fn capacity(&self) -> usize {
+        self.chunks
+            .iter()
+            .map(|c| c.bitmap.number_of_bits())
+            .sum()
+    }
```
ppenna, line 473: We should simply cache this info in the structure.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+    ///
+    pub fn set_or_add_chunk(&mut self, index: usize, chunk_bits: usize) -> Result<(), Error> {
+        if !self.covers(index) {
+            let chunk_offset = (index / chunk_bits) * chunk_bits;
+            self.add_chunk(chunk_offset, chunk_bits)?;
+        }
+        self.set(index)
+    }
+
+    /// Finds the chunk containing `index`, if any.
```
ppenna: Please use same documentation style of previous functions.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+    fn find_chunk(&self, index: usize) -> Option<&Chunk> {
```
ppenna: Please use same documentation style of previous functions.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+/// [`ErrorCode::ResourceBusy`] when the bit is already set, [`Self::clear`]
+/// returns [`ErrorCode::BadAddress`] when the bit is not set (including
+/// when no chunk covers the index), and [`Self::test`] returns `Ok(false)`
+/// for uncovered indices rather than an error.
+///
+#[derive(Debug, Default)]
+pub struct SparseBitmap {
+    /// Chunks, kept sorted by `offset` and non-overlapping.
+    chunks: Vec<Chunk>,
+}
```
ppenna, line 125: You may want to include the index of the latest bit allocated and use that to amortize allocations while bitmap is not full.

ppenna (changes requested):

src/libs/sparse-bitmap/src/lib.rs
danbugs: Added a `next_chunk_hint: usize` field–index into chunks where the last successful allocation landed. `alloc()` and `alloc_range()` both start their scan from the hint and wrap around, so repeated allocations against the same not-yet-full chunk are amortized. When the hinted chunk fills up, the walk continues and the hint moves with it. Added `alloc_hint_amortizes_subsequent_allocs` covering the happy path and the wrap-on-free case.

src/libs/sparse-bitmap/src/lib.rs
danbugs: Done.

src/libs/sparse-bitmap/src/lib.rs
danbugs: Done.

src/libs/sparse-bitmap/src/lib.rs
danbugs: Added a `capacity_bits: usize` field. Added `from_bitmap_initializes_cached_capacity` and strengthened `capacity_sums_chunks` (renamed `capacity_is_cached_and_maintained`).

src/libs/sparse-bitmap/src/lib.rs
danbugs: `alloc_range` docstring now explains the design decision: ranges spanning chunk boundaries are not supported because chunks represent disjoint regions in the global index space–stitching would either cross a gap (incorrect) or silently ignore the boundary. Callers that need contiguous cross-chunk ranges should declare a single larger chunk up front. Added a regression test (`alloc_range_does_not_stitch_across_chunks`) that asserts the refusal. What do you think?

src/libs/sparse-bitmap/src/lib.rs
danbugs: Same change covers this–`Bitmap::new` is no longer called inside `add_chunk`. The `set_or_add_chunk` helper was dropped entirely; callers that need that behaviour can compose covers -> add_chunk -> set explicitly.

src/libs/sparse-bitmap/src/lib.rs
danbugs: Changed add_chunk signature from `add_chunk(offset, num_bits)` to `add_chunk(offset, bitmap: Bitmap)`. The caller now owns the backing storage decision: heap (`Bitmap::new`), static BSS buffer (`Bitmap::from_raw_array`), or anywhere else. The library no longer allocates on the caller's behalf.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,851 @@
...
+            }
+        }
+
+        let insert_at = self
+            .chunks
+            .iter()
+            .position(|c| c.offset > offset)
+            .unwrap_or(self.chunks.len());
+        self.chunks.insert(insert_at, Chunk { offset, bitmap });
+        self.capacity_bits += num_bits;
```
ppenna: As we may insert in-between the hint pointer, we may need to update it.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,851 @@
...
+    /// [`ErrorCode::OutOfMemory`] if no chunk has a free bit.
+    ///
+    pub fn alloc(&mut self) -> Result<usize, Error> {
+        let n = self.chunks.len();
+        if n == 0 {
+            return Err(Error::new(
+                ErrorCode::OutOfMemory,
+                "no free bit in any chunk",
+            ));
+        }
```
ppenna: See my comment on constructor.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,851 @@
...
+                "count must be non-zero",
+            ));
+        }
+        let n = self.chunks.len();
+        if n == 0 {
+            return Err(Error::new(
+                ErrorCode::OutOfMemory,
+                "no chunk has a contiguous free range of the requested size",
+            ));
+        }
```
ppenna: See my comment on constructor

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,851 @@
...
+                },
+                Err(e) if e.code == ErrorCode::OutOfMemory => continue,
+                Err(e) => return Err(e),
+            }
+        }
+        Err(Error::new(
+            ErrorCode::OutOfMemory,
+            "no free bit in any chunk",
+        ))
+    }
```
ppenna, line 289: This should be built on top of `self.alloc_range(1)`

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,851 @@
...
+                },
+                Err(e) if e.code == ErrorCode::OutOfMemory => continue,
+                Err(e) => return Err(e),
+            }
+        }
+        Err(Error::new(
+            ErrorCode::OutOfMemory,
+            "no chunk has a contiguous free range of the requested size",
+        ))
+    }
```
ppenna, line 375: This is still not handling allocations that may span across chunks.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,851 @@
...
+
+    ///
+    /// # Description
+    ///
+    /// Returns whether the given global `index` falls within any existing
+    /// chunk.
+    ///
+    pub fn covers(&self, index: usize) -> bool {
+        self.find_chunk(index).is_some()
+    }
```
ppenna: Drop this function and just expose `find_chunk`

ppenna (changes requested):

src/libs/sparse-bitmap/src/lib.rs
danbugs: Done. `covers` is gone; `find_chunk` is now `pub fn find_chunk(&self, index: usize) -> Option<&Chunk>`. To let callers actually use the returned reference, I also made `Chunk` public with `offset()`, `len()`, and `end()` accessors–the inner `Bitmap` stays private so per-bit state still only moves through `set` / `clear` / `test`.

A caller that only wants the yes/no answer writes `s.find_chunk(idx).is_some()`. A caller that's trying to decide whether to register a neighbouring chunk (the Hyperlight scratch-books use case) gets the existing chunk's `offset`/`end` for free from the same lookup.

The tests previously calling `.covers(idx)` are updated to `.find_chunk(idx).is_some()`.

src/libs/sparse-bitmap/src/lib.rs
danbugs: Addressed in the updated `alloc_range`. The implementation now has two passes:

1. **Single-chunk pass**, starts at the hint, walks all chunks once, tries `Bitmap::alloc_range` on each whose capacity is >= `count`. This stays the fast path.
2. **Cross-chunk pass** fires only when no single chunk fits. `try_alloc_cross_chunk_from(entry, count)` looks for a run that starts inside `chunk[entry]`'s trailing-free suffix and continues into the prefix of `chunk[entry + 1]`, `chunk[entry + 2]`, etc. Only *touching* chunks are eligible–`chunk[i].end() == chunk[i+1].offset`–because a gap means the global indices aren't contiguous.

The old `alloc_range_does_not_stitch_across_chunks` test is inverted into `alloc_range_stitches_across_touching_chunks`. Added `alloc_range_refuses_to_stitch_across_gap` to pin the non-touching case, `alloc_range_stitches_across_three_chunks` for a three-chunk walk, and `alloc_range_stitch_requires_free_tail_of_entry` for the case where the entry chunk's tail is occupied so stitching is impossible.

src/libs/sparse-bitmap/src/lib.rs
danbugs: Done. `alloc` is now a one-liner `self.alloc_range(1)`, and the chunk-walk + hint-maintenance logic lives in one place. The earlier split implementation also meant cross-chunk handling (your next comment) only had to be written once.

src/libs/sparse-bitmap/src/lib.rs
danbugs: Updated `add_chunk` to maintain the hint invariant across insertions:
- If insertion at `insert_at` happens at or before `next_chunk_hint`, the chunk the hint was pointing at shifts one position to the right, so the hint bumps too.
- If insertion happens after the hint, the hint keeps pointing at the same chunk.

Note: a fresh `SparseBitmap::new()` has `next_chunk_hint = 0` but no chunks–the hint is "empty." I snapshot `hint_was_valid = next_chunk_hint < chunks.len()` before mutating so the very first `add_chunk` doesn't bump the hint past its freshly-inserted chunk. Caught by running the tests after the hint-maintenance change–`alloc_walks_chunks` regressed until the `hint_was_valid` guard was added.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,989 @@
...
+    /// error for every index until a chunk is added via
+    /// [`Self::add_chunk`] or via [`Self::from_bitmap`].
+    ///
+    pub const fn new() -> Self {
+        Self {
+            chunks: Vec::new(),
+            capacity_bits: 0,
+            next_chunk_hint: 0,
+        }
+    }
```
ppenna: There should not be a concept as a sparse bitmap:
- We must not support dynamic growth and shrink of this collection.
- This makes validations harder.

ppenna: This should be turned into const a constructor where we receive an array of of `(usize, Bitmap)`.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,989 @@
...
+        self.capacity_bits += num_bits;
+
+        // Keep the hint pointing at the same chunk it used to. Inserting
+        // at or before the hint shifts the hinted chunk one position to
+        // the right; inserting after it leaves the hint alone.
+        if hint_was_valid && insert_at <= self.next_chunk_hint {
+            self.next_chunk_hint += 1;
+        }
+        Ok(())
+    }
```
ppenna: We should not support dynamic growth of this collection. Drop this function.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,989 @@
...
+        // Pass 2: cross-chunk over touching neighbours. Scan forward-
+        // only starting from each possible entry chunk; we never wrap
+        // across the Vec end here because a range that "wraps" would by
+        // definition skip whatever sits before index 0.
+        for entry in 0..n {
+            if let Some(global_start) = self.try_alloc_cross_chunk_from(entry, count)? {
+                self.next_chunk_hint = entry;
+                return Ok(global_start);
+            }
+        }
```
ppenna, line 368: This incurs in a quadratic cost and may be bad for systems that rely on this. SInce we own the chunks we can add tracking logic to the first and last bit set of each logic and use it for handling this case. I am fining about open an issue to keep track of this improvement. If so, please link issue here with `TODO: (#XXX)`

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,989 @@
...
+        for bit in start_bit_in_entry..(start_bit_in_entry + take_from_entry) {
+            self.chunks[entry].bitmap.set(bit)?;
+        }
+        let mut cur = entry;
+        for take in consumed_per_chunk {
+            cur += 1;
+            for bit in 0..take {
+                self.chunks[cur].bitmap.set(bit)?;
+            }
+        }
```
ppenna, line 461: What if fail half-way through? Per construction it should not happen, so use an expect instead or, better add recover logic and than fail gracefully.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,989 @@
...
+        if idx == 0 {
+            return None;
+        }
+        let chunk = &mut self.chunks[idx - 1];
+        if chunk.covers(index) {
+            Some(chunk)
+        } else {
+            None
+        }
+    }
```
ppenna, line 528: Better:
```rust
/// Returns the index into `self.chunks` of the chunk covering `index`, if any.
fn find_chunk_index(&self, index: usize) -> Option<usize> {
    let idx = self.chunks.partition_point(|c| c.offset <= index);
    if idx == 0 {
        return None;
    }
    let candidate = idx - 1;
    if self.chunks[candidate].covers(index) {
        Some(candidate)
    } else {
        None
    }
}

pub fn find_chunk(&self, index: usize) -> Option<&Chunk> {
    self.find_chunk_index(index).map(|i| &self.chunks[i])
}

fn find_chunk_mut(&mut self, index: usize) -> Option<&mut Chunk> {
    self.find_chunk_index(index).map(|i| &mut self.chunks[i])
}
```

danbugs: Took your sketch almost verbatim. `find_chunk_index(&self, index) -> Option<usize>` is the only place the `partition_point` walk lives. `find_chunk` and `find_chunk_mut` are one-line wrappers over it that `.map` into `&self.chunks[i]` / `&mut self.chunks[i]` respectively. Borrow checker is happy because `find_chunk_index` takes `&self` and the mutable borrow only starts after the helper returns.

src/libs/sparse-bitmap/src/lib.rs
danbugs: Switched to `.expect()`. The preceding check walk verifies every bit in the proposed range is free, so a failure in `bitmap.set` would be an internal-invariant break in `Bitmap` rather than a runtime condition we can meaningfully recover from. Added a comment at the commit block spelling out why expect is load-bearing–a partial-commit path would leave bits set with no caller-visible free that maps to them, and any "recovery" would have to re-read the exact state we just validated.

src/libs/sparse-bitmap/src/lib.rs
danbugs: Kept the trailing-free / leading-free walk for this first cut (it's only hit when the single-chunk fast path can't satisfy the request) but called out the cost and the fix explicitly with `TODO (#2058)` right above `try_alloc_cross_chunk_from`. When you open the issue I'll swap the placeholder for the real number and do the caching work there. The optimisation is local–since every bit mutation goes through `SparseBitmap::{set, clear, alloc, alloc_range}`, `Chunk` can cache `first_set`/`last_set` on each call and serve both the forward and reverse scans in O(1).

src/libs/sparse-bitmap/src/lib.rs
danbugs: Dropped together with `from_bitmap` and the hint-maintenance logic that existed solely to cover insertion into the middle. All chunks go through the new new constructor.

src/libs/sparse-bitmap/src/lib.rs
danbugs: Dropped dynamic growth entirely. `from_bitmap` and `add_chunk` are gone; the only constructor is `pub fn new(chunks: Vec<(usize, Bitmap)>) -> Result<Self, Error>`, which takes the full chunk set up front, sorts by offset, and rejects overlap + `usize` overflow in one pass. After construction the chunk vector is never mutated. This also lets `capacity_bits` be computed once at construction and read O(1) forever after, and removes the `next_chunk_hint bookkeeping` that `add_chunk` needed to keep valid.

src/libs/sparse-bitmap/src/lib.rs
```diff
@@ -0,0 +1,872 @@
...
+                "count must be non-zero",
+            ));
+        }
+        let n = self.chunks.len();
+        if n == 0 {
+            return Err(Error::new(
+                ErrorCode::OutOfMemory,
+                "no chunks registered",
+            ));
+        }
```
ppenna: By construction this cannot be 0. Double check for similar logic in the file.

ppenna (dismissed):
After fix, squash commits using a detailed commit message, rebase on top of latest origin/dev and mark PR as open.

PR #2042, opened 2026-04-14
[kernel] E: Batch page allocation in ELF loader

esaurez:
## Summary

Batch page allocation in the ELF segment loader and chunk-based processing
in `alloc_upages` to reduce the cost of loading large ELF binaries.

## Problem

The ELF loader allocates every page of every segment individually via
`alloc_upages(vmem, vaddr, 1, ...)`. For a ~14 MB binary like Python 3.12
(~3,584 pages), this results in ~3,584 separate frame-allocator + page-table
operations, each involving a bitmap scan, PTE write, and (for BSS pages) a
full-page memset.

### `src/kernel/src/mm/elf.rs` -- Batch ELF segment allocation

The ELF segment loading loop is restructured into two phases:

1. **Phase 1 (allocation):** Scan pages within each segment and collect
   contiguous runs of unmapped pages that share the same `clear` requirement.
   Allocate each run with a single `alloc_upages(nframes)` call. This
   typically produces 2-3 runs per segment (one for data-backed pages, one
   for BSS pages).

2. **Phase 2 (data copy):** Copy segment data into the allocated pages,
   unchanged from the original per-page copy loop.

Pages that overlap with prior segments (requiring permission merging via
`ctrl_upage`) flush the current run before being handled individually, then
a new run starts after the overlap.

### `src/kernel/src/mm/virt/manager.rs` -- Chunked `alloc_upages`

`alloc_upages` is refactored to process frames in fixed-size chunks
(`MAX_CHUNK = 32`). Each chunk allocates a small `Vec<UserFrame>` that is
dropped before the next chunk begins. This prevents large batch requests
(e.g., 1000+ pages for a BSS segment) from exhausting the 128 KB kernel slab.

Error handling is improved with two helpers:
- `rollback_mapped_pages`: unmaps pages that were successfully mapped
- `free_remaining_frames`: frees allocated but unmapped frames to prevent
  resource leaks when a map or memset operation fails mid-chunk

## Benchmark (Python 3.12 hello world, WHP standalone, on `bugfix-perf-regressions`)

| Metric | Before | After | Change |
|---|---|---|---|
| `guest_exec` | 2,757,370 us | 2,439,118 us | **-11.5%** |

Profiling with `rdtsc` markers showed the `spawn_servers` phase (ELF loading +
process creation) consumed 94% of kernel boot time before this change.

### DIFF
--- a/src/kernel/src/mm/elf.rs
+++ b/src/kernel/src/mm/elf.rs
@@ -263,32 +263,48 @@ fn do_elf32_load(
 
         let phys_addr_end: usize = phys_addr_base + phdr.p_filesz as usize;
 
-        // Load segment page by page.
+        // Load segment: batch-allocate contiguous page runs, then copy data.
         debug!(
             "loading segment (virt_addr_base={:#x}, virt_addr_end={:#x}, phys_addr_base={:#x}, \
              phys_addr_end={:#x}, access={:?})",
             virt_addr_base, virt_addr_end, phys_addr_base, phys_addr_end, access
         );
 
-        let mut uframe_buf: Vec<UserFrame> = Vec::with_capacity(1);
-        for vaddr in (virt_addr_base..virt_addr_end).step_by(mem::PAGE_SIZE) {
-            let vaddr: VirtualAddress = VirtualAddress::new(vaddr);
+        // Phase 1: Batch-allocate pages.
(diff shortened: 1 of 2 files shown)


ppenna:
Supersede by #2922

PR #2046, opened 2026-04-14
[kernel][mm] F: platform-isolated FrameAllocator and Kpool over SparseBitmap

danbugs:
## Summary

Reworks the kernel's frame allocator and kernel page pool around [`SparseBitmap`](https://github.com/nanvix/nanvix/pull/2033) (stacked on #2033 — merge that first) and moves their inner state into per-platform modules under `hal/platform/{microvm,hyperlight}/`, so `mm/phys/` has no cfg gates.

## Key changes

**Dependency on `sparse-bitmap`** — `src/kernel/Cargo.toml` adds the workspace dep; `src/kernel/src/collections/mod.rs` re-exports it alongside `bitmap`, `raw-array`, and `slab`.

**Platform-isolated `Inner` types** — state that used to live inline in `mm/phys/{frame,kpool}.rs` moves into:
  - `hal/platform/microvm/{frame,kpool}.rs` — the real implementation (single-chunk `SparseBitmap` at offset 0 over the dense identity / pool range; the BSS-backed bitmap storage moves here too).
  - `hal/platform/hyperlight/{frame,kpool}.rs` — byte-identical mirror of microvm for now, flagged as temporary scaffolding. Exists so the kernel compiles with `--features hyperlight`; replaced wholesale with a scratch-backed implementation when Nanvix-on-Hyperlight (CoW snapshot/restore) lands.

**Facade shape in `mm/phys/`** — `Kpool`, `KernelFrame`, `FrameAllocator` keep their public shape:
  - `FrameAllocator` owns its `Inner` directly (single-owner: the upool consumes it at init).
  - `Kpool` wraps `Rc<RefCell<Inner>>` so `KernelFrame::Drop` can safely dispatch a free through the shared `RefCell` — the guard prevents double-free during nested-drop scenarios (dropping a kframe while someone else holds the kpool borrow causes `borrow_mut` to error out and the bit to leak, instead of corrupting state).
  - `SparseBitmap::set_or_add_chunk` is used where foreign-address registration grows chunks on demand.

## Why this shape

One `SparseBitmap` abstraction, one set of `mm/phys/` facades, two platform `Inner`s that will encode genuinely different allocation strategies (dense bitmap on microvm; scratch bump on Hyperlight). No cfg in `mm/phys/`.

## Stacking notes

This PR is stacked on #2033. Please merge the library PR first; this PR will then rebase cleanly onto `dev`.


danbugs:
Closing per our discussions of changing the abstraction level here.

PR #2048, opened 2026-04-14
[ci] E: Add bench tuning and perf gate

esaurez:
## Summary

This PR improves the reliability and correctness of Windows benchmark CI by addressing two problems:

1. **High performance variability** on self-hosted Windows runners -- benchmark results had extreme tail latency (cold-start p99 was ~400ms vs ~57ms p50) due to background OS interference.
2. **No performance regression gating** -- benchmark regressions could silently land since results were only posted as PR comments without blocking merges.

### Benchmark System Tuning (self-hosted `windows-benchmark` job)

- Added `bench-setup.ps1 -SaveState` step before benchmarks to apply:
  - High Performance power plan with CPU locked at 100%
  - Windows Defender exclusion for the repo directory
  - Stops noisy background services (WSearch, DiagTrack, SysMain, TabletInputService)
- Added `bench-teardown.ps1 -KeepDefenderExclusion` as an always-run cleanup step to restore original system state.
- **Impact measured on both CI servers:**
  - cold-start p50-to-p99 spread: **654% to 5.1%** (S1), **671% to 5.6%** (S2)
  - boot-time p50-to-p99 spread: **9.1% to 2.4%** (S1)
  - In absolute terms, cold-start tail jitter reduced from **~350-380ms to ~3ms**

### Performance Regression Gate

- Added `ci-gate` subcommand to `scripts/benchmark.py` that compares p50 values of current benchmark results against dev baselines.
- Gate **fails the workflow** if any benchmark regresses more than **50% on p50**.
- Applied to both `windows-benchmark-finalize` (self-hosted) and `windows-benchmark-ci-finalize` (GitHub-hosted) jobs.
- Gate runs **after** the benchmark report step, so PR comments are always posted even on regression.
- Gate is **skipped on dev branch** to avoid blocking baseline persistence.
- VFS benchmark is excluded from gating (its variability is guest-internal, not host noise).
- Gracefully handles missing baselines and malformed CSVs.

## Testing

- Verified `ci-gate` exits 0 on no regression, exits 1 when threshold exceeded
- Verified graceful handling of missing baselines (skip with warning)
- Verified graceful handling of malformed CSV values (skip with warning)
- Existing benchmark-related tests pass (5/5)
- Benchmark tuning evaluated on both CI servers with before/after measurements

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1348,6 +1348,13 @@ jobs:
         run: |
           .\z.ps1 build -- all MACHINE=$env:MACHINE_TYPE RELEASE=yes LOG_LEVEL=panic PROFILER=yes
 
+      - name: "Benchmark Setup"
+        # Requires admin privileges. The self-hosted benchmark runners are
+        # configured to run the Actions runner service as Administrator.
+        shell: pwsh
+        run: |
+          .\scripts\bench\bench-setup.ps1 -SaveState
+
       - name: "Run Micro-Benchmarks"
         shell: pwsh
         env:
@@ -1369,6 +1376,12 @@ jobs:
               exit $LASTEXITCODE
             }
           }
+
+      - name: "Benchmark Teardown"
+        if: always()
+        shell: pwsh
+        run: |
+          .\scripts\bench\bench-teardown.ps1 -KeepDefenderExclusion
 
       - name: "Upload Benchmark Results"
         if: always() && !cancelled()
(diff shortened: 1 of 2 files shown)


ppenna:
This PR has been decoupled into two smaller, independent PRs:

- **#2055** — CI performance regression gate (`ci-gate` subcommand + workflow steps). Updated to use a **moving average of the last 10 baseline p50 values** with a **20% regression threshold**.
- **#2056** — Windows benchmark system tuning (`bench-setup.ps1` / `bench-teardown.ps1` wired into CI).

Closing this in favor of those two PRs.

ppenna:
@esaurez splitting this PR because:
- Proposed changes are independent of each other.
- Performance tuning for Windows system is not working and needs extra debugging.

PR #2055, opened 2026-04-15
[ci] Add performance regression gate

ppenna:
## Summary

Decoupled from #2048. Adds a CI performance regression gate that compares current benchmark p50 values against the **moving average of the last 10 baseline data points**. Fails the workflow if any benchmark regresses more than **40%** (configurable).

### `scripts/benchmark.py`
- New `ci-gate` subcommand with `--regression-threshold` (default: 20%) and `--baseline-window` (default: 10).
- New helper `_read_baseline_moving_avg()` reads the tail of baseline CSVs and averages p50 values.
- Main dispatch now propagates non-zero exit codes via `sys.exit()`.

### `.github/workflows/ci.yml`
- Added "Gate on Performance Regressions" step to all 4 finalize jobs:
  - `linux-benchmark-finalize` (baremetal)
  - `windows-benchmark-finalize` (windows-baremetal)
  - `linux-benchmark-ci-finalize` (github-hosted)
  - `windows-benchmark-ci-finalize` (windows-github-hosted)
- Gate is skipped on `dev` branch to avoid blocking baseline persistence.
- Excluded benchmarks: vfs-bench, echo-breakdown (different CSV schema), round-trip-latency (multi-row).

## Testing
- `ci-gate` smoke-tested against real baseline data in `data/baremetal/`.
- All 130 pytest tests pass.
- All 6 pre-commit configurations pass.

### DIFF
--- a/scripts/benchmark.py
+++ b/scripts/benchmark.py
@@ -6,15 +6,18 @@
 # ======================================================================
 
 import argparse
+import collections
 import csv
 import io
 import itertools
+import math
 import os
 import pathlib
 import platform
 import re
 import signal
 import subprocess
+import sys
 from typing import Optional
 
 # ======================================================================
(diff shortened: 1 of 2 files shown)


scripts/benchmark.py
ppenna, line 1: Like `regression-threshold`, `window` for moving average should be a CLI argument passed in by `ci.yml`

PR #2084, opened 2026-04-17
[uservm] Implement two-phase sandbox lifecycle

### LINKED ISSUE #2087: [uservm] Migrate Hyperlight/Nanvix integration to evolve sandbox lifecycle
## Summary

Adopt Hyperlight's two-phase `evolve()` → `call("kmain")` sandbox lifecycle for the Nanvix guest kernel, replacing the single-phase boot path.

## Motivation

The current Hyperlight integration boots the kernel in a single `evolve()` call that runs straight through to `kmain`. Hyperlight's intended design separates sandbox initialization (`UninitializedSandbox` → `evolve()` → `MultiUseSandbox`) from guest function dispatch (`sandbox.call()`). Aligning with this protocol:

- Enables future `sandbox.call()` re-entry for hot functions without rebooting the kernel.
- Makes the kernel compatible with Hyperlight's snapshot/restore and multi-use sandbox features.
- Moves expensive one-time initialization (heap, PEB handle) to the evolve phase so it is not repeated on each call.

### Evolve phase (`hyperlight_pre_kmain`)
1. Initialize the kernel heap early (needed for `FunctionCall` deserialization).
2. Compute the PEB base address and store a `GuestHandle` for reuse.
3. Halt the VM via `VmAction::Halt` with the `_nanvix_dispatch` address in EAX, causing `evolve()` to return a `MultiUseSandbox`.

### Call phase (`_nanvix_dispatch` → `nanvix_dispatch_function`)
1. Assembly stub restores the boot stack and calls the Rust dispatcher.
2. Deserialize `FunctionCall` from the PEB input buffer.
3. Match on `function_name` and dispatch (`"kmain"` → `crate::kmain()`).
4. Return `GuestError` via PEB output buffer for unrecognized functions.

### Host side (UserVM)
- Split `run()` into `evolve()` + `sandbox.call("kmain", ())`.
- Handle `GuestAborted` errors from both phases.

## Acceptance criteria

- [ ] Kernel boots via `evolve()` then enters `kmain` via `sandbox.call("kmain", ())`.
- [ ] `kheap::init()` runs once during evolve, skipped in `kmain`.
- [ ] Unrecognized function names produce a `GuestError` response.
- [ ] All standalone Windows integration tests pass.
- [ ] All microvm Linux integration tests pass (no regression).

ppenna:
Backport the Hyperlight two-phase boot from `feature-hyperlight-stable`. Instead of running `kmain` directly during `evolve()`, the kernel now halts the VM before `kmain` to let the host transition `UninitializedSandbox` into `MultiUseSandbox`, then re-enters via `sandbox.call("run", ())` to run `kmain` for real.  ## Kernel-side changes  - Add `_nanvix_dispatch` `global_asm!` entry point that restores the boot stack and calls `kmain` after the host issues `call("run", ())`. - Add `halt_for_evolve()` to halt the VM during evolve with the dispatch address in EAX (port `0x6c`). ESP is 16-byte aligned before the halt to satisfy Hyperlight's post-initialise register validation. - Add `hyperlight_pre_kmain()`, called from `_do_start2` via a conditional `.if` directive, so that `kmain` remains completely platform-agnostic. - Add `HYPERLIGHT_EVOLVE` const to `start.rs` to gate the pre-kmain call at assembly level without touching the `kmain` code path.  ## Host-side changes (uservm)  - Split `Vmm::run()` into two phases: `evolve()` returns a `MultiUseSandbox`, then `call("run", ())` re-enters the guest at `_nanvix_dispatch`. - Handle evolve-phase errors by extracting `GuestAborted` from the nested `HyperlightVmError(Initialize(...))` error chain. - Treat "Stack pointer is out of bounds" after `call()` as a clean exit, since the dispatch handler halts via port `0x6c` rather than using Hyperlight's guest function return convention.  ## Rename  - `signal_boot_complete` ΓåÆ `signal_startup_complete` on both microvm and hyperlight platforms for consistency.  ## Testing  Build (`MACHINE=hyperlight`) and all standalone Windows integration tests pass: `empty`, `arch-rust`, `echo-rust-nostd`, `thread-rust`, `stress-rust`, `linux-app`.
Closes #2087

### DIFF
--- a/src/kernel/src/hal/arch/x86/asm/start.rs
+++ b/src/kernel/src/hal/arch/x86/asm/start.rs
@@ -22,6 +22,17 @@ use ::core::arch::global_asm;
 /// static data.
 const CLEAR_BSS: u32 = if cfg!(feature = "microvm") { 0 } else { 1 };
 
+/// Whether to run the Hyperlight evolve phase before calling `kmain`.
+///
+/// On Hyperlight, the bootstrap code must halt the VM during `evolve()` before entering `kmain()`.
+/// The host then calls `sandbox.call("kmain", ())` which re-enters the guest at `_nanvix_dispatch`,
+/// restores the boot stack, and calls `kmain()` for real.
+///
+/// Evaluates to 0 on all other platforms; the assembler `.if` directive eliminates the
+/// `call hyperlight_pre_kmain` instruction entirely at compile time.
+///
+const HYPERLIGHT_EVOLVE: u32 = if cfg!(feature = "hyperlight") { 1 } else { 0 };
+
 //==================================================================================================
 // Bootstrap Section — BSP Entry Point
 //==================================================================================================
@@ -85,6 +96,13 @@ global_asm!(
     "    xorl %edi, %edi",
 
     // Call kernel main function.
+    //
+    // On Hyperlight, the evolve phase must halt the VM before kmain runs. `hyperlight_pre_kmain()`
+    // registers the dispatch entry point and halts; it never returns. The host then calls
+    // sandbox.call("kmain", ()) which enters `_nanvix_dispatch` → `kmain``.
+    ".if {HYPERLIGHT_EVOLVE}",
+    "    call hyperlight_pre_kmain",
+    ".endif",
     "    push %esp",
     "    call kmain",
     "    addl $4, %esp",
@@ -97,6 +115,7 @@ global_asm!(
     "    jmp 1b",
 
     CLEAR_BSS = const CLEAR_BSS,
+    HYPERLIGHT_EVOLVE = const HYPERLIGHT_EVOLVE,
     PAGE_SIZE = const PAGE_SIZE,
     CONTEXT_HW_SIZE = const ContextInformation::CONTEXT_HW_SIZE,
     KSTACK_GUARD_PATTERN = const constants::KSTACK_GUARD_PATTERN,
--- a/src/kernel/src/hal/platform/hyperlight/mod.rs
+++ b/src/kernel/src/hal/platform/hyperlight/mod.rs
@@ -49,6 +49,7 @@ use crate::{
 };
 use ::alloc::{
     collections::linked_list::LinkedList,
+    format,
     string::{
         String,
         ToString,
@@ -65,12 +66,22 @@ use ::config::hyperlight::{
     PEB_SIZE,
 };
 use ::hyperlight_common::{
+    flatbuffer_wrappers::{
+        function_call::FunctionCall,
+        function_types::FunctionCallResult,
+        guest_error::{
+            ErrorCode as GuestErrorCode,
+            GuestError,
+        },
+    },
     layout::MAX_GVA,
     mem::{
         FileMappingInfo,
         HyperlightPEB,
     },
+    outb::VmAction,
 };
+use ::hyperlight_guest::guest_handle::handle::GuestHandle;
 use ::sys::{
     config::memory_layout,
     error::{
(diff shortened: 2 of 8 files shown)


src/kernel/src/hal/arch/x86/asm/start.rs
```diff
@@ -22,6 +22,13 @@ use ::core::arch::global_asm;
 /// static data.
 const CLEAR_BSS: u32 = if cfg!(feature = "microvm") { 0 } else { 1 };
 
+/// Whether to run the Hyperlight evolve phase before calling `kmain`.
+///
+/// On Hyperlight, the bootstrap code must halt the VM during `evolve()` before entering
+/// `kmain()`. The host then calls `sandbox.call("run", ())` which re-enters the guest at
+/// `_nanvix_dispatch`, restores the boot stack, and calls `kmain()` for real.
+const HYPERLIGHT_EVOLVE: u32 = if cfg!(feature = "hyperlight") { 1 } else { 0 };
```
danbugs, line 34: Looks like this isn't hyperlight specific and so is compiled in the microvm too?

src/kernel/src/kmain.rs
```diff
@@ -272,6 +272,7 @@ pub extern "C" fn kmain(kargs: &KernelArguments) {
         IoMemoryAllocator,
         LinkedList<KernelModule>,
     );
+    #[allow(unused_mut)]
```
danbugs: Might be worth a cleanup?

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -457,8 +457,53 @@ impl Vmm {
...
+                    .blocking_lock()
+                    .control_tx
+                    .blocking_send(VcpuControlResponse::Shutdown)
+                {
+                    error!("run(): failed to notify vmm thread (error={e:?})");
+                }
+
+                // During evolve(), GuestAborted is wrapped in
+                // HyperlightVmError(Initialize(...)) so extract it from the error chain.
+                return match Self::extract_guest_abort(&error) {
```
danbugs, line 481: Instead of using `abort` for hyperlight completion, you can now use `halt`, which was added w/ the upstreaming of the hw-interrupts feature. Consider swapping so you don't have to do this "extract_guest_abort" logic?

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -359,10 +391,70 @@ pub(in crate::hal::platform) fn do_shutdown(status: usize) -> ! {
...
 ///
-/// Signals the VMM that the kernel has finished booting and user-space
-/// applications are about to start. No-op on the Hyperlight platform.
+/// Halts the VM during the evolve phase by writing the `_nanvix_dispatch`
+/// address to EAX and issuing a `VmAction::Halt` (port 0x6c). This causes
+/// `evolve()` to return on the host with a `MultiUseSandbox`. The VMM then
+/// calls `sandbox.call("run", ())`, which re-enters the guest at
+/// `_nanvix_dispatch`.
+///
+fn halt_for_evolve() {
```
danbugs: Need to at least make this function fn halt_for_evolve() -> !. Though, it's worth noting that we used to have a similar function in HL, but we kept running into issues because it's a bit of a mess to have a Rust function whose stack epilogue never runs in a situation where we can reenter the same execution context. This doesn't really work, and we ended up having to move all of the places that can do hlt on x64 into arch-specific assembly stubs that handle stack frame creation and teardown properly. Reference: https://github.com/hyperlight-dev/hyperlight/blob/2b803de0d03bc09d79aef4d324552ad339cd7786/src/hyperlight_guest_bin/src/arch/amd64/init.rs#L160-L184

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -359,10 +391,70 @@ pub(in crate::hal::platform) fn do_shutdown(status: usize) -> ! {
...
+            "and esp, 0xFFFFFFF0",
+            "mov eax, {0:e}",
+            "mov dx, 0x6c",
+            "out dx, al",
+            in(reg) dispatch_addr,
+            out("eax") _,
+            out("edx") _,
+            options(nomem),
+        );
+    }
```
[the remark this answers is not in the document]
danbugs: +1 for asm!(..., options(noreturn))

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -84,6 +84,38 @@ use ::sys::{
...
+    ".code32",
+    ".extern kstack",
+    ".extern kmain",
+    ".globl _nanvix_dispatch",
+    "_nanvix_dispatch:",
+    // Restore the boot stack. The KernelArguments (magic, info) are still
+    // at kstack-8 from _do_start2's pushes during the evolve phase.
+    "    movl $kstack, %esp",
+    "    movl %esp, %ebp",
+    "    subl $8, %esp",
```
danbugs, line 141: Just flagging that this will no longer be true after adding snapshot/restore because the ktack is CoWed.

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -359,10 +391,70 @@ pub(in crate::hal::platform) fn do_shutdown(status: usize) -> ! {
...
+///
+/// Registers the dispatch function (`_nanvix_dispatch`) with Hyperlight by
+/// halting the VM via port 0x6c with the dispatch address in EAX. This function
+/// never returns — the VM is halted and `evolve()` completes on the host side.
+///
+/// The host subsequently calls `sandbox.call("run", ())`, which enters the guest
+/// at `_nanvix_dispatch`, restores the boot stack, and calls `kmain()`.
+///
+#[unsafe(no_mangle)]
+pub extern "C" fn hyperlight_pre_kmain() -> ! {
```
danbugs: Doesn't have to be `pub extern`, I think?

src/uservm/src/vmm/hyperlight/mod.rs
ppenna: This code path will be triggered only by Hyperlight runtime. Not sure what is the action here.

src/uservm/src/vmm/hyperlight/mod.rs
```diff
@@ -501,6 +549,24 @@ impl Vmm {
...
+    /// read fails with "Stack pointer is out of bounds" (an `AnyhowError` from `shared_mem.rs`).
+    ///
+    /// We detect this by checking the `Display` representation, which is the stable interface for
+    /// `thiserror`-derived errors. The `AnyhowError` variant's display format is `"Anyhow Error was
+    /// returned: <inner>"`.
+    ///
+    fn is_halt_exit(error: &HyperlightError) -> bool {
+        error.to_string().contains("Stack pointer is out of bounds")
+    }
+
```
ppenna, line 575: @danbugs I am unsure about this.

danbugs: #2088 tracking

Issue #2093, opened 2026-04-18
[dlfcn] Circular dependency detection uses mutex lock state instead of visited set


esaurez:
## Summary

Symbol resolution detects circular dependencies by checking `dlfile.is_locked()` (the spinlock's internal state). This conflates ""currently being resolved by this call stack"" with ""already visited in this specific lookup chain"" and can produce false positives on legal diamond-shaped dependency graphs.

## Root Cause

```rust
// src/libs/syscall/src/dlfcn/syscall/dynlib.rs:480-488
if dlfile.is_locked() {
    let reason: &str = ""circular dependency detected"";
    return Err(Error::new(ErrorCode::BadFile, reason));
}
let dlfile: MutexGuard<'_, DynamicLibrary> = dlfile.lock();
```

The recursive `lookup()` function locks each library's mutex during resolution. If a diamond dependency (`A -> B, A -> C, B -> D, C -> D`) causes `D` to be visited via two paths, the second visit sees `D`'s mutex as locked and incorrectly reports a cycle.

## Reproduction Scenario

```
A depends on B and C.
B depends on D.
C depends on D.
Both B and C export symbol "foo", but only D defines it.

dlsym(A_handle, "foo"):
  lookup in A -> not found
  lookup in B (locks B) -> not found in B
    lookup in D (locks D) -> found, returns
  lookup in C (locks C) -> not found in C
    lookup in D -> D.is_locked() == true -> ERROR "circular dependency"
```

## Expected Behavior

Use a per-lookup visited set (e.g., `HashSet<DlHandle>`) passed through the recursive call chain, not mutex lock state. A library should only be considered circular if it appears in the current lookup's ancestor chain.

## Linux Reference

glibc's [`elf/dl-lookup.c`](https://github.com/bminor/glibc/blob/master/elf/dl-lookup.c) (`do_lookup_x`) avoids this entirely by using a flat BFS iteration over pre-built `r_scope_elem.r_list[]` arrays. The scope lists are acyclic by construction, so no runtime cycle detection is needed during lookup.

## Affected Files

- `src/libs/syscall/src/dlfcn/syscall/dynlib.rs:480-488` — `is_locked()` cycle detection in `lookup()`

esaurez:
[...]

PR #2129, opened 2026-04-20
[dlfcn] Fix dlsym to search full dependency tree for all symbols

### LINKED ISSUE #2127: [dlfcn] dlsym does not search dependency tree for symbols absent from root library's dynsym
## Summary

`dlsym(handle, symbol)` fails to find symbols exported by a library's `DT_NEEDED` dependencies when the symbol is not referenced in the root library's own `.dynsym` table. This breaks the POSIX contract for `dlsym`.

## Root Cause

The `lookup()` method in `DynamicLibrary` only searches the dependency tree if `find()` returns an **undefined** entry for the symbol in the root library's `.dynsym`. If the symbol is not in `.dynsym` at all (i.e., the root library never references it), `find()` returns `None` and the dependency tree is never searched.

```rust
// src/libs/syscall/src/dlfcn/syscall/dynlib.rs (current code)
if let Some(symbol) = self.find(symbol_name) {
    if symbol.is_undefined() {
        // Only searches dependencies if symbol is in dynsym as undefined.
        for (_dlname, dlfile) in self.dependencies.iter() { ... }
    } else {
        return Ok(Some((...)));
    }
}
// If find() returns None -> dependencies are NEVER searched.
Ok(None)
```

## Reproduction Scenario

```c
// libprovider.so exports: provider_add(), provider_mul(), helper_func()
// libconsumer.so has DT_NEEDED: libprovider.so
// libconsumer.so calls provider_add() (referenced in dynsym)
// libconsumer.so does NOT call helper_func() (not in dynsym)

void *h = dlopen("lib/libconsumer.so", RTLD_NOW);  // succeeds

dlsym(h, "provider_add");   // works: symbol is undefined in consumer's dynsym
dlsym(h, "helper_func");    // FAILS: symbol not in consumer's dynsym at all
                             //        -> dependencies never searched
```

This also affects `ctypes.CDLL` in Python, which uses `dlsym` to look up arbitrary symbols from a library and its dependencies.

## Expected Behavior

Per POSIX ([dlsym specification](https://pubs.opengroup.org/onlinepubs/9799919799/functions/dlsym.html)):

> If handle is a handle returned by `dlopen()`, the `dlsym()` function shall search for the named symbol in the **specified object and in all objects loaded as a result of loading that object** (that is, the DT_NEEDED entries of that object and of all such dependent objects, recursively).

The search should cover the object itself **and its entire dependency tree**, regardless of whether the symbol appears in the root object's `.dynsym`.

## Linux Reference

glibc's `dlsym` implementation in [`elf/dl-lookup.c`](https://github.com/bminor/glibc/blob/master/elf/dl-lookup.c) (`do_lookup_x`) walks the object's `l_searchlist` — a pre-built BFS list of the object and all its dependencies. It does not gate the search on whether the root object references the symbol.

## Affected Files

- `src/libs/syscall/src/dlfcn/syscall/dynlib.rs` — `lookup()` method

esaurez:
## Summary

Restructure `lookup()` to always search self -> dependencies -> global scope, even when the symbol is not present in the root library's dynsym table.

Fixes #2127.

## Root Cause

Previously, `lookup()` only searched the dependency chain if `find()` found the symbol as **undefined** in the root library's dynsym. If the symbol was not in dynsym at all (e.g., `dlsym(extension_handle, "malloc")` where the extension never referenced `malloc`), the dependency tree was never searched and `dlsym` would incorrectly fail.

Per [POSIX dlsym](https://pubs.opengroup.org/onlinepubs/9799919799/functions/dlsym.html), `dlsym(handle, sym)` must search the object **and all objects loaded as a result of loading that object** (the DT_NEEDED dependency tree), regardless of whether the root object references the symbol.

## Fix

The `lookup()` method now:

1. Checks if the symbol is **defined** locally -> returns it.
2. Always searches the full dependency tree (regardless of dynsym content).
3. Falls back to the global symbol table (needed for relocation resolution).

Step 3 is a pre-existing behavior needed for relocation resolution but not strictly correct for `dlsym(handle, ...)`. Separating the two lookup paths is tracked as a follow-up.

## Linux Reference

glibc's `dlsym` in [`elf/dl-sym.c`](https://github.com/bminor/glibc/blob/master/elf/dl-sym.c) passes `match->l_scope` (the object's load group) to `_dl_lookup_symbol_x` in [`elf/dl-lookup.c`](https://github.com/bminor/glibc/blob/master/elf/dl-lookup.c), which iterates the `l_searchlist` (BFS-ordered list of the object + all DT_NEEDED deps). The search is not gated on whether the root object references the symbol.

## Affected Files

- `src/libs/syscall/src/dlfcn/syscall/dynlib.rs` — `lookup()` restructured.

## Testing

All existing integration tests pass (both HTTP and terminal executors).

### DIFF
--- a/src/libs/syscall/src/dlfcn/syscall/dynlib.rs
+++ b/src/libs/syscall/src/dlfcn/syscall/dynlib.rs
@@ -457,42 +457,66 @@ impl DynamicLibrary {
     }
 
     /// Looks up a symbol in the dynamic library.
+    ///
+    /// Search order:
+    /// 1. The library itself (defined symbols).
+    /// 2. The library's DT_NEEDED dependency tree (recursive, no global fallback).
+    /// 3. The global symbol table (main executable symbols).
+    ///
+    /// NOTE: Step 3 is needed for relocation resolution (symbols from the main
+    /// executable). Strictly, POSIX `dlsym(handle, ...)` should only search
+    /// the object's load group (steps 1-2), not the global scope. Separating
+    /// the two lookup paths is tracked in #2130.
     pub fn lookup(&self, symbol_name: &str) -> Result<Option<(usize, usize)>, Error> {
         ::syslog::trace!("lookup(): symbol={}, dlname={:?}", symbol_name, self.filename);
 
+        // Search self and dependency tree without global fallback.
+        if let Some(result) = self.lookup_in_load_group(symbol_name)? {
+            return Ok(Some(result));
+        }
+
+        // Fall back to the global symbol table (symbols from the main
+        // executable, registered via --export-dynamic). This fallback is
+        // performed only once at the top level, not during recursive
+        // dependency traversal.
+        if let Some(addr) = super::global_symbol_lookup(symbol_name) {
+            // Global symbols are absolute addresses, so base is 0.
+            return Ok(Some((0, addr)));
+        }
+
+        Ok(None)
+    }
+
+    /// Searches for a symbol in this library and its dependency tree only.
+    ///
+    /// Does NOT fall back to the global symbol table. This ensures that
+    /// recursive dependency searches do not short-circuit to the global scope
+    /// before the entire dependency tree has been checked.
+    fn lookup_in_load_group(&self, symbol_name: &str) -> Result<Option<(usize, usize)>, Error> {
         if let Some(symbol) = self.find(symbol_name) {
-            // Check if symbol is defined in the library or in a dependency.
-            if symbol.is_undefined() {
-                // Symbol is defined in a dependency, search dependencies.
-                for (_dlname, dlfile) in self.dependencies.iter() {
-                    if let Some(dlfile) = dlfile {
-                        // Check if dependency is locked.
-                        if dlfile.is_locked() {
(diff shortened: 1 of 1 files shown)


esaurez:
Addressed the review comments:

**Comment 1 (global fallback in recursion):** Fixed. Split `lookup()` into two methods:
- `lookup()` — public API: searches self + deps + global fallback (once at top level).
- `lookup_in_load_group()` — private: searches self + deps only (used for recursive traversal).

This ensures the global fallback fires exactly once after the full dependency tree is exhausted, not during each recursive step.

**Comment 2 (is_locked cycle detection):** Pre-existing behavior, tracked in #2093. Nanvix is single-threaded (spinlocks for reentrancy), so the concurrency scenario doesn't apply today.

**Comment 3 (doc comment says RTLD_GLOBAL):** Fixed. Updated to say ""main executable symbols"" which matches the current state of this branch. The RTLD_GLOBAL extension is a separate PR.

**Comment 4 (self-referential #2127):** Fixed. Removed the specific issue number, says ""follow-up change"" instead.

ppenna (dismissed):
LGTM

PR #2151, opened 2026-04-22
[uservm] Host kernel stack correlation via ETW (Windows)

esaurez:
## Summary

Adds host kernel stack correlation via ETW/WPR for the Nanvix guest flamegraph profiler on Windows, producing unified flamegraphs with `[GUEST]` and `[HOST]` root frames that span from guest user-space code through the guest kernel, host VMM (nanvixd), and host OS kernel (vid.sys, ntoskrnl). Also introduces a cross-platform profiler abstraction (`HostKernelSession` type alias + no-op stub) and timestamp infrastructure so the follow-up Linux PR can plug in `perf record`-based host profiling without changing `lib.rs`.

### Host kernel tracing (ETW)
- **etw.rs**: `EtwSession` manages WPR start/stop lifecycle. Automatically starts an ETW CPU sampling session when nanvixd runs as admin, stops and saves the ETL trace on exit. Sampling frequency is configurable via `NANVIX_PROFILER_FREQ_HZ` (default 1kHz, clamped to 1-10kHz). xperf path is overridable via `NANVIX_XPERF_PATH` env var.

### Cross-platform profiler abstraction
- **mod.rs**: Re-exports `EtwSession` as `HostKernelSession` so `lib.rs` uses a single code path with zero `#[cfg(target_os)]` blocks in the profiler lifecycle.
- **host_session_stub.rs**: No-op stub for non-Windows platforms so the code compiles on Linux/KVM without the ETW implementation. A follow-up Linux PR will replace the stub with a `perf record`-based implementation.

### Timestamp infrastructure (samples.rs)
- `timestamp_now()` / `timestamp_frequency()`: Cross-platform high-resolution timestamps (QPC on Windows, CLOCK_MONOTONIC_RAW on Linux) for future host trace correlation.
- `qpc_timestamp` field on `StackSample`: Records when each guest sample was captured.
- `write_folded_from_samples()`: Allows draining samples once and using them for multiple outputs (avoids double-drain bug).

### Build system
- `release-profiling` Cargo profile: inherits release but keeps symbols (`strip=false`, `debug=line-tables-only`).
- `kernel.mk`: Automated `objcopy --strip-debug` removes .debug_* sections while keeping .symtab (~60KB overhead vs 1.4MB).
- `nanvixd.mk`: Copies PDB alongside exe on Windows.

### E2E scripts
[...]

PR #2168, opened 2026-04-22
[uservm] Linux KVM guest sampling and perf host correlation

esaurez:
## Summary

Adds Linux KVM guest sampling and `perf record`-based host kernel stack correlation, completing the cross-platform profiler started in PR #2151 (Windows ETW). Plugs into the `HostKernelSession` type alias introduced there -- `lib.rs` has zero changes in this PR.

### KVM guest sampling (`vmm/microvm/mod.rs`)
- Timer thread sends `SIGUSR2` to the vCPU thread at configurable frequency (`NANVIX_PROFILER_FREQ_HZ`, clamped 1-10kHz)
- Uses `SIGUSR2` (not `SIGUSR1`) because `SIGUSR1` is used by the orchestrator for shutdown and sets the `SHUTDOWN` flag
- Separate no-op `profiler_signal_handler` installed alongside the existing shutdown handler
- On `Interrupted` exit: reads `KVM_GET_REGS` + `KVM_GET_SREGS` for EIP/EBP/CR3, walks frame-pointer chain
- When profiler is active, `Interrupted` exits continue the VM instead of shutting down (checks `SHUTDOWN` flag to distinguish real shutdown from profiler tick)
- Both `get_regs` and `get_sregs` must succeed for a valid sample -- skips if either fails

### Host kernel tracing (`perf_linux.rs`)
- `PerfSession` wrapping `perf record` with system-wide sampling (`-a`) when running as root for kernel stack visibility (kvm_intel, ioctl paths)
- Falls back to per-PID recording when unprivileged (no kernel stacks but still captures VMM userspace)
- Waits for `perf_event` file descriptors in `/proc/PID/fd` before returning from `start()` -- ensures perf is actually sampling before the workload begins
- Detects early `perf` exit via `try_wait()` and returns `Err` (not silent success)
- `chmod 644` on `perf.data` after stop so the E2E script can read it without root
- Configurable frequency via `NANVIX_PROFILER_FREQ_HZ`

### Platform abstraction (`mod.rs`)
- Re-exports `PerfSession` as `HostKernelSession` for Linux, replacing the no-op stub from the Windows PR
- Stub remains for truly unsupported platforms (`not(windows OR linux)`)

### E2E script (`full-flamegraph.sh`)
- Mirrors `full-flamegraph.ps1` structure: builds ramfs, runs nanvixd, extracts host stacks, merges `[GUEST]`/`[HOST]`, generates SVG
- Uses `perf script -F comm,pid,tid,time,event,ip,sym,dso` to omit the period field -- each perf event collapses to count=1, matching guest profiler's 1-per-sample weighting
- Filters collapsed stacks by `grep -iE "nanvixd|uservm"` (more robust than `--comms` which truncates at 15 chars)
- Requires `inferno` and `rustfilt` (`cargo install inferno rustfilt`)

### Ungated register access (`kvm/vcpu/mod.rs`)
- Removed `#[cfg(feature = "gdb")]` from `get_regs()` and `get_sregs()` -- the profiler needs these on every `Interrupted` exit, not just during GDB sessions

## How to use

```bash
# Build
python3 z.py build --profile --release -- DEPLOYMENT_MODE=standalone

# Run as root for full E2E (guest + host kernel stacks)
sudo -E env "PATH=$PATH" scripts/bench/full-flamegraph.sh \
    --guest-elf bin/python.elf \
    --user-symbols /path/to/python.elf.sym
```

## Known limitations

- **No timestamp-based correlation**: Same as the Windows PR -- guest and host stacks are merged by appending, not temporally correlated
- **System-wide perf noise**: When running with `-a`, other processes' stacks are captured but filtered out by the `nanvixd|uservm` grep. Very busy systems may have slightly higher overhead.
- **WSL2**: `perf record` may not work on WSL2 without WSL2-specific kernel tools

## Testing

Tested on Azure Linux 3.0 (KVM with nested virtualization). Produces balanced `[GUEST]`/`[HOST]` flamegraphs with 100% symbol resolution across Nanvix kernel, CPython, nanvixd VMM, and host Linux kernel (futex, vfs, ext4, kvm).

## Stacked on

- PR #2151: [uservm] E: Host kernel stack correlation via ETW (Windows)

### DIFF
--- a/src/uservm/src/guest_profiler/perf_linux.rs
+++ b/src/uservm/src/guest_profiler/perf_linux.rs
@@ -1,0 +1,286 @@
+// Copyright(c) The Maintainers of Nanvix.
+// Licensed under the MIT License.
+
+//! `perf`-based host kernel stack capture for guest flamegraphs (Linux).
+//!
+//! This is the Linux equivalent of the Windows ETW module (`etw.rs`).
+//! It uses `perf record` to capture CPU sampling with kernel stacks
+//! while the guest profiler runs, producing `perf.data` for later
+//! host/guest stack merging.
+//!
+//! # How it works
+//!
+//! 1. **Before VM run**: Start `perf record` with CPU sampling and
+//!    kernel+user stacks on the nanvixd process.
+//! 2. **During VM run**: Our profiler records guest samples while
+//!    `perf record` runs concurrently on the host.
+//! 3. **After VM exit**: Stop `perf record`, convert to folded stacks
+//!    with `perf script | inferno-collapse-perf`, filter by nanvixd.
+//!
+//! # Sampling rate alignment
+//!
+//! Both our profiler and `perf record` sample at ~1kHz for broadly
+//! comparable analysis. Timestamp-based correlation is not implemented;
+//! host and guest stacks are merged by stack identity, not by time.
+
+use std::{
+    os::unix::fs::PermissionsExt,
+    process::{
+        Child,
+        Command,
+        Stdio,
+    },
+};
+
+use ::libc::c_int;
+
+//==================================================================================================
+// Constants
+//==================================================================================================
+
+/// Default profiling frequency in Hz. Controls the `perf record -F` sampling
+/// rate. Matches the guest profiler's default for consistent weighting.
+const DEFAULT_FREQ_HZ: u64 = 1000;
+
+/// Minimum allowed profiling frequency (Hz).
+const MIN_FREQ_HZ: u64 = 1;
+
+/// Maximum allowed profiling frequency (Hz). Values above this can cause
+/// excessive overhead from perf ring buffer processing.
+const MAX_FREQ_HZ: u64 = 10_000;
+
+/// Environment variable controlling the profiling frequency (Hz).
+const PROFILER_FREQ_ENV: &str = "NANVIX_PROFILER_FREQ_HZ";
+
+/// Path to the kernel perf_event_paranoid sysctl. Values <= 0 allow
+/// system-wide profiling without root.
+const PERF_EVENT_PARANOID_PATH: &str = "/proc/sys/kernel/perf_event_paranoid";
+
+/// Identifier string for perf_event file descriptors in /proc/<pid>/fd
+/// symlink targets (e.g. `anon_inode:[perf_event]`).
+const PERF_EVENT_FD_IDENTIFIER: &str = "perf_event";
+
+/// Maximum time (seconds) to wait for perf to open its perf_event fds
+/// before giving up on readiness detection.
+const PERF_READINESS_TIMEOUT_SECS: u64 = 5;
+
+/// File permissions applied to perf.data after recording. perf record
+/// running as root creates the file with mode 0600; we relax it so the
+/// merge script can read it without root.
+const PERF_DATA_PERMISSIONS: u32 = 0o644;
+
+/// Nanoseconds per second — timestamp frequency for CLOCK_MONOTONIC_RAW.
+const NANOS_PER_SECOND: u64 = 1_000_000_000;
+
+/// Manages a `perf record` session for kernel stack sampling.
+pub struct PerfSession {
+    /// Path to the output perf.data file.
+    output_path: String,
+    /// Running `perf record` child process.
+    child: Option<Child>,
+    /// PID of the nanvixd process (for filtering).
+    pid: u32,
+}
+
+impl PerfSession {
+    /// Creates a new perf session that will write to the given path.
+    pub fn new(output_path: &str) -> Self {
+        Self {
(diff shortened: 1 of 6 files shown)


scripts/bench/full-flamegraph.sh
ppenna, line 1: Should we have this as a python script so that it works cross-platform?

src/uservm/src/guest_profiler/perf_linux.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn start(&mut self) -> Result<(), String> {
+        if self.child.is_some() {
+            return Err("perf session already active".to_string());
+        }
+
+        let freq_hz: u64 = std::env::var("NANVIX_PROFILER_FREQ_HZ")
+            .ok()
+            .and_then(|v| v.parse().ok())
+            .unwrap_or(1000)
+            .clamp(1, 10_000);
```
ppenna: We should have constants for these `FALLBACK_*` instead of harcoding values.

src/uservm/src/guest_profiler/perf_linux.rs
```diff
@@ -0,0 +1,222 @@
...
+            self.output_path
+        );
+        Ok(())
+    }
+
+    /// Check if we can use system-wide perf recording (requires root or
+    /// `perf_event_paranoid <= 0`).
+    fn can_system_wide() -> bool {
+        // Check perf_event_paranoid: -1 or 0 allows system-wide without root.
+        if let Ok(val) = std::fs::read_to_string("/proc/sys/kernel/perf_event_paranoid") {
```
ppenna: Should we have a filescope constant for this path?

src/uservm/src/guest_profiler/perf_linux.rs
```diff
@@ -0,0 +1,222 @@
...
+            std::fs::Permissions::from_mode(0o644),
+        );
+
+        eprintln!("PERF_SESSION: saved perf data to {}", self.output_path);
+        Ok(self.output_path.clone())
+    }
+
+    /// Returns the timestamp frequency (nanoseconds for CLOCK_MONOTONIC_RAW).
+    pub fn timestamp_frequency(&self) -> u64 {
+        1_000_000_000
```
ppenna: Why this magic value?

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -596,6 +610,27 @@ impl Vmm {
...
+                sa_flags: 0,
+                sa_restorer: None,
+            };
+            sigaction(PROFILER_SIGNAL, &profiler_action, std::ptr::null_mut())
+        };
+        if ret2 != 0 {
+            let errno: i32 = unsafe { *libc::__errno_location() };
+            error!("error installing profiler signal handler (errno={errno:?})");
+        }
     }
```
ppenna: Extract this to a function.

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -636,6 +671,33 @@ impl Vmm {
...
+        // interrupt KVM_RUN periodically for stack sampling.
+        let profiler_stop = Arc::new(AtomicBool::new(false));
+        let profiler_thread = if self.guest_profiler.is_some() {
+            let stop = profiler_stop.clone();
+            let vcpu_tid = unsafe { libc::pthread_self() };
+            let freq_hz: u64 = std::env::var("NANVIX_PROFILER_FREQ_HZ")
+                .ok()
+                .and_then(|v| v.parse().ok())
+                .unwrap_or(1000)
+                .clamp(1, 10_000); // 1 Hz min (avoid div-by-zero), 10 kHz max (avoid spin).
```
ppenna: See my comment on hardcode values.

src/uservm/src/vmm/microvm/mod.rs
```diff
@@ -725,6 +837,12 @@ impl Vmm {
             }
         };
 
+        // Stop profiler timer.
+        profiler_stop.store(true, Ordering::Relaxed);
+        if let Some(t) = profiler_thread {
+            let _ = t.join();
```
ppenna: Capture error and issue warning in case of failure.

scripts/bench/full-flamegraph.sh
esaurez: Added a python script in the previous PR and extend it in this one

src/uservm/src/guest_profiler/perf_linux.rs
esaurez: Added

src/uservm/src/guest_profiler/perf_linux.rs
esaurez: Added a more descriptive name

PR #2183, opened 2026-04-23
[kernel] Prevent scratch region from overlapping user address space

### LINKED ISSUE #2182: Scratch region overlaps user stack on Hyperlight with 256 MB memory
## Summary

Starting with v0.12.457, Hyperlight builds configured with `MEMORY_SIZE=256` fail to spawn user processes. The scratch region base address is computed using `MAX_GPA` (lowered to `0xFEDF_FFFF` to avoid the KVM APIC access page), which pushes the scratch region into the user stack address range when the memory size is large enough.

## Root Cause

PR #2149 (`fceb2072e1` — "Allocators in scratch region") changed how the scratch region base address is computed:

```
scratch_base = MAX_GPA - scratch_size + 1
```

The pinned Hyperlight dependency (rev `3fd3b7f`) sets `MAX_GPA = 0xFEDF_FFFF` (18 MB below the 32-bit max) to avoid the KVM APIC access page at `0xFEE00000`.

For a 256 MB configuration the scratch region is ~241 MB, so:

```
scratch_base = 0xFEDFFFFF - ~241MB + 1 = 0xEFD1E000
```

The user stack is hardcoded at `[0xEFC00000, 0xF0000000)` (`USER_END_RAW = 0xF0000000`, `USER_STACK_SIZE = 4 MB`).

This creates a **~2.9 MB overlap**:

```
User stack:  [0xEFC00000, 0xF0000000)
Scratch:     [0xEFD1E000, 0xFEE00000)
                  ^^^^^^^^^^^^
                  ~2.9 MB OVERLAP
```

When the kernel tries to allocate user stack pages at `0xEFFF8000`, the address falls inside the scratch region where no user page table exists, causing:

```
[ERROR][vmem] lookup_user_page_table(): page table not found
```

Process creation fails, `nanvixd` gets no response, and reports `Broken pipe (os error 32)`.

## Why 128 MB Works

With 128 MB, scratch is ~113 MB → `scratch_base ≈ 0xF7D1E000`, which is above the user stack. No overlap.

## Reproduction

**Prerequisites:** Nanvix toolchain installed, SQLite port checked out.

1. Build Nanvix at v0.12.457 or later with 256 MB memory:

```bash
cd nanvix/
git checkout v0.12.461
make MACHINE=hyperlight DEPLOYMENT_MODE=multi-process RELEASE=yes TARGET=x86 MEMORY_SIZE=256 all install
```

2. Build SQLite for Nanvix using the sysroot:

```bash
cd distro/usr/lib/sqlite/
make -f Makefile.nanvix build
```

3. Run the functional test (or any test that spawns a process via nanvixd):

```bash
make -f Makefile.nanvix test-functional MEMORY_SIZE=256
```

**Expected:** Tests complete successfully.
**Actual:** `Error: Broken pipe (os error 32)` — process creation fails.

**Confirming the bisection:** Repeat step 1 with `v0.12.448` — tests pass at 256 MB.

| Version    | 128 MB | 256 MB |
|------------|--------|--------|
| v0.12.448  | ✅ pass | ✅ pass |
| v0.12.457+ | ✅ pass | ❌ fail |

## Suggested Fix Directions

- **Cap the scratch region** so `scratch_base` never descends below `USER_END_RAW`.
- **Lower `USER_END_RAW`** below the worst-case scratch base address.
- **Add a runtime validation** in `init()` that asserts `scratch_base >= USER_END_RAW` and fails with a clear error instead of a silent page-table miss.

## References

- Nanvix PR #2149 (`fceb2072e1`) — first-bad commit
- SQLite PR nanvix/sqlite#113 — version bump that exposed the regression
- Hyperlight `MAX_GPA` definition: `hyperlight-common/src/arch/i686/layout.rs:23`
- User stack constants: `src/libs/config/src/lib.rs:91-122`
- Scratch layout: `src/kernel/src/hal/platform/hyperlight/mod.rs:~1131`

ppenna:
On Hyperlight the scratch region is placed at the top of the 32-bit GPA
space ([MAX_GPA - scratch_size + 1, MAX_GPA + 1)).  Because scratch_size
can be as large as MEMORY_SIZE, the MMIO structures at the bottom of the
scratch region can descend into the user address space, corrupting the
user stack.

## Changes

**Config crate** (`src/libs/config/src/lib.rs`):
- Replace the fixed `USER_END_RAW` (`0xf000_0000`) with a Hyperlight-aware
  computation: `floor_4MB(HYPERLIGHT_GPA_CEILING - MEMORY_SIZE)`.
  The non-Hyperlight branch keeps the original value.
- Extract the magic `0xFEE0_0000` (`MAX_GPA + 1`) into the named constant
  `HYPERLIGHT_GPA_CEILING` so the derivation is self-documenting.
- Add a const assert that `USER_END_RAW > USER_BASE_RAW` to catch
  underflows if the `MEMORY_SIZE` ceiling is raised in the future.

**Kernel HAL** (`src/kernel/src/hal/platform/hyperlight/mod.rs`):
- Add a compile-time assertion that `HYPERLIGHT_GPA_CEILING` equals
  `MAX_GPA + 1` from `hyperlight-common`, so the two values cannot silently
  diverge.
- Add a runtime guard in `init()` that returns an error with diagnostics
  when `scratch_base_address < USER_END_RAW`, providing a clear failure
  instead of silent memory corruption.

Closes #2182

### DIFF
--- a/src/kernel/src/hal/platform/hyperlight/mod.rs
+++ b/src/kernel/src/hal/platform/hyperlight/mod.rs
@@ -235,6 +235,9 @@ pub struct Platform {
     (::config::kernel::KPOOL_SIZE / mem::PAGE_SIZE).is_multiple_of(u8::BITS as usize)
 );
 
+// Ensure the config crate's GPA ceiling matches the upstream MAX_GPA from hyperlight-common.
+::static_assert::assert_eq!(memory_layout::HYPERLIGHT_GPA_CEILING == MAX_GPA + 1);
+
 // Scratch-reserved layout.
 //
 // These structures are allocated in the scratch region (outside the CoW snapshot) so that
@@ -1129,6 +1132,24 @@ pub fn init(
     // MAX_GPA (not MAX_GVA) because the guest uses identity mapping (GVA == GPA)
     // and the KVM memory slot is placed relative to MAX_GPA.
     let scratch_base_address: usize = MAX_GPA - scratch_size + 1;
+
+    // Sanity check: the scratch region must not descend into the user virtual address space.
+    // The config crate already computes USER_END_RAW to avoid this overlap, but verify at
+    // runtime in case the memory layout parameters diverge from compile-time assumptions.
+    // When they do, the reserved MMIO structures (input/output buffers) at the bottom of the
+    // scratch region would collide with user stack pages.
+    if scratch_base_address < memory_layout::USER_END_RAW {
+        let reason: &str = "scratch region overlaps user address space";
+        error!(
+            "init(): {} (scratch_base={:#010x} < USER_END_RAW={:#010x}, scratch_size={:#x})",
+            reason,
+            scratch_base_address,
+            memory_layout::USER_END_RAW,
+            scratch_size
+        );
+        return Err(Error::new(ErrorCode::InvalidArgument, reason));
+    }
+
     // scratch_end is the exclusive end of the full scratch range, including the last page
     // reserved for Hyperlight bookkeeping metadata.
     let scratch_end_address: usize =
--- a/src/libs/config/src/lib.rs
+++ b/src/libs/config/src/lib.rs
@@ -83,12 +83,53 @@ pub mod memory_layout {
     ///
     pub const USER_BASE_RAW: usize = KERNEL_END_RAW;
 
-    ///
-    /// # Description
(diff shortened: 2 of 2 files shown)


ppenna:
@danbugs fyi.

PR #2185, opened 2026-04-23
[kernel] Relocate kernel heap backing storage

### LINKED ISSUE #2153: [kernel] Relocate HEAP_STORAGE from BSS to scratch memory
## Parent Issue

Part of #2152.

## Description

`HEAP_STORAGE` in `src/kernel/src/mm/kheap.rs` is a page-aligned `[u8; MIN_HEAP_SIZE]` (~1.25 MiB with 10 Hyperlight slabs) that currently lives in BSS. On Hyperlight, this is part of the CoW snapshot, so every slab allocation dirties snapshot pages.

## Approach

1. Add `kheap::set_backing_storage(ptr: *mut u8, size: usize) -> Result<(), Error>` that validates alignment and size, then records the pointer.
2. Change `kheap::init()` to use the recorded pointer instead of `&mut HEAP_STORAGE`.
3. **microvm**: keep the existing BSS `static mut HEAP_STORAGE` and pass its address in `Platform::init()`.
4. **Hyperlight**: allocate `MIN_HEAP_SIZE` bytes from the scratch region (add `HEAP_OFFSET` and `HEAP_STORAGE_SIZE` constants to the scratch layout) and pass that address.
5. Increase `SCRATCH_RESERVED_SIZE` to account for the new allocation.

## Notes

- `HEAP_STORAGE` must be page-aligned (4 KiB).
- On Hyperlight, `kheap::init()` is called during `hyperlight_pre_kmain()` (evolve phase), which runs **before** `Platform::init()`. The scratch region discovery and heap backing must be sequenced carefully — the heap storage address may need to be computed from the PEB or a known fixed offset during evolve.
- The `ALLOCATOR` (ArenaAllocator) ZST and `HEAP` (Option<Kheap>) can remain in BSS since they are small.

ppenna:
Closes #2153 

Move the kernel heap backing buffer out of BSS and into the scratch-reserved region on Hyperlight so that heap allocations no longer dirty Copy-on-Write snapshot pages. This reduces the number of pages that must be restored on every sandbox warm-start.

The change decouples the heap allocator from the backing storage location:

- **kheap**: remove the BSS-resident `HeapStorage` static; add `set_backing_storage()` to accept an externally-provided buffer, and require it to be called before `init()`.
- **microvm**: adopt the `HeapStorage` static (BSS-resident) and expose `setup_heap_backing_storage()` to wire it to the heap allocator.
- **hyperlight**: reserve a `HEAP_STORAGE` slot in the scratch layout and derive its address from the PEB input buffer pointer during `hyperlight_pre_kmain()`, before initialising the heap.
- **kmain**: call `platform::setup_heap_backing_storage()` before `kheap::init()` on non-Hyperlight platforms.

### DIFF
--- a/src/kernel/src/hal/platform/hyperlight/mod.rs
+++ b/src/kernel/src/hal/platform/hyperlight/mod.rs
@@ -264,6 +264,9 @@ scratch_layout! {
     /// IDTR backing storage.
     IDTR               : size = idt::IDTR_SIZE,
                          align = WORD_ALIGNMENT;
+    /// Kernel heap backing storage (relocated from BSS to avoid dirtying CoW snapshot pages).
+    HEAP_STORAGE       : size = crate::mm::kheap::MIN_HEAP_SIZE,
+                         align = PAGE_ALIGNMENT;
 }
 
 //==================================================================================================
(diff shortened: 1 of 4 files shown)


ppenna:
@danbugs fyi

PR #2191, opened 2026-04-27
[kernel] Split bootstrap assembly into per-platform modules

ppenna:
Move the shared `start.rs` and `start16.rs` from `hal::arch::x86::asm` into platform-specific copies under `hal::platform::microvm` and `hal::platform::hyperlight`, resolving the `cfg!`-based conditionals at the source level.

### microvm/start.rs
- Remove BSS-clearing loop (`CLEAR_BSS=0`): the host already zeroes guest memory via `mmap(MAP_ANONYMOUS)` / `VirtualAlloc(MEM_COMMIT)` and the ELF loader's `write_bytes()`.
- Remove Hyperlight evolve phase (`HYPERLIGHT_EVOLVE=0`).

### hyperlight/start.rs
- Inline BSS-clearing loop unconditionally (`CLEAR_BSS=1`).
- Inline `call hyperlight_pre_kmain` unconditionally (`HYPERLIGHT_EVOLVE=1`).
- Remove the `.if`/`.endif` assembler directives and the associated Rust constants.

### start16.rs (both platforms)
- Identical copies of the 16-bit trampoline; no behavioral changes.

### Supporting changes
- `asm::constants` visibility widened to `pub(crate)` so platform modules can reference `KSTACK_GUARD_PATTERN`, `KSTACK_SIZE`, and `KREDZONE_SIZE`.
- `hal::arch::x86::asm` and `hal::arch::x86_64::asm` visibility widened to `pub`.
- Remove `mod start` and `mod start16` from `asm/mod.rs`.


ppenna:
@danbugs fyi

Issue #1709, opened 2026-03-19
[tests] Test parity between `microvm` and `hyperlight` machines


ppenna:
## Summary

Several tests in `test/test-standalone.toml` are currently restricted to `MACHINE=microvm` via the `runs_on = ["microvm"]` filter and are **skipped** when running with `MACHINE=hyperlight`. We should investigate each case and either enable the test on hyperlight or document why it cannot apply.

## Affected Tests

The following tests are microvm-only today:

| Executor | Program | Config Location |
|----------|---------|-----------------|
| `http` | `./bin/test-mmio-fault.elf` | `test/test-standalone.toml` |
| `http` | `./bin/vfs-test.elf` | `test/test-standalone.toml` |
| `terminal` | `./bin/test-mmio-fault.elf` | `test/test-standalone.toml` |
| `terminal` | `./bin/vfs-test.elf` | `test/test-standalone.toml` |

### Details

- **`test-mmio-fault.elf`** — Tests MMIO fault handling. Currently microvm-only because MMIO is specific to the microvm machine type. Investigate whether an equivalent fault-handling mechanism exists on hyperlight and whether this test can be adapted or a hyperlight-specific variant created.
- **`vfs-test.elf`** — Tests the VFS layer. Currently restricted to microvm. Determine why this test does not run on hyperlight and what work is needed to enable it.

## Acceptance Criteria

- [ ] For each microvm-only test, determine whether it can run on hyperlight.
- [ ] Enable tests on hyperlight where possible (remove `runs_on` restriction or change to `runs_on = ["microvm", "hyperlight"]`).
- [ ] For tests that fundamentally cannot run on hyperlight, document the reason in a comment in `test/test-standalone.toml`.
- [ ] Ensure no regressions on microvm after any changes.

PR #2239, opened 2026-05-01
[kernel] Fix kctrl PDE allocation and enable MMIO tests on hyperlight

danbugs:
## Summary
- Fix `kctrl()` to allocate missing page directory entries when mapping MMIO regions. In debug builds, the larger kernel binary shifts the RAMFS MMIO region across a PDE boundary that the host page tables don't cover. Mirrors the existing allocation pattern from `map_kpage()`.
- Enable `mmio_ramfs` subtest in test-kernel on hyperlight (removed `#[cfg(not(feature = "hyperlight"))]` source guard)
- Enable `test-mmio-fault` across all three deployment modes (removed `runs_on = ["microvm"]` from standalone, single-process, and multi-process TOML configs)

## Test plan
- [x] `make MACHINE=hyperlight DEPLOYMENT_MODE=standalone LOG_LEVEL=trace run-nanvix-tests`
- [x] `make MACHINE=hyperlight DEPLOYMENT_MODE=single-process LOG_LEVEL=trace run-nanvix-tests`
- [x] `make MACHINE=hyperlight DEPLOYMENT_MODE=multi-process LOG_LEVEL=trace run-nanvix-tests`
- [x] `make MACHINE=hyperlight DEPLOYMENT_MODE=standalone RELEASE=yes LOG_LEVEL=error run-nanvix-tests`
- [x] `make MACHINE=hyperlight DEPLOYMENT_MODE=single-process RELEASE=yes LOG_LEVEL=error run-nanvix-tests`
- [x] `make MACHINE=hyperlight DEPLOYMENT_MODE=multi-process RELEASE=yes LOG_LEVEL=error run-nanvix-tests`
- [x] `make MACHINE=hyperlight DEPLOYMENT_MODE=standalone lint-check`
- [x] `make MACHINE=hyperlight DEPLOYMENT_MODE=standalone format-check`

Progress on #1709.

### DIFF
--- a/src/kernel/src/mm/platform_vas.rs
+++ b/src/kernel/src/mm/platform_vas.rs
@@ -15,8 +15,11 @@ use crate::{
         arch::x86::mem::mmu::page_table::PageTable,
         mem::{
             Address,
+            FrameAddress,
             MemoryRegion,
             MemoryRegionType,
+            MmioCachePolicy,
+            PageAddress,
             PageAligned,
             PageTableAddress,
             PageTableAligned,
@@ -38,6 +41,7 @@ use ::arch::mem::{
         PteWord,
     },
     PAGE_TABLE_LENGTH,
+    PGTAB_ALIGNMENT,
 };
 use ::sparse_bitmap::SparseBitmap;
 use ::sys::error::{
@@ -187,7 +191,7 @@ pub fn init(
     // owns all paging structures. This ensures the kernel can find and modify PTEs for MMIO
     // permission changes without depending on host page table memory.
     let kernel_pages: LinkedList<KernelPage> = LinkedList::new();
-    let kernel_page_tables: LinkedList<(PageTableAddress, PageTable<PageTableStorage>)> = {
+    let mut kernel_page_tables: LinkedList<(PageTableAddress, PageTable<PageTableStorage>)> = {
         let mut page_tables: LinkedList<(PageTableAddress, PageTable<PageTableStorage>)> =
             LinkedList::new();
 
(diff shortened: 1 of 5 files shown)


src/kernel/src/pm/process/manager/mod.rs
```diff
@@ -2024,23 +2024,30 @@ impl ProcessManager {
         })?;
         let perm: AccessPermission = region.perm();
 
+        let page_table_allocator = || {
+            let mm: &mut VirtMemoryManager = unsafe { VirtMemoryManager::get_mut() };
+            let kpage = mm.alloc_kpage(true)?;
+            let storage = crate::mm::PageTableStorage::KernelPage(kpage);
+            Ok(crate::hal::arch::x86::mem::mmu::page_table::PageTable::new(storage))
+        };
```
ppenna: On platforms that enable `platform-root-virtual-address-space-bootstrap`, this should not be used at all, as all page tables are provisioned eagerly. I will open an issue to make sure we gate properly.

ppenna (commented):
All MMIO regions are added to the kernel address virtual address space at startup time for all platforms. This is to allow simple page table manipulation whenever we need to grand permissions of certain pages to user processes.

The fix in this PR is a workaround because after startup time, the set of kernel page tables must not change at all, unless lazy identity mapping is in place, which is not the case for hyperlight.

Did you check if when we are creating a new address space on hyperlight, we are mapping all "kernel" pages?

ppenna:
> All MMIO regions are added to the kernel address virtual address space at startup time for all platforms. This is to allow simple page table manipulation whenever we need to grand permissions of certain pages to user processes.
> 
> The fix in this PR is a workaround because after startup time, the set of kernel page tables must not change at all, unless lazy identity mapping is in place, which is not the case for hyperlight.
> 
> Did you check if when we are creating a new address space on hyperlight, we are mapping all "kernel" pages?

I will gather more findings on my end as well and share them here.

ppenna:
> > All MMIO regions are added to the kernel address virtual address space at startup time for all platforms. This is to allow simple page table manipulation whenever we need to grand permissions of certain pages to user processes.
> > The fix in this PR is a workaround because after startup time, the set of kernel page tables must not change at all, unless lazy identity mapping is in place, which is not the case for hyperlight.
> > Did you check if when we are creating a new address space on hyperlight, we are mapping all "kernel" pages?
> 
> I will gather more findings on my end as well and share them here.

@danbugs it looks like the root cause of the problem is because Hyperlight is not adding underling pages of the mapped file itself to the host-provided page tables.

ppenna:
Filed #2240 to track the upstream Hyperlight page table gap and the gating of `page_table_allocator` in `kctrl()` on `platform-root-virtual-address-space-bootstrap` platforms.

Reference: https://github.com/nanvix/nanvix/pull/2239#discussion_r3175470583

ppenna:
@danbugs https://github.com/nanvix/nanvix/pull/2242 introduced improvements that help this work.

ppenna (changes requested):
This is a workaround.

The correct fix is to create a virtual memory initialization routine for platforms that provide an initial paging structures and call it from [`platform_vas::init()`](https://github.com/nanvix/nanvix/blob/ff9ebd4206b1cda25fe6452f21a2d09e1f21612d/src/kernel/src/mm/platform_vas.rs#L154). See [`kernel_vas:init()`](https://github.com/nanvix/nanvix/blob/ff9ebd4206b1cda25fe6452f21a2d09e1f21612d/src/kernel/src/mm/kernel_vas.rs#L112) for reference.

src/kernel/src/pm/process/manager/mod.rs
danbugs: hyperlight expects guests to still map regions mapped w/ map_file_cow.

src/kernel/src/pm/process/manager/mod.rs
```diff
@@ -2024,23 +2024,30 @@ impl ProcessManager {
         })?;
         let perm: AccessPermission = region.perm();
 
+        let page_table_allocator = || {
```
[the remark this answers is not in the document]
danbugs: Outdated.

src/kernel/src/mm/virt/vmem.rs
```diff
@@ -1314,11 +1314,12 @@ impl Vmem {
...
     /// - [`ErrorCode::NoSuchEntry`]: The page table entry was not found (dry run only).
     ///
-    pub fn kctrl(
+    pub fn kctrl<T: Fn() -> Result<PageTable<PageTableStorage>, Error>>(
         &mut self,
         vaddr: PageAligned<VirtualAddress>,
         access: AccessPermission,
         dry_run: bool,
+        page_table_allocator: &T,
     ) -> Result<(), Error> {
```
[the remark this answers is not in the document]
danbugs: Outdated.

src/kernel/src/mm/virt/vmem.rs
```diff
@@ -1344,14 +1346,31 @@ impl Vmem {
...
 
-            // Check if corresponding page table does not exist.
             if !pde.is_present() {
-                let reason: &str = "page table not present";
-                error!("{reason}");
-                return Err(Error::new(ErrorCode::NoSuchEntry, reason));
-            };
-
-            self.lookup_kernel_page_table(&pde)?
+                if dry_run {
```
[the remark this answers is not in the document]
danbugs: Outdated.

src/kernel/src/mm/platform_vas.rs
```diff
@@ -238,5 +238,46 @@ pub fn init(
...
+        let end: usize = base + region.size();
+        let mut addr: usize = base;
+        while addr < end {
+            if let Ok(vaddr) = PageAligned::<VirtualAddress>::from_raw_value(addr) {
+                vmem.ensure_pde(vaddr, &page_table_allocator)?;
+            }
+            // Advance to the next 4MB PDE boundary.
+            addr = (addr + mem::PGTAB_SIZE) & !(mem::PGTAB_SIZE - 1);
+        }
+    }
```
ppenna, line 381: We are going on the right direction, but instead of adding logic for ensuring PDE, we should introduce the equivalent of:

https://github.com/nanvix/nanvix/blob/96597bfd6aa12e268c7977ac0b3ecd9ae857a291/src/kernel/src/mm/virt/boot_init.rs

This creates set of additional root page tables that are passed to the virtual memory allocator, and handles append to kernel pages transparently.

This will ensure that address space of new processes is correctly crated.

ppenna (changes requested):

src/kernel/src/mm/platform_vas.rs
```diff
@@ -238,5 +243,112 @@ pub fn init(
...
+    host_page_tables: &LinkedList<(PageTableAddress, PageTable<PageTableStorage>)>,
+) -> Result<LinkedList<(PageTableAddress, PageTable<PageTableStorage>)>, Error> {
+    let mut page_tables: LinkedList<(PageTableAddress, PageTable<PageTableStorage>)> =
+        LinkedList::new();
+
+    for region in mmio_regions.iter() {
+        let base: usize = region.start().into_raw_value();
+        if crate::hal::platform::is_scratch_address(base) {
+            continue;
+        }
```
ppenna: This should not be required. If it is, something else is broken as well (ie, registration of issues in the PAL).

src/kernel/src/mm/platform_vas.rs
```diff
@@ -238,5 +243,112 @@ pub fn init(
...
+                PageTableAddress::new(PageTableAligned::from_raw_value(pt_aligned)?);
+
+            // Skip if the host already has a page table for this PDE range.
+            let covered_by_host: bool = host_page_tables
+                .iter()
+                .any(|(addr, _)| addr.into_raw_value() == pt_addr.into_raw_value());
+            if covered_by_host {
+                raw_vaddr += mem::PAGE_SIZE;
+                continue;
+            }
```
ppenna: Remove this check. Having a host PDE does not mean that underlying PTEs are set and correctly initialized.

src/kernel/src/mm/platform_vas.rs
```diff
@@ -238,5 +243,112 @@ pub fn init(
...
+/// with the region's cache policy. Reuses the last page table if consecutive pages fall in the
+/// same 4MB PDE range.
+fn build_mmio_page_tables(
+    mmio_regions: &LinkedList<TruncatedMemoryRegion<VirtualAddress>>,
+    host_page_tables: &LinkedList<(PageTableAddress, PageTable<PageTableStorage>)>,
+) -> Result<LinkedList<(PageTableAddress, PageTable<PageTableStorage>)>, Error> {
+    let mut page_tables: LinkedList<(PageTableAddress, PageTable<PageTableStorage>)> =
+        LinkedList::new();
+
+    for region in mmio_regions.iter() {
```
ppenna, line 289: Sort `mmio_region` by start address to (i) have an easy collision logic, (ii) you can pop back page tables on region walk logic.

src/kernel/src/mm/platform_vas.rs
```diff
@@ -238,5 +243,112 @@ pub fn init(
...
+            page_table.map(
+                PageAddress::new(PageAligned::from_raw_value(raw_vaddr)?),
+                paddr,
+                true,
+                cache_policy.write_through(),
+                cache_policy.cache_enabled(),
+                region.perm(),
+            )?;
+            page_tables.push_back((page_table_addr, page_table));
+            raw_vaddr += mem::PAGE_SIZE;
```
ppenna: Nitpick: You may consider using the `fill()` kernel function to colapse this operation for PTEs in the same page table, as we do in https://github.com/nanvix/nanvix/blob/96597bfd6aa12e268c7977ac0b3ecd9ae857a291/src/kernel/src/mm/virt/boot_init.rs.

ppenna (changes requested):
We are getting there....

src/kernel/src/mm/platform_vas.rs
danbugs: Removed the is_scratch_address skip entirely. All MMIO regions are now processed uniformly. We use gva_to_gpa() for the physical address, which returns the address unchanged for identity-mapped regions and does proper GVA->GPA translation for scratch regions — so it works correctly for both without needing to special-case.

src/kernel/src/mm/platform_vas.rs
danbugs: Removed the covered_by_host skip entirely. We now attempt to map PTEs into ALL page tables — existing host-copied ones included. If a PTE is already present (host already initialized it), map() returns ResourceBusy and we treat that as a no-op. If it's not present, we set it up correctly.

src/kernel/src/mm/platform_vas.rs
danbugs: Regions are sorted by start address before traversal. We drain the page tables into a sorted Vec and use positional lookup to find/insert page tables at the correct sorted position, maintaining order throughout.

src/kernel/src/mm/platform_vas.rs
danbugs: Leaving this for a follow-up.

src/kernel/src/mm/platform_vas.rs
```diff
@@ -238,5 +242,102 @@ pub fn init(
         page_tables
     };
 
+    // Map MMIO regions into existing or new page tables. Regions mapped after snapshot
+    // construction (e.g., RAMFS via map_file_cow) may fall in 4MB ranges without a PDE.
+    // This follows the boot_init::init() pattern: build page tables with proper PTEs and
+    // pass them to the virtual memory manager so that process address space cloning
+    // includes these regions from the start.
```
ppenna: Reword this commenbt. `map_file_cow` is something specific to hyperlight and the current file is more generic than that:

```
// Map MMIO regions into existing or new page tables.
```

src/kernel/src/mm/platform_vas.rs
```diff
@@ -238,5 +242,102 @@ pub fn init(
...
+
     VirtMemoryManager::init(kernel_pages, kernel_page_tables)
 }
+
+/// Maps MMIO regions into existing or new page tables.
+///
+/// For each page in each MMIO region, finds the page table covering the corresponding PDE range
+/// in `kernel_page_tables` and maps the PTE into it. If no page table exists for that range, a
+/// new BSS-backed page table is allocated and inserted. Regions are sorted by start address to
+/// enable ordered traversal.
```
ppenna: Stick to existing documentation style in this function.

src/kernel/src/mm/platform_vas.rs
ppenna: Okay. Can you create issue and link it to the source code?

src/kernel/src/mm/platform_vas.rs
```diff
@@ -238,5 +242,102 @@ pub fn init(
...
+            match pts[idx].1.map(
+                PageAddress::new(PageAligned::from_raw_value(raw_vaddr)?),
+                paddr,
+                true,
+                cache_policy.write_through(),
+                cache_policy.cache_enabled(),
+                region.perm(),
+            ) {
+                Ok(()) => {},
+                Err(e) if e.code == ErrorCode::ResourceBusy => {},
```
ppenna: This should be an error. Address space collision means bad platform initialization code.

ppenna (changes requested):

PR #2265, opened 2026-05-05
[kernel] Fix hyperlight frame allocator size

ludfjig:
This fixes an issue where the bitmap could be too small

... will update this descritpion with testing criteria once I have finished running them again to confirm


ppenna (approved):
LGTM

PR #2272, opened 2026-05-05
[kernel] Translate kernel GVA to GPA in user-copy paths

ludfjig:
Fixes the kernel<->user copy paths on on hyperlight i686-guest, where super::memcpy expects GPAs on both endpoints but the kernel side was passing a GVA.

Brings cpython on hyperlight standalone to parity with microvm: the hello test now passes (it failed before with "Hello test did not produce expected output"). Remaining regrtest failures are unrelated

### DIFF
--- a/src/kernel/src/mm/virt/vmem.rs
+++ b/src/kernel/src/mm/virt/vmem.rs
@@ -731,8 +731,9 @@ impl Vmem {
 
                 if !dry_run {
                     // Copy memory from user space to kernel space.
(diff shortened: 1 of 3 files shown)


ppenna (changes requested):
Address translation is wrong. We should use `virt_to_phys()` instead. In Hyperlight that will perform a software page table walk.

However, the important question is: why the current implementation is not triggering any failures?

Can you introduce a test case that triggers the failures you observing? This way, we can validate that this change is actually fixing the bug we observed.

PR #2293, opened 2026-05-06
[kernel] Page-align hyperlight initrd module size

ludfjig:
Page-align the hyperlight initrd `module_size` at the producer (`parse_bootinfo`) instead of relying on `kmain` to round it up at the consumer.

- #2219 added page alignment in `kmain`'s module-region registration loop.
- #2252 extended `module_size` to cover all four host-placed components (`INITRD_SIZE_BYTES + initrd_size + WIRE_SIZE + cmdline_len`) and fixed the spill-page fault that escaped in 0.12.521.

`module_size` covers every host-placed byte, but its end is not page aligned (e.g. `0x117002` for `echo-cpp`). The frame allocator gets the right pages today only because `kmain` does `align_up`. This PR makes the page-aligned size explicit at the source so the property is local to where the size is computed.

### DIFF
--- a/src/kernel/src/hal/platform/hyperlight/mod.rs
+++ b/src/kernel/src/hal/platform/hyperlight/mod.rs
@@ -1451,6 +1451,12 @@ pub fn parse_bootinfo(magic: u32, info: usize) -> Result<BootInfo, Error> {
         // must span from initrd_base to the end of the cmdline payload.
         let module_size: usize =
             INITRD_SIZE_BYTES + initrd_size + CmdlineArgsLen::WIRE_SIZE + cmdline_len.as_usize();
+        let module_size: usize =
+            ::sys::mm::align_up(module_size, PAGE_ALIGNMENT).ok_or_else(|| {
+                let reason: &str = "initrd module size align_up overflow";
+                error!("parse_bootinfo(): {reason}");
+                Error::new(ErrorCode::BadFile, reason)
+            })?;
 
         let module: KernelModule =
             KernelModule::new(PhysicalAddress::from_raw_value(initrd_base)?, module_size, cmdline);


src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -1925,11 +1925,24 @@ pub fn init(
...
+            // Snapshot tail. Reserve every page in [kpool_end, snapshot_end_address)
+            // not already claimed above so the frame allocator never hands out a
+            // snapshot frame as a user frame.
+            //
+            // Covers the trailing args blob the single binary path appends past
+            // actual_initrd_size.
+            let mut tail_start: usize = kpool_end;
+            for region in memory_regions.iter() {
+                let region_start: usize = region.start().into_raw_value();
+                let region_end: usize = region_start + region.size();
```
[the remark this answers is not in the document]
danbugs: This is a needed change.

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -1925,11 +1925,24 @@ pub fn init(
...
+                if region_start < kpool_end || region_end > snapshot_end_address {
+                    continue;
+                }
+                if region_end > tail_start {
+                    tail_start = region_end;
+                }
+            }
```
ppenna: Why do we need to traverse the memory regions here? Isn't the `init_data` region the last region that we push anyways? If so, why the previous calculation is not correctly capturing the required padding?

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -1925,11 +1925,24 @@ pub fn init(
...
-            // The host allocates a full snapshot budget that may extend beyond the last module.
-            let init_data_end: usize = init_data_start + init_data_size;
-            let tail_start: usize =
-                ::sys::mm::align_up(init_data_end, PAGE_ALIGNMENT).unwrap_or(init_data_end);
+            // Snapshot tail. Reserve every page in [kpool_end, snapshot_end_address)
+            // not already claimed above so the frame allocator never hands out a
+            // snapshot frame as a user frame.
+            //
+            // Covers the trailing args blob the single binary path appends past
+            // actual_initrd_size.
```
ppenna: If we are appending `args` past the `actual_initrd_size`, it is likely that the root case is on the initrd module registration and the fix belongs there.

src/kernel/src/hal/platform/hyperlight/mod.rs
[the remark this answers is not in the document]
danbugs: Maybe do:
```
  let region_end: usize = match region_start.checked_add(region.size()) {                                              
      Some(end) => end,                                           
      None => continue,
  };
```

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -1433,27 +1432,25 @@ pub fn parse_bootinfo(magic: u32, info: usize) -> Result<BootInfo, Error> {
         // Single-binary format: parse old size-header + args layout.
         info!("parse_bootinfo(): single-binary initrd detected");
 
-        let (initrd_base, initrd_size, (cmdline_len, cmdline)) = unsafe {
-            let (base, size, cmdline) = parse_initrd_image(current_data_start, total_size)?;
-            (base, size, cmdline)
+        let (initrd_base, initrd_size, region_size, (_cmdline_len, cmdline)) = unsafe {
+            let (base, size, region_size, cmdline) =
+                parse_initrd_image(current_data_start, total_size)?;
+            (base, size, region_size, cmdline)
```
ppenna: Please revert these changes to keep PR minimal. Renaming variables is out of the scope of this PR.

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -1433,27 +1432,25 @@ pub fn parse_bootinfo(magic: u32, info: usize) -> Result<BootInfo, Error> {
...
         };
 
         info!(
-            "initrd_base={:#010x}, initrd_size={:#010x}, cmdline={:?}",
-            initrd_base, initrd_size, cmdline
+            "initrd_base={:#010x}, initrd_size={:#010x}, region_size={:#010x}, cmdline={:?}",
+            initrd_base, initrd_size, region_size, cmdline
```
ppenna: Please revert these changes to keep PR minimal. Renaming variables is out of the scope of this PR.

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -1433,27 +1432,25 @@ pub fn parse_bootinfo(magic: u32, info: usize) -> Result<BootInfo, Error> {
...
-        let module: KernelModule =
-            KernelModule::new(PhysicalAddress::from_raw_value(initrd_base)?, module_size, cmdline);
+        let initrd_phys: PhysicalAddress = PhysicalAddress::from_raw_value(initrd_base)?;
+        let module: KernelModule = KernelModule::new_with_region(
+            initrd_phys,
+            initrd_size,
+            initrd_phys,
+            region_size,
+            cmdline,
+        );
```
ppenna: Keep inline construction of `initrd_phys` to keep PR minimal.

src/kernel/src/hal/platform/hyperlight/mod.rs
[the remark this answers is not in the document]
ludfjig: outdated

src/kernel/src/hal/platform/hyperlight/mod.rs
```diff
@@ -2372,7 +2371,19 @@ unsafe fn parse_initrd_image(
     }
     let initrd_base: usize = init_data_start;
 
-    Ok((initrd_base, actual_initrd_size, initrd_cmdline))
+    // Compute the page-aligned region size so the caller registers every page the
+    // host placed at init_data_start (size header + ELF payload + args trailer)
+    // with the frame allocator. Without this, trailing args bytes that spill into
+    // a page beyond the ELF's page-aligned end would be left unbooked and the
+    // frame allocator could later hand them out as user frames.
```
ppenna: Reword documentation to remove cross-references that are hard to maintain:

```
Compute the page-aligned region size so the caller registers every page the
host placed at init_data_start (size header + ELF payload + args trailer)
with the frame allocator.
```

src/kernel/src/hal/platform/hyperlight/mod.rs
ludfjig: I am introducing a new variable `region_size`, would you prefer not to log it with `info!`?

src/kernel/src/hal/platform/hyperlight/mod.rs
ludfjig: I am adding a new one, not sure what you mean could you clarify?

src/kernel/src/hal/platform/hyperlight/mod.rs
ludfjig: initrd_phys is used twice, do you want me to inline it twice identically?

src/kernel/src/hal/platform/hyperlight/mod.rs
ludfjig: fixed

src/kernel/src/hal/platform/hyperlight/mod.rs
ppenna: No, thanks for pointing out!

ppenna (approved):

PR #2307, opened 2026-05-09
[kernel] Widen identity-map visibility for dynamic kernel frames

ppenna:
## Summary

Expose identity-mapping helpers and add new functions needed by the unified physical memory allocator. This is preparatory infrastructure — no allocator behavior changes yet.

## Motivation

Part of the physical memory allocator unification (#1270). Once the kpool is removed, kernel frames can be allocated from any physical address. These frames must be identity-mapped before use, and their PDEs must be propagated into user page directories so that context switches don't fault.

## PR Stack

This is **PR 3 of 4** in the kpool removal series:

1. #2305 — `Bitmap::usage()` / `SparseBitmap::usage()` (no dependencies)
2. #2306 — Config: add `kernel_watermark` parameter (no dependencies)
3. **This PR** — Identity-map infrastructure for dynamic kernel frames (no dependencies)
4. #2297 — Unify allocator: delete kpool, add kframe, watermark gating (depends on 1-3)

> **Note:** CI will report dead-code warnings for `identity_map_page`, `propagate_kernel_pdes`, and `propagate_pde_to_current_pd` because their callers are introduced in PR 4. These warnings are expected and will resolve when the stack merges.

### Visibility widening
- `identity_map_page` — `fn` → `pub(crate) fn`
- `with_kernel_address_space` — `fn` → `pub(crate) fn`
- `memset` — `use` → `pub(in crate::mm) use`
- `KernelStack::base()`, `KernelStack::size()` — `fn` → `pub(crate) fn`

### New functions
- `propagate_pde_to_current_pd()` — copies a kernel PDE into the active user PD
- `propagate_kernel_pdes()` — copies kernel PDEs for a physical address range into a target PD
- `with_kernel_address_space()` in `no_identity_map.rs` — no-op passthrough for hyperlight

### Behavioral changes
- `kstack.rs` — wraps guard-page write/check in `with_kernel_address_space()` so they work when CR3 points to a user PD

### Re-exports
- `mm::virt::mod.rs` — re-exports `identity_map_page`, `with_kernel_address_space`, `propagate_kernel_pdes`
- `mm::mod.rs` — re-exports `propagate_kernel_pdes`, `with_kernel_address_space`

## Testing

Guard-page operations are now wrapped in `with_kernel_address_space()` which is a no-op when already in the kernel address space — no behavioral change for existing code paths. Existing tests pass.

PR #2320, opened 2026-05-11
[kernel] Pre-allocate identity-mapping PDEs at boot and sync to new address spaces

ppenna:
## Summary

Replace the purely lazy identity-mapping strategy with a two-phase approach that pre-allocates kernel page directory entries at boot and syncs them into new user address spaces.

### Commit 1: Pre-allocate kernel PDEs

- `identity_map::init()` now returns `Result<(), Error>` and pre-allocates BSS page tables for every PDE index in `[0, min(MEMORY_SIZE, KERNEL_END))` via a new `identity_map_bound()` const fn.
- Add `identity_map::sync_kernel_pdes(PageDirectoryAddress)` that copies all present kernel PDEs into a target page directory in a single pass.
- `Vmem::clone()` calls `sync_kernel_pdes()` after mapping root page tables so new user address spaces inherit kernel identity-mapping entries.
- Add `debug_assert!` ensuring `pde_count <= PAGE_TABLE_LENGTH`.
- Update module-level documentation to describe the new two-phase model.

### Commit 2: Add in-kernel tests

- `test_init_preallocates_identity_map_pdes`: verifies every PDE in the pre-allocated range is present after `init()`.
- `test_sync_kernel_pdes_copies_to_target`: allocates a zeroed kernel page as a target PD, syncs, and verifies every present kernel PDE was faithfully copied with the correct frame address.
- Tests are an inline `#[cfg(feature = "test")]` submodule of `identity_map`, invoked via `run_test!()` after `VirtMemoryManager::init()`.

## Testing

- `./z build -- all` passes.
- `./z build -- run-nanvix-tests` passes (no kernel panics).
- `./z build -- format lint` passes.

### DIFF
--- a/src/kernel/src/hal/platform/microvm/mod.rs
+++ b/src/kernel/src/hal/platform/microvm/mod.rs
@@ -85,8 +85,13 @@ use crate::hal::platform::pit::Pit;
 /// Number of page tables needed for identity-mapping physical memory regions.
 ///
 /// On microvm all physical memory is contiguous starting at GPA 0, so the base count
-/// (one page table per `PGTAB_SIZE` bytes) is sufficient.
+/// (one page table per `PGTAB_SIZE` bytes) is sufficient. When the WHP backend is enabled,
+/// an additional page table is needed for the LAPIC MMIO region at `0xFEE0_0000`, which
+/// lies outside the identity-mapped physical memory range.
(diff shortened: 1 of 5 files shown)


ppenna:
Supersedes #2307.

PR #2373, opened 2026-05-18
[kernel] Prepare kheap for Verus verification

tangruize:
Minimal exec-only refactoring to prepare kheap for upcoming Verus formal verification:

- Make `SLAB_COUNT` and `SlabSize` public (Verus visibility rules require types referenced in pub spec functions to also be pub)
- Extract `let tier` and `let r` variables in `allocate`/`deallocate`/`layout_to_allocator` (proof blocks must sit between the match result and the return)
- Rename `|_|` to `|_e|` in `map_err` closures (Verus requires named closure parameters)
- Add `#[allow(unused_variables)] let align = layout.align()` in `allocate` (needed by alignment proof)
- Add `usize_to_mut_ptr` helper (Verus parser rejects `addr as *mut u8` in verified bodies)
- Replace `1 << PAGE_SHIFT` with literal `4096` + static_assert in arch crate (Verus cannot evaluate bit-shift in const context)

### DIFF
--- a/src/kernel/src/mm/kheap.rs
+++ b/src/kernel/src/mm/kheap.rs
@@ -28,9 +28,8 @@ use ::type_safe::usize_to_mut_ptr;
 const NUM_OF_SLABS: usize = 7;
 /// Number of slabs per slab size. Each slab size is allocated a fixed number of slabs.
 pub(super) const SLAB_COUNT: usize = 32;
-/// Minimum heap size in bytes. This is the minimum size of the backing storage that must be
-/// provided to initialize the heap. It is calculated as the number of slabs multiplied by the size
-/// of each slab, which is determined by the number of slabs per slab size and the page size.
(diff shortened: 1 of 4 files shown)


src/kernel/src/mm/kheap.rs
```diff
@@ -74,6 +74,11 @@ static mut ALLOCATOR: ArenaAllocator = ArenaAllocator;
 // Implementations
 //==================================================================================================
 
+#[inline]
+fn usize_to_mut_ptr(addr: usize) -> *mut u8 {
+    addr as *mut u8
+}
```
ppenna: @jaylorch / @tangruize @arsene1995 we probably want to move this to some `utils` crate, as it is likely that we need to make such conversions in many places to support Verus?

tangruize: I agree, maybe we can use another PR to do this?

PR #2374, opened 2026-05-18
[kernel] E: Add Verus formal verification for kheap allocator

tangruize:
Add Verus formal verification for the kernel heap allocator, proving:
- Allocation returns non-null, aligned pointers to previously-free memory
- Deallocation correctly removes addresses from the abstract state
- The heap invariant (no overlapping live allocations) is preserved across all operations
- `layout_to_allocator` selects the smallest sufficient slab tier

Adds `kheap.spec.rs` (abstract KheapView, caller-facing contracts), `kheap.proof.rs` (concrete invariant, View impl, proof lemmas), Verus annotations on kheap.rs, arch crate Verus support for PAGE_SIZE/PAGE_SHIFT, and kernel to VERUS_CRATES in Makefile.

Depends on #2373 and #2371.


src/kernel/src/mm/kheap.rs
```diff
@@ -173,10 +319,38 @@ impl Kheap {
...
+                    &&& t >= spec_layout_size(*layout)
+                    &&& forall|s: usize| is_supported_tier(s) && s >= spec_layout_size(*layout) ==> t <= s
+                },
+                Err(_) => {
+                    spec_layout_size(*layout) == 0
+                        || spec_layout_size(*layout) > 512
+                },
+            },
+    )]
     pub fn layout_to_allocator(layout: &Layout) -> Result<SlabSize, AllocError> {
```
jaylorch: This probably shouldn't be `pub` in the original implementation, since we don't want anyone outside this module calling it. I suggest we remove the `pub` so that we don't have to make `is_supported_tier` part of the trusted spec, i.e., so we can move `is_supported_tier` out of `kheap.spec.rs` and into `kheap.proof.rs`.

src/kernel/src/mm/kheap.rs
```diff
@@ -142,9 +214,45 @@ impl Kheap {
...
+                    &&& final(self)@.allocations[ptr as int] >= spec_layout_size(layout) as nat
+                    &&& final(self)@ =~= old(self)@.spec_allocate(
+                        ptr as int,
+                        final(self)@.allocations[ptr as int],
+                    )
+                    &&& (ptr as int) % (spec_layout_align(layout) as int) == 0
+                },
+                Err(_) => {
+                    &&& final(self)@ == old(self)@
+                    &&& !old(self).can_allocate_layout(layout)
```
jaylorch: It's a little odd to use `can_allocate_layout` in the spec for liveness of `allocate`, since it's implementation-specific and closed. If this were a spec on a `pub` function, this wouldn't be allowed. But the only reason this isn't `pub` is that it's called by an external-body function that Verus doesn't easily support. I suggest having the spec be a reasonable public one anyway since it needs to be audited. Depending on a closed spec (for `can_allocate_layout`) that only appears in the `.proof.rs` file makes the spec too opaque.

src/kernel/src/mm/kheap.proof.rs
```diff
@@ -0,0 +1,1268 @@
...
+/// to serve a size-`size` request returning `ptr`, and the ghost map was
+/// updated accordingly, show `self.inv()`.
+///
+/// Assumptions (all discharged by slab's spec + the match branch context):
+///   * pre.inv() held,
+///   * the chosen slab's new state differs from the old by inserting `ptr`
+///     into its allocated_addrs,
+///   * all OTHER slab tiers are frame-unchanged,
+///   * alloc_map' == alloc_map.insert(ptr, size).
+#[verifier::spinoff_prover]
```
jaylorch: The use of `spinoff_prover` is often a sign that the proof is brittle. I suggest fixing the proof (e.g., by breaking it into pieces) so that it doesn't need this annotation. Note that this annotation appears many times in this file; I suggest this for every instance of it.

src/kernel/src/mm/kheap.rs
```diff
@@ -74,12 +96,53 @@ static mut ALLOCATOR: ArenaAllocator = ArenaAllocator;
...
 //==================================================================================================
 
-#[inline]
+// VERUS DEVIATION: `addr as *mut u8` is unsupported in verified bodies.
+// Keep the executable cast isolated behind a specified helper.
+#[verus_verify(external_body)]
+#[verus_spec(result =>
+    ensures
+        result as usize == addr,
+)]
```
jaylorch: You should add `#[inline]` here to make sure performance is unaffected.

src/kernel/src/mm/kheap.rs
```diff
@@ -74,12 +96,53 @@ static mut ALLOCATOR: ArenaAllocator = ArenaAllocator;
...
     addr as *mut u8
 }
 
+// Ghost-state constructor for the zero-sized allocation map used by the proof.
+#[verus_verify(external_body)]
+#[verus_spec(result =>
+    ensures
+        result@ =~= Map::empty(),
+)]
+fn empty_alloc_map() -> Ghost<Map<int, nat>> {
```
jaylorch: There's no need to have an `external_body` function to do this. Just use `Ghost(Map::<int, nat>::empty())`

src/kernel/src/mm/kheap.rs
jaylorch: It may be best to put this into `kheap.spec.rs` to make sure that it gets audited.

src/kernel/src/mm/kheap.rs
```diff
@@ -74,12 +96,53 @@ static mut ALLOCATOR: ArenaAllocator = ArenaAllocator;
...
+    /// # Success
+    /// - `addr` is page-aligned and non-zero
+    /// - `size` is a positive multiple of `MIN_HEAP_SIZE`
+    /// - Every slab tier is freshly initialized with zero live allocations
+    /// - The resulting `KheapView` is empty (no live allocations).
+    ///
+    /// # Failure
+    /// Returns `InvalidArgument` on alignment or sizing violations.
+    #[verus_spec(result =>
+        requires
```
jaylorch: Since this is called by unverified code, it would be best to avoid preconditions. So I suggest checking for these conditions at runtime in the body of the function and returning `Err` if they happen. That way you can remove the preconditions and make it reasonable to call from the `#[verifier::external]` function `init`.

src/kernel/src/mm/kheap.spec.rs
```diff
@@ -0,0 +1,139 @@
...
+        &&& forall|a1: int, a2: int| #![auto]
+                self.allocations.dom().contains(a1)
+                && self.allocations.dom().contains(a2)
+                && a1 != a2
+                ==> a1 + self.allocations[a1] as int <= a2
+                    || a2 + self.allocations[a2] as int <= a1
+    }
+
+    /// Abstract state of a freshly constructed heap: no live allocations.
+    pub open spec fn spec_new() -> KheapView {
```
jaylorch: I suggest just calling it `new`. The struct `KheapView` is obviously a view struct, meant only for use in specs. So it's redundant to say `spec_new` instead of just `new`.

src/kernel/src/mm/kheap.spec.rs
```diff
@@ -0,0 +1,139 @@
...
+///   * `align` does not exceed `size` (the allocator picks a block whose
+///     size is ≥ `size`, and that block's alignment equals its size; so
+///     `align ≤ size` is sufficient for the returned pointer to be
+///     `align`-aligned). `align` being a power of two comes from `Layout`'s
+///     own type invariant, modeled by `vstd::layout::valid_layout`.
+pub open spec fn layout_ok_for_kheap(layout: core::alloc::Layout) -> bool {
+    let s = spec_layout_size(layout);
+    let a = spec_layout_align(layout);
+    &&& 0 < s <= 512
+    &&& a <= s
```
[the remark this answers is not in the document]
jaylorch: I agree with the `Copilot` comment above. But I would put this comment in `allocate`, since that's where it's an invalid precondition. I'll comment over there.

src/kernel/src/mm/kheap.rs
```diff
@@ -142,9 +214,45 @@ impl Kheap {
...
+    /// - Returns a non-null, properly aligned pointer
+    /// - The returned address was not previously live
+    /// - The abstract view records the new size at the returned addr.
+    ///
+    /// # Failure
+    /// Abstract state is unchanged.
+    #[verus_spec(result =>
+        requires
+            old(self).inv(),
+            layout_ok_for_kheap(layout),
```
jaylorch: Since this code is called by unverified code, we should keep assumptions to the minimum. We shouldn't have `layout_ok_for_kheap` here. We should check for a layout problem at runtime, and reflect that in the postconditions (i.e., it's OK to return `Err` if the layout isn't OK for the Kheap).

src/kernel/src/mm/kheap.rs
```diff
@@ -158,10 +266,48 @@ impl Kheap {
...
+    ///
+    /// # Success
+    /// Removes `ptr` from the abstract view.
+    ///
+    /// # Failure
+    /// Abstract state is unchanged.
+    #[verus_spec(result =>
+        requires
+            old(self).inv(),
+            spec_layout_size(layout) > 0,
```
jaylorch: Because this code is called by unverified code, we should keep preconditions to a minimum. That means getting rid of this `spec_layout_size` condition. Instead, have the postconditions reflect that it might return `Err` because of a problem with the layout size. This may necessitate an extra runtime check.

src/kernel/src/mm/kheap.rs
```diff
@@ -188,10 +362,20 @@ impl Kheap {
...
         r
     }
 }
 
+//==================================================================================================
+// GlobalAlloc
+//==================================================================================================
+
+#[verus_verify(external)]
 unsafe impl GlobalAlloc for ArenaAllocator {
```
jaylorch, line 383: You need not mark this and the subsequent two items in this file as `#[verus_verify(external)]`. They're not called by verified code, as far as I can tell.

src/kernel/src/mm/kheap.spec.rs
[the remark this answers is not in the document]
tangruize: resolved by c9a644cad0dda78b98de95cad120f431007a37a2

src/kernel/src/mm/kheap.rs
tangruize: commit: e2166de24b8ea3eaad3703bcb7f60ccc3e19d16f 
`can_allocate_layout` has been moved to `kheap.spec.rs` as a `pub closed spec fn`. However, making it `pub open spec` would expose internal slab tier details (`self.slab_8_bytes@.free_addrs, etc.`) Would that be acceptable?

src/kernel/src/mm/kheap.rs
tangruize: Attribute mode doesn't seem to support `Ghost(Map::<int, nat>::empty())` in struct initializers:

```txt
error[E0423]: expected function, tuple struct or tuple variant, found struct `Ghost`
   --> src/kernel/src/mm/kheap.rs:224:24
    |
224 |             alloc_map: Ghost(Map::<int, nat>::empty()),
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Is there a recommended approach to fix this?

src/kernel/src/mm/kheap.rs
tangruize: moved to kheap.spec.rs using an assume_specification

src/kernel/src/mm/kheap.rs
jaylorch: No, an open spec shouldn't expose internal slab details. It should be open, but not expose internal implementation details.

src/kernel/src/mm/kheap.rs
jaylorch: Yes. Use `proof_with!{ alloc_map: Ghost(Map::<int, nat>::empty())) };` just before the struct initialization instruction.

jaylorch:
Please don't use force-push while the PR is being reviewed. It's hard to see the changes you made.

src/kernel/src/mm/kheap.rs
```diff
@@ -137,35 +221,150 @@ impl Kheap {
...
+                        ptr as int,
+                        final(self)@.allocations[ptr as int],
+                    )
+                    &&& (ptr as int) % (spec_layout_align(layout) as int) == 0
+                },
+                Err(_) => {
+                    &&& final(self)@ == old(self)@
+                    &&& !old(self).can_allocate_layout(layout)
+                        || spec_layout_align(layout) > spec_layout_size(layout)
+                        || spec_layout_align(layout) > 512
```
jaylorch: I think this should be `spec_layout_size(layout) > 512`, not `spec_layout_align(layout) > 512`.

src/kernel/src/mm/kheap.proof.rs
```diff
@@ -0,0 +1,1248 @@
...
+    assert(6 * slab_size <= isize::MAX as usize) by (nonlinear_arith)
+        requires
+            slab_size == size / NUM_OF_SLABS,
+            size <= isize::MAX as usize,
+            NUM_OF_SLABS == 7,
+            size as int % NUM_OF_SLABS as int == 0;
+}
+
+/// Whether `t` is one of the supported slab tier sizes.
+closed spec fn is_supported_tier(t: usize) -> bool {
```
jaylorch: I don't see why this needs to be closed. It's not pub anymore.

src/kernel/src/mm/kheap.rs
```diff
@@ -74,8 +96,61 @@ static mut ALLOCATOR: ArenaAllocator = ArenaAllocator;
 // Implementations
 //==================================================================================================
 
+// VERUS DEVIATION: `addr as *mut u8` is unsupported in verified bodies.
+// Keep the executable cast isolated behind a specified helper.
+#[inline]
+fn usize_to_mut_ptr(addr: usize) -> *mut u8 {
```
jaylorch: Shouldn't this be marked `unsafe`?

src/kernel/src/mm/kheap.rs
jaylorch: Sorry that I wasn't clear about what I meant. I meant to move the entire body of the function, including its spec, into `kheap.spec.rs`.

src/kernel/src/mm/kheap.rs
jaylorch: The error condition that used to be a precondition and now is a runtime check isn't reflected in the postcondition. Indeed, the postcondition on `deallocate` now has a very weak failure case.

src/kernel/src/mm/kheap.rs
tangruize: Fixed: 15843b9506a54ee9dc244a141c3e0fe726871c29

src/kernel/src/mm/kheap.rs
tangruize: Making `can_allocate_layout` open without exposing slab internals is difficult since `KheapView` doesn't contain capacity information. Adding capacity to the view would tie the abstraction to slab. Could we just remove this clause from Err as caller does not need to know it.

src/kernel/src/mm/kheap.rs
tangruize: `usize_to_mut_ptr` is an exec function, but `kheap.spec.rs` is cfg-gated, moving the entire body would break the build. @ppenna also suggested moving it to some utils crate in https://github.com/nanvix/nanvix/pull/2373#discussion_r3260560253, would that be a better approach?

src/kernel/src/mm/kheap.rs
tangruize: fixed: f92976551c7c78379d5f511de1d12fd54c67723d

src/kernel/src/mm/kheap.rs
tangruize: fixed by 884b5a1488a7da421a34dfdfb8b54d4d09146aba (slab: addr either in `allocated_addrs` or `free_addrs`) and 2b472274ec0cdffdaf63e22e2dbcd3b5d14d3b91

src/kernel/src/mm/kheap.rs
tangruize: If we add `!old(self)@.allocations.dom().contains(ptr as int)`, it is unprovable as a caller could pass a layout that routes to a different slab tier than the original allocation. Is there a better way to abstract this?

src/kernel/src/mm/kheap.proof.rs
tangruize: removed `closed`: 13754b98e

src/kernel/src/mm/kheap.rs
tangruize: Fixed in 613730b. Note that this `unsafe` isn't required by Rust but added for semantic clarity.

src/kernel/src/mm/kheap.rs
```diff
@@ -137,35 +212,150 @@ impl Kheap {
...
+                        ptr as int,
+                        final(self)@.allocations[ptr as int],
+                    )
+                    &&& (ptr as int) % (spec_layout_align(layout) as int) == 0
+                },
+                Err(_) => {
+                    &&& final(self)@ == old(self)@
+                    &&& !old(self).can_allocate_layout(layout)
+                        || spec_layout_align(layout) > spec_layout_size(layout)
+                        || spec_layout_align(layout) > 512
```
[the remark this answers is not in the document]
tangruize: Resolved (false positive)

src/kernel/src/mm/kheap.rs
```diff
@@ -53,6 +71,10 @@ struct Kheap {
     slab_128_bytes: Slab,
     slab_256_bytes: Slab,
     slab_512_bytes: Slab,
+    // Ghost state — alloc_map is the ground truth for abstract KheapView.
+    // Ghost<T> is zero-sized; it costs nothing at runtime.
+    #[cfg(verus_keep_ghost_body)]
```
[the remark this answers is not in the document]
tangruize: Resolved (false positve)

src/kernel/src/mm/kheap.rs
```diff
@@ -137,35 +212,154 @@ impl Kheap {
...
+                        ptr as int,
+                        final(self)@.allocations[ptr as int],
+                    )
+                    &&& (ptr as int) % (spec_layout_align(layout) as int) == 0
+                },
+                Err(_) => {
+                    &&& final(self)@ == old(self)@
+                    &&& !(old(self)@.allocations =~= Map::<int, nat>::empty())
+                        || spec_layout_align(layout) > spec_layout_size(layout)
+                        || spec_layout_size(layout) > 512
```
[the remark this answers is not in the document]
tangruize: false postive

src/kernel/src/mm/kheap.spec.rs
```diff
@@ -0,0 +1,122 @@
...
+
+// Error constructor — preserves its `code`/`reason` arguments.
+pub assume_specification[ sys::error::Error::new ](
+    code: sys::error::ErrorCode,
+    reason: &'static str,
+) -> (result: sys::error::Error)
+    ensures
+        result.code == code,
+        result.reason == reason,
+;
```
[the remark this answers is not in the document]
tangruize: Fixed by eaa6b1bc5c7c3c15cab6b1b6a1ed200e7675fb5c

src/kernel/src/mm/kheap.rs
tangruize: Changed alloc_map to store layout size instead of block size and strengthened the deallocate Err postcondition: 6311bcdea

src/kernel/src/mm/kheap.rs
```diff
@@ -18,29 +29,28 @@ use ::sys::error::{
...
-/// Number of slabs per slab size. Each slab size is allocated a fixed number of slabs.
-pub(super) const SLAB_COUNT: usize = 32;
-/// Minimum size of a single slab in bytes. It is calculated as the number of slabs per slab size
-/// multiplied by the page size.
-const MIN_SLAB_SIZE: usize = SLAB_COUNT * mem::PAGE_SIZE;
-/// Minimum heap size in bytes. This is the minimum size of the backing storage that must be
-/// provided to initialize the heap.
-pub(crate) const MIN_HEAP_SIZE: usize = NUM_OF_SLABS * MIN_SLAB_SIZE;
+#[verus_verify]
+pub const NUM_OF_SLABS: usize = 7;
```
jaylorch: Why did `(super)` get deleted here?

src/kernel/src/mm/kheap.rs
```diff
@@ -18,29 +29,28 @@ use ::sys::error::{
...
+
+#[verus_verify]
+pub const SLAB_COUNT: usize = 32;
+#[verus_verify]
+pub const MIN_SLAB_SIZE: usize = SLAB_COUNT * mem::PAGE_SIZE;
+#[verus_verify]
+pub const MIN_HEAP_SIZE: usize = NUM_OF_SLABS * MIN_SLAB_SIZE;
```
jaylorch: Why did `(crate)` get deleted here?

src/kernel/src/mm/kheap.rs
```diff
@@ -81,8 +96,51 @@ static mut ALLOCATOR: ArenaAllocator = ArenaAllocator;
...
+            match result {
+                Ok(kheap) => {
+                    &&& kheap.inv()
+                    &&& kheap@ == KheapView::new()
+                },
+                Err(e) => e.code == ErrorCode::InvalidArgument,
+            },
+    )]
     unsafe fn from_raw_parts(addr: usize, size: usize) -> Result<Kheap, Error> {
+        // Check if start address is zero.
```
jaylorch, line 131: I think all these changes to the actual executable code should be part of a separate PR. We should first change the executable in one PR, then add the proof in another.

src/kernel/src/mm/kheap.rs
```diff
@@ -106,9 +164,19 @@ impl Kheap {
...
         let heap_start_addr: *mut u8 = usize_to_mut_ptr(addr);
         let slab_size: usize = size / NUM_OF_SLABS;
+
+        // `size` is a multiple of `MIN_HEAP_SIZE = NUM_OF_SLABS * MIN_SLAB_SIZE`,
+        // so it's also a multiple of `NUM_OF_SLABS`, giving `slab_size * NUM == size`.
+        proof! {
+            lemma_from_raw_parts_ptr_preconditions(addr, size, slab_size, heap_start_addr);
+        }
+
+        #[cfg(not(verus_keep_ghost))]
```
jaylorch: Rather than putting `#[cfg(not(verus_keep_ghost))]` on every call to `info!`, can we change the `info!` macro to include `#[cfg(not(verus_keep_ghost))]`?

src/kernel/src/mm/kheap.rs
```diff
@@ -144,12 +212,53 @@ impl Kheap {
...
+                    &&& !(old(self)@.allocations == Map::<int, nat>::empty())
+                        || spec_layout_align(layout) > spec_layout_size(layout)
+                        || spec_layout_size(layout) > 512
+                },
+            },
+    )]
     unsafe fn allocate(&mut self, layout: Layout) -> Result<*mut u8, AllocError> {
-        let tier: SlabSize = Kheap::layout_to_allocator(&layout)?;
-        let r: Result<*mut u8, AllocError> = match tier {
+        // Reject layouts where alignment exceeds size or the maximum slab tier (512 bytes).
```
jaylorch: It seems that changes to the executable code should be part of a separate PR.

src/kernel/src/mm/kheap.rs
```diff
@@ -160,13 +269,60 @@ impl Kheap {
         };
         #[allow(unused_variables)]
         let align = layout.align();
+        proof! {
```
jaylorch, line 272: This is a lot of inline proof. Pedro will be unhappy. :-O I suggest refactoring all (or nearly all) of this into a lemma and putting that lemma in `kheap.proof.rs`.

src/kernel/src/mm/kheap.rs
jaylorch: Sure, that sounds like a good approach. But this PR doesn't do that yet.

src/kernel/src/mm/kheap.proof.rs
```diff
@@ -0,0 +1,1447 @@
...
+        SlabSize::Slab256 => 256,
+        SlabSize::Slab512 => 512,
+    }
+}
+
+impl Kheap {
+    /// Public invariant: abstract well-formedness + slab-level structural facts.
+    pub open spec fn inv(&self) -> bool {
+        &&& self@.inv()
+        &&& self.internal_inv()
```
arsene1995: why is closed spec internal_inv part of the open spec?

jaylorch: Because the public spec says, "The implementation has an internal invariant that each method maintains. You don't have to worry about, or even know, what that internal invariant is."

That said, I think `inv` should be in `kheap.spec.rs`, not `kheap.proof.rs`. The `internal_inv` belongs in `kheap.proof.rs`, since it doesn't need to be audited.

src/kernel/src/mm/kheap.rs
tangruize: ooo, this is a bit weird. I finally found that the `(super)` and `(crate)` below were introduced by commit https://github.com/nanvix/nanvix/pull/2373/changes/2f9f6ad787e56811d953246fe686f65f7982333e in PR https://github.com/nanvix/nanvix/pull/2373. However, that commit changed visibility and was not checked together with Verus proofs. If I restore that visibility, Verus reports:

```txt
error: in pub open spec function, cannot refer to private const item
  --> src/kernel/src/mm/kheap.rs:45:41
   |
45 | pub(crate) const MIN_HEAP_SIZE: usize = NUM_OF_SLABS * MIN_SLAB_SIZE;
   |                                         ^^^^^^^^^^^^
```

The reason seems to be that `NUM_OF_SLABS` is private, but used by the `pub(crate) const MIN_HEAP_SIZE`. Rust seems to allow a const to be initiazlied by a less-visible consts, while Verus rejects this.

I am not sure what's the best visiblity should be here, @ppenna do you have a suggestion?

src/kernel/src/mm/kheap.rs
tangruize: See above

src/kernel/src/mm/kheap.rs
tangruize: I agree. Will do when this PR is ready.

src/kernel/src/mm/kheap.rs
tangruize: I agree.

src/kernel/src/mm/kheap.rs
tangruize: Good point, I gated the logging macros (`info!`, `trace!`, `debug!`, `warn!` and `error!`) in https://github.com/nanvix/nanvix/pull/2374/commits/9b1a532bf4fc15771d9af93149a1e541b6a03436. Much cleaner now!

src/kernel/src/mm/kheap.rs
tangruize: moved most of the proof to `kheap.proof.rs`: c77a39d82d86cca2e4a2d80c64d990a0df31c5a7. I could not put everything in a single lemma because the inline part updated the ghost variable `alloc_map`, verus reports the following message if trying to move that:

```txt
error: cannot mutate through a spec-mode mutable reference
   --> src/kernel/src/mm/kheap.proof.rs:693:6
    |
693 |     *post.alloc_map = post.alloc_map@.insert(ptr as int, size as nat);
    |      ^^^^
```

Do you have any suggestions for handling this?

src/kernel/src/mm/kheap.proof.rs
tangruize: Kheap `inv` moved to spec: e51e4777979bcb68a9c8d05a970bbc27fe8e46b6

src/kernel/src/mm/kheap.rs
jaylorch: Verus allows a const to be initialized using less-visible ones. And Verus correctly forbids you from having its spec depend on less-visible specs. For instance, this is fine in Verus:
```
#[verus_spec(
    ensures
        ATTRIBUTE_PRIVATE_CONSTANT == 1,
)]
const ATTRIBUTE_PRIVATE_CONSTANT: u8 = 1;

#[verus_spec(
    ensures
        ATTRIBUTE_DEPENDENT_CONSTANT > 0,
)]
pub const ATTRIBUTE_DEPENDENT_CONSTANT: u8 =
    if ATTRIBUTE_PRIVATE_CONSTANT < u8::MAX {
        ATTRIBUTE_PRIVATE_CONSTANT + 1
    }
    else {
        u8::MAX
    };
```

tangruize: Thanks! This if helpful, I will update this.

tangruize: Sorry, I still couldn't get this method to work. It turns out `verus_spec` makes the const exec-only, so it can't be used in specs, the proofs fail with:

```
error: cannot read const with mode exec
    --> src/kernel/src/mm/kheap.proof.rs:1614:17
1614 |         size >= MIN_HEAP_SIZE,
     |                 ^^^^^^^^^^^^^
```

I finally used `verus! {}` directly and making `MIN_HEAP_SIZE` a `closed` const, which works. I think there's still an issue with the `verus_verify` attributes, it does not seem to expose the `open`/`closed` visibility modifier.

tangruize:
This PR is stale, new PR is #2539.

PR #2382, opened 2026-05-19
[slab] Add free_addrs postcondition to from_raw_parts

tangruize:
## Summary

`Slab::from_raw_parts` previously did not specify the initial state of `free_addrs` in its postcondition — only `allocated_addrs == Set::empty()` was guaranteed. This gap was originally identified by Tianyu.

Since `kheap` already relies heavily on `free_addrs` in its specs and proofs, removing `free_addrs` from `SlabView` would require significant rewrites. Instead, this PR strengthens the `from_raw_parts` postcondition to specify that all valid block addresses are initially in `free_addrs`.


ppenna (approved):
LGTM, I will wait confirmation from @arsene1995 / @jaylorch

jaylorch (approved):
Thanks!

PR #2461, opened 2026-06-05
[hostfsd] Follow stat over hostfs paths

ppenna:
## Summary

Path-based `stat()` / `fstatat()` (without `AT_SYMLINK_NOFOLLOW`) over hostfs paths previously returned `OperationNotSupported`. This adds a path-based **following-stat** operation so that following stat now works over hostfs.

## Details

- The follow branch resolves symlinks within the sandbox and follows the final path component, mirroring the existing `lstat` path but using `metadata()` instead of `symlink_metadata()`.
- The new operation reuses the existing `lstat` wire format (`LstatRequest` / `LongLstatRequest` / `LstatResponse`) and the `lstat` completion path, so only the host-side resolution differs. This keeps the change small and avoids a new response decoder.
- The no-follow branch (`AT_SYMLINK_NOFOLLOW`) is unchanged and continues to use `lstat`.

## Tests

Adds hostfsd unit tests covering: regular files, directories, symlinks to files, symlinks to directories, dangling symlinks (host `ENOENT`), and nonexistent paths.

### DIFF
--- a/src/daemons/hostfsd/src/handler.rs
+++ b/src/daemons/hostfsd/src/handler.rs
@@ -170,7 +170,8 @@ impl HostFsHandler {
             | SystemCallMessageHeader::HostFsRmdirRequestPart
             | SystemCallMessageHeader::HostFsSymlinkRequestPart
             | SystemCallMessageHeader::HostFsReadlinkRequestPart
-            | SystemCallMessageHeader::HostFsLstatRequestPart => {
+            | SystemCallMessageHeader::HostFsLstatRequestPart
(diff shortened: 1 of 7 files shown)


ppenna:
Re: the Copilot review note about `SystemCallMessageHeader` implicit discriminants (https://github.com/nanvix/nanvix/pull/2461#discussion_r3360103452) — this is safe in the current single-source-tree build (all components share the same `syscall` crate, and the headers match existing insertion convention). Tracking the optional hardening (explicit discriminant values) separately in #2466.

PR #2515, opened 2026-06-11
[kernel] Handle already-CoW parent pages on re-fork

### LINKED ISSUE #2512: [kernel] Re-forking a CoW page maps it read-only without the CoW bit
## Summary

Copy-on-write (CoW) is broken for **chained forks**: when a process that was itself created by `fork` forks again *before* writing to a shared writable page, the new child receives a plain read-only mapping with **no** CoW bit. A subsequent write from that child is treated as a genuine protection fault (process killed) instead of being resolved as a copy-on-write fault.

## Affected code

- `VirtMemoryManager::link_user_pages` and `VirtMemoryManager::link_one_user_page` in `src/kernel/src/mm/virt/manager.rs`.

The selection filter snapshots the parent PTE's writability directly from the hardware bit:

```rust
buf[count].write((vaddr, frame, pte.flags().is_writable()));
```

and `link_one_user_page` only installs CoW when `writable == true`.

## Root cause

The filter conflates *hardware* read-only with *logically* read-only. A page that was made CoW by a previous fork is stored as `RO + CoW` (the writable bit is cleared by `mark_cow`). On the next fork, `pte.flags().is_writable()` returns `false` for such a page, so it is classified as read-only and shared as a plain `RDONLY` mapping with no CoW bit on either side.

## Reproduction (logical trace)

1. **Fork 1**: parent `P` has a writable page (frame refcount = 1).
   - Filter sees `is_writable() == true` → `writable = true`.
   - Child `C1` mapped `RDWR`; `P` and `C1` marked `RO + CoW`. Refcount = 2.
2. **Fork 2**: `P` forks again *before* any write.
   - Filter reads `P`'s PTE, now `RO + CoW` → `is_writable() == false` → `writable = false`.
   - `C2` gets a plain `RDONLY` mapping, **no CoW bit**; `P` left untouched. Refcount = 3.
3. **`C2` writes** to the page.
   - Fault has `P=1, W=1, U=1`. `resolve_cow_at` finds the PTE but `!pte.is_cow()` → returns `Ok(false)`.
   - The write is forwarded as a protection fault and the process is killed, even though `C2` forked a logically-writable page and should receive a private copy.

## Expected behavior

A page that is `RO + CoW` is logically writable. Forking it again should share it CoW with the new child (child mapped CoW, parent left CoW), so a later write from any sharer triggers a normal copy-on-write resolution.

## Suggested fix

Treat a page as logically writable when it is writable **or** already CoW:

```rust
let writable = pte.flags().is_writable() || pte.is_cow();
```

`link_one_user_page` must then handle an already-CoW parent specially:
- The child must still be mapped `RDWR` and then CoW-marked.
- The parent's `mark_user_page_cow` must be **skipped** when the parent PTE is already CoW (`mark_cow` errors on a non-writable PTE).

The rollback path (`rollback_linked_pages`) must also avoid unmarking a parent page it did not newly mark, since the parent may have been CoW before this `link_user_pages` call. Passing through whether the parent was newly marked (rather than inferring it solely from the child's CoW bit) would make rollback correct for the re-fork case.

## Impact

Any workload that forks a child which then forks again (grandchild creation) before writing to shared writable pages will see the grandchild crash on its first write to such a page.

ppenna:
## Summary

Fixes a fatal page fault when a process is forked more than once and a shared
page is re-forked while still copy-on-write (CoW) from an earlier fork
(issue #2512).

Such pages are read-only in hardware with the AVL CoW bit set. The previous
link logic snapshotted only the parent PTE's hardware writable bit, so it
classified the already-CoW page as a plain read-only mapping and shared it with
the new child without arming copy-on-write. The child's first write was then
misclassified as a genuine protection fault and the process was killed instead
of taking a private copy.

### Kernel (`src/kernel/src/mm/virt/manager.rs`)

- Treat a mapping as logically writable when it is writable in hardware *or*
  already copy-on-write, and carry the parent's current CoW state through the
  link snapshot.
- In `link_one_user_page`, map the child RDWR and arm CoW on it, but mark the
  parent CoW only when it is not already CoW (`mark_cow` requires a writable
  PTE). The partial-rollback path mirrors this: only unmark the parent if this
  call installed the mark.
- In `rollback_linked_pages`, unmap the child pages this call linked, restricted
  to addresses that still have a present mapping in `parent` so unrelated
  pre-existing child mappings are not torn down. Parent CoW marks are
  intentionally **left untouched** during rollback: a newly-marked parent cannot
  be reliably told apart from one that was already CoW before this call (PTE
  state is ambiguous and the shared frame's refcount is not a reliable signal
  either), and wrongly unmarking a page another sharer relies on would break its
  copy-on-write. Leaving a page CoW that could have been writable costs at most
  one extra CoW fault on a rare rollback path, so never unmarking is the safe
  choice.
- Add `LinkUserMapping`, `LinkUserMappingSlot`, and `LinkUserMappingBuf` type
  aliases and expand the surrounding doc comments.

### Tests (`src/tests/test-kernel/src/duplicate.rs`)

- Add `test_duplicate_cow_refork`, a regression test that:
  - Primes a shared byte while the page is private.
  - Forks a first child (kept blocked in `recv()`) to turn the page CoW in the
    parent.
  - Forks a second child to re-fork the now-CoW page.
  - Has the parent take a private copy, then releases the second child, which
    must observe the pre-fork value, write via a CoW fault rather than being
    killed, and report back.
  - Asserts both CoW isolation invariants. With the bug present, the second
    child is killed on its first write and never replies, so the test fails by
    timing out.

## Issues

Closes #2512

### DIFF
--- a/src/kernel/src/mm/virt/manager.rs
+++ b/src/kernel/src/mm/virt/manager.rs
@@ -74,6 +74,16 @@ const ORDER: Ordering = Ordering::Relaxed;
 /// parent's user mappings to a small number of passes for typical user processes.
 const LINK_CHUNK: usize = 32;
 
+/// Snapshot entry for one parent user mapping consumed by
+/// [`VirtMemoryManager::link_user_pages`].
+type LinkUserMapping = (PageAligned<VirtualAddress>, FrameAddress, bool, bool);
+
+/// Uninitialized slot in a link snapshot chunk.
+type LinkUserMappingSlot = MaybeUninit<LinkUserMapping>;
+
+/// Fixed-size chunk buffer used while linking user pages.
+type LinkUserMappingBuf = [LinkUserMappingSlot; LINK_CHUNK];
+
 //==================================================================================================
 // Global Variables
 //==================================================================================================
@@ -237,14 +247,18 @@ impl VirtMemoryManager {
     ///
     /// For each present user mapping in `parent`, this function adds a new reference to
     /// the underlying physical frame and installs the same mapping in `child` at the same
-    /// virtual address. If the parent mapping is writable, both the parent and child PTEs
-    /// are then marked copy-on-write (read-only in hardware + AVL bit set), so that the
-    /// first write from either side triggers a page fault. The in-kernel page-fault
-    /// handler resolves such faults by allocating a private frame and pointing the
-    /// faulting PTE at it.
+    /// virtual address. A mapping is treated as *logically writable* when it is writable in
+    /// hardware or already copy-on-write (the latter arises when the page was shared
+    /// writable by an earlier fork, leaving it read-only with the AVL CoW bit set). For a
+    /// logically-writable mapping the child is mapped RDWR and then marked copy-on-write,
+    /// and the parent is marked copy-on-write too unless it already was, so that the first
+    /// write from any sharer triggers a page fault. The in-kernel page-fault handler
+    /// resolves such faults by allocating a private frame and pointing the faulting PTE at
+    /// it.
     ///
-    /// Read-only mappings (e.g. the text segment) are shared without the copy-on-write
-    /// bit set: a write to them must continue to fault as a regular protection fault.
+    /// Genuinely read-only mappings (e.g. the text segment) are shared without the
+    /// copy-on-write bit set: a write to them must continue to fault as a regular
+    /// protection fault.
     ///
     /// # Parameters
     ///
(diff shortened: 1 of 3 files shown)


src/kernel/src/mm/virt/manager.rs
```diff
@@ -400,13 +441,35 @@ impl VirtMemoryManager {
...
-                    if let Err(re) = parent.unmark_user_page_cow(vaddr) {
-                        warn!(
-                            "link_user_pages(): rollback unmark failed (vaddr={vaddr:?}, \
-                             error={re:?})"
-                        );
+                    // Unmark the parent only if it will become the sole owner of the frame
+                    // once the child's reference is released. The frame is currently held by
+                    // both the parent and the child (refcount >= 2). A refcount of exactly 2
+                    // means this call newly marked the parent, which must be restored to
+                    // writable; a higher count means the parent shared the frame with an
```
jaylorch: I don't think a `refcount` of exactly 2 necessarily means that this call newly marked the parent. Another possible explanation is that it was already 2 before the fork (because of a different child) and this child never got around to linking to it (i.e., the error that triggered this rollback occurred before the child linked this particular page).

jaylorch: A possible fix to this issue is to just never unmark any pages as CoW during such a rollback. It'll have minor performance implications, especially because this rollback will very rarely occur.

ppenna: Good catch.

PR #2555, opened 2026-06-15
[kernel] Fix panic on top-of-space aligned frame address

tangruize:
## Root cause

FrameNumber::MAX is defined as `mem::MAX_ADDRESS / mem::FRAME_SIZE - 1` (note the trailing -1), so the highest representable frame number is one short of the last frame-aligned slot in the address space. The Inner bitmap routines converted an incoming PhysicalAddress to its frame index via `addr.into_frame_number().into_raw_value()`, and `into_frame_number()` is checked: it calls `FrameNumber::from_raw_value(addr >> FRAME_SHIFT)` and `unwrap()`s the result. A page-aligned address at the very top of the address space maps to frame index `FrameNumber::MAX + 1`, for which `from_raw_value()` returns None, so the `unwrap()` panics.

## Consequence

The panic fires *before* the existing range guards in alloc/free/clear/set/is_allocated/coverage paths, so a top-of-space aligned address that those guards are meant to reject out-of-range instead crashes the kernel. This is a boundary-only denial-of-service; the allocator's normal semantics for in-range frames are unaffected.

## How it was identified

Found while verifying the physical-memory allocator with Verus (spec-first proving phase). The specification of `into_frame_number()` carries a `requires spec_frame_number(self@) <= spec_max_frame_number()` precondition modelling the `unwrap()`. The prover could not discharge that precondition at these call sites, pinpointing the off-by-one panic on the top-of-space boundary that the surrounding guards were assumed to cover.

## Fix

Compute the frame index directly with `addr.into_raw_value() / mem::FRAME_SIZE` instead of the checked `into_frame_number()`. For every in-range frame the two expressions are value-identical, but the division never panics, so an out-of-range index now flows into the existing guards (bitmap bounds checks / coverage comparisons) and is rejected gracefully. Applied to all six Inner bitmap sites (alloc, free, clear, set, is_allocated, and the region coverage helper).

### DIFF
--- a/src/kernel/src/hal/mem/types/address/phys.rs
+++ b/src/kernel/src/hal/mem/types/address/phys.rs
@@ -98,7 +98,9 @@ impl PhysicalAddress {
     pub fn into_frame_number(self) -> FrameNumber {
         let raw_addr: usize = self.0.into_raw_value();
         let frame_number: usize = raw_addr >> mem::FRAME_SHIFT;
-        // Safety: the following unwrap is safe because a physical address has a valid frame number.
+        // The unwrap below never panics: `FrameNumber::MAX` is the number of the frame that
+        // contains `MAX_ADDRESS`, so `raw_addr >> FRAME_SHIFT <= FrameNumber::MAX` holds for
+        // every address in the space.
         FrameNumber::from_raw_value(frame_number).unwrap()
     }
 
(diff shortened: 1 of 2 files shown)


ppenna (approved):
LGTM please rebase on latest `dev` and squash commits into a single one using a detailed commit message.

PR #2584, opened 2026-06-17
[vfs] Add pipe buffer and FD layer for POSIX unnamed pipes

ppenna:
## Summary

Introduces the VFS-library foundation for `pipe(2)`. All pipe buffer state lives in the `no_std` `vfs` crate next to the existing FD machinery; the daemon (vfsd) owns no buffer bytes and drives blocking from its side via the non-blocking primitives added here.

This PR is the **library layer only**. The vfsd daemon, syscall, and test consumers will follow in subsequent commits/PRs.

## Changes

**New module — `src/libs/vfs/src/pipe.rs`**
- `PipeInner`: a fixed-capacity (64 KiB) byte ring buffer plus reader/writer reference counts.
- `PipeEnd`: one end of a pipe over a shared `Arc<Mutex<PipeInner>>`, with an `is_write` direction flag and a copied-in `pipe_id` for lockless identity lookups. `Drop` decrements the reader/writer count; waking suspended callers is left to the daemon, which owns the blocked-request queues.
- Non-blocking outcomes (`PipeReadOutcome`/`PipeWriteOutcome`) modelling `Read`/`WouldBlock`/`Eof` and `Wrote`/`WouldBlock`/`BrokenPipe`, with `PIPE_BUF` (4 KiB) write atomicity and direction enforcement.
- Process-global monotonic pipe-id allocator so stale wakeups can never target the wrong pipe.

**FD integration — `src/libs/vfs/src/fd.rs`**
- Add `VfsFileHandle::Pipe` and route the read/write/seek/size/ftruncate/fallocate/fsync handlers to the appropriate not-supported results.
- Add pipe FD helpers: `vfs_pipe`, `vfs_pipe_read`, `vfs_pipe_write`, `vfs_pipe_id`, `vfs_pipe_is_last_ref`, `vfs_get_status_flags`.
- Track per-description open file status flags; implement `fcntl` `F_GETFL`/`F_SETFL` for the mutable subset (`O_NONBLOCK`) and accept `F_GETFD`/`F_SETFD` as no-ops. `vfs_fcntl` now takes the `F_SETFL` argument.
- Report pipe (`S_IFIFO`) metadata from `fstat` with a synthetic pipefs device id distinct from regular VFS files.
- Extend `vfs_process_exit` to return `ProcessExitReclaim`, surfacing both orphaned hostfs fds and pipe count-to-zero closures so the daemon can fire EOF/`EPIPE` wakeups for suspended counterparts on process exit.

## Behavioral notes
- Zero-length writes succeed with `Wrote(0)` even when no readers remain, matching Linux (no `EPIPE`/`SIGPIPE` for empty writes).

## Testing
15 new unit tests for the buffer (round-trip, ring wrap-around, EOF, broken pipe, `PIPE_BUF` atomicity, partial large writes, direction, identity) and the FD layer (allocation, outcome matrix, EOF/broken pipe, `O_NONBLOCK` round-trip, exit closures).

## Issues

https://github.com/nanvix/nanvix/issues/343

### DIFF
--- a/src/daemons/vfsd/src/handler/short.rs
+++ b/src/daemons/vfsd/src/handler/short.rs
@@ -132,7 +132,8 @@ pub(crate) fn handle_fcntl(source: ThreadIdentifier, msg: SystemCallMessage) -> 
     let req: FileControlRequest = FileControlRequest::from_bytes(msg.payload);
     let fd: i32 = req.fd;
     let cmd: i32 = req.cmd;
-    match ::vfs::fd::vfs_fcntl(fd, cmd) {
+    let arg: i32 = req.arg;
+    match ::vfs::fd::vfs_fcntl(fd, cmd, arg) {
         Ok(ret) => {
             FileControlResponse::build(source, ret, ProcessIdentifier::VFSD, MessageType::Ipc)
         },
--- a/src/libs/vfs/src/fd.rs
+++ b/src/libs/vfs/src/fd.rs
@@ -42,6 +42,7 @@ use ::sysapi::{
     fcntl::{
         file_control_request,
         file_creation_flags,
+        file_status_flags,
     },
     ffi::c_int,
     sys_stat::{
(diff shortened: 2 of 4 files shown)


ppenna:
Superseded by #2587

PR #2623, opened 2026-06-19
[kernel] Authoritative proc role in sched event

### LINKED ISSUE #2616: [proc] Make process role and parent authoritative in termination events
## Context

procd reconstructs a process's *role* (init / daemon / forked child) by inference, because the
kernel's termination event is identity-poor. `ProcessTerminationInfo` carries only `{ pid, status }`
(`src/libs/sys/src/sys/event/scheduling/termination.rs`), so on a termination procd cannot tell what
the process was and must guess from prior state using several fragile heuristics:

- `init_proc`, which is only set if the process explicitly signed up (most workloads never do);
- name-based `is_daemon()` (`src/libs/proc/src/daemon/mod.rs`), whose name is empty until the
  daemon's `Signup` has been processed;
- a `parent == ProcessIdentifier::KERNEL` lineage fallback; and
- a hardcoded daemon-PID exclusion list `{ PROCD, MEMD, VFSD }`, duplicated in both
  `adoptive_init()` and `finalize_known_termination()`.

These heuristics are unreliable precisely during the race windows the buffering exists to handle
(e.g. a buffered termination replayed before the daemon's `Signup` has an empty name). This is the
root cause of the init/daemon-shutdown bypass (#2541) and of the regression introduced by its point
fix (#2559), where the reworked routing mis-classifies a raced termination and the standalone
`vfs-bench` guest never powers off (45-minute CI hang).

## Proposed Solution

Make role and lineage **authoritative in the event** rather than inferred in procd. The kernel
already knows both at spawn time (it spawns init and the daemons directly and owns the well-known
daemon PIDs).

- Add `parent: ProcessIdentifier` and an explicit `role { Init, Daemon, User }` to
  `ProcessTerminationInfo` (creation already carries `parent`).
- Have the kernel populate these fields authoritatively when it publishes the termination event.
- Collapse procd's termination routing to a single total match:
  `Init => shutdown(status)`, `Daemon => deregister (crash-shutdown if status != 0)`,
  `User => reap`.
- Delete the name-based `is_daemon()` routing, the `parent == KERNEL` fallback, and both copies of
  the `{ PROCD, MEMD, VFSD }` exclusion list.

This eliminates the entire class of "role inferred wrong during a race" defects in one place and
removes the duplicated, drift-prone heuristics.

## Alternative Solutions

- **Keep inference but centralize it** into one helper. Rejected: it remains racy — the underlying
  facts still are not known to procd until other events are processed.
- **Add `role` only, not `parent`.** Insufficient: `parent` is still needed for re-parenting and
  reaping bookkeeping, and shipping both keeps the event fully self-describing.

## Related Issues

- #2618 — companion structural fix (per-process creation-before-termination ordering); complementary to this.
- #2541 — the init/daemon-shutdown bypass this inference causes.
- #2559 — the point fix whose reworked inference regressed `vfs-bench` shutdown.
- #2534, #2535 — sibling procd lifecycle defects from the same reconstruct-from-events design.

ppenna:
## Summary

Have the kernel stamp an authoritative `ProcessRole` (`Init`, `Daemon`, or `User`) into
process-creation and process-termination scheduling events, and route terminations in `procd`
on that role instead of re-inferring a process's identity from prior, race-prone daemon state.

## Motivation

`procd` previously reconstructed whether a terminating process was the init process, a daemon,
or a forked user process from local state (`init_proc` recorded at signup, name-based
`is_daemon` checks, and `parent == KERNEL` lineage fallbacks). That inference was fragile:

- A daemon's name is only known after signup, and `procd` itself never signs up.
- A forked child that signed up before its creation event was observed could be misclassified
  as init and trigger a spurious system-wide shutdown.

The kernel already owns this information authoritatively, since it spawns the init process and
daemons directly and owns the well-known daemon process identifiers.

## Changes

- **sys:** Add `ProcessRole { Init, Daemon, User }` (`repr(u32)`) with a `classify(pid, parent)`
  constructor and `u32` round-trip helpers. An unrecognized raw value decodes to `User` so a
  malformed event cannot spuriously shut the system down.
- **sys:** Add the role to `ProcessCreationInfo` (now 12 bytes) and add the parent and role to
  `ProcessTerminationInfo` (now 16 bytes); extend the native-endian (de)serializers and add
  round-trip / classification unit tests.
- **kernel/pm:** Populate the role at fork/spawn time and build the termination record
  `(pid, status, parent, role)` in `harvest_zombies`; `harvest_zombies` now returns
  `ProcessTerminationInfo`.
- **kernel/event:** Serialize each scheduling notification using its own length (the two
  variants differ in size) and assert each fits within the message payload, replacing the fixed
  `SCHEDULING_INFO_SIZE`.
- **kernel/kcall:** Thread the richer termination record through the idle loop and PROCD
  shutdown detection.
- **proc daemon:** Route terminations on the kernel-supplied role; record the init process from
  its creation event (used only to re-parent orphans); drop the signup-time and lineage-based
  init/daemon heuristics and the tentative-init reconciliation. Consolidate per-termination
  bookkeeping into a `cleanup_terminated()` helper.

## Compatibility

No on-wire message types change; only the scheduling-event payload layouts grow.

## Testing

- `./z build -- run-unit-tests` — pass
- `./z build -- run-nanvix-tests` (standalone integration) — pass

## Issues

Closes #2616

### DIFF
--- a/src/kernel/src/pm/process/manager/mod.rs
+++ b/src/kernel/src/pm/process/manager/mod.rs
@@ -191,6 +191,12 @@ pub struct ProcessManager {
     /// momentarily saturated. Bounded by the number of live processes, which is itself capped at
     /// [`config::kernel::MAX_PROCESSES`].
     pending_terminations: VecDeque<ProcessTerminationInfo>,
+    /// Process identifiers of the system daemons the kernel spawned directly (procd, memd, and the
+    /// like), recognized by program name at spawn time. Used to classify a process's
+    /// [`ProcessRole`] authoritatively even when a daemon occupies a process identifier that another
+    /// deployment would assign to the init workload — e.g. a deployment without a separate VFS
+    /// daemon lets the workload take the VFS daemon's conventional identifier.
+    daemon_pids: Vec<ProcessIdentifier>,
 }
 
 impl ProcessManager {
@@ -232,7 +238,17 @@ impl ProcessManager {
             deferred_exec_vmem: Vec::new(),
             pending_creations: VecDeque::new(),
             pending_terminations: VecDeque::new(),
+            daemon_pids: Vec::new(),
         }
+    }
+
+    /// Classifies the [`ProcessRole`] of a process from its identifier and parent, using the set of
+    /// daemon identifiers recorded at spawn time. A process the kernel spawned as a system daemon is
+    /// always [`ProcessRole::Daemon`], regardless of the identifier it received; a non-daemon
+    /// process spawned by the kernel is the init process; anything else was forked by a user
+    /// process.
+    fn classify_role(&self, pid: ProcessIdentifier, parent: ProcessIdentifier) -> ProcessRole {
+        ProcessRole::classify(parent, self.daemon_pids.contains(&pid))
     }
 
     ///
@@ -575,6 +591,13 @@ impl ProcessManager {
         // NOTE: if we fail beyond this point we need to free page mappings.
         //==============================================================
 
+        // Recognize a kernel-spawned system daemon by its program name (the first token of the
+        // command-line arguments) so it is classified as a daemon regardless of the process
+        // identifier it receives. This keeps daemon identification correct for deployments that
+        // omit a daemon and let the init workload take that daemon's conventional identifier.
+        let is_daemon: bool =
+            ::config::daemons::is_system_daemon(args.split_whitespace().next().unwrap_or(""));
+
         Ok(no_fail!(ProcessIdentifier, {
             // Commit the next process and thread identifiers now that all fallible operations have succeeded.
             self.next_pid = next_pid;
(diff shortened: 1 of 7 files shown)


src/kernel/src/pm/process/manager/mod.rs
```diff
@@ -574,6 +591,13 @@ impl ProcessManager {
         // NOTE: if we fail beyond this point we need to free page mappings.
         //==============================================================
 
+        // Recognize a kernel-spawned system daemon by its program name (the first token of the
+        // command-line arguments) so it is classified as a daemon regardless of the process
+        // identifier it receives. This keeps daemon identification correct for deployments that
+        // omit a daemon and let the init workload take that daemon's conventional identifier.
+        let is_daemon: bool =
+            ::config::daemons::is_system_daemon(args.split_whitespace().next().unwrap_or(""));
```
ppenna, line 599: The kernel needs to ensure that if a system daemon is already running with that name, no other process can claim that name.

Issue #2609, opened 2026-06-18
[networkd] Socket file descriptors are shared, not duplicated, across fork()


esaurez:
## Description

Socket file descriptors are **shared**, not duplicated, across `fork()`. When a
process forks, the child inherits the parent's socket descriptors as the *same*
underlying host socket with no per-process ownership or reference count. As a
result, when the child closes an inherited socket descriptor (explicitly via
`close()`, or implicitly when the child exits), the kernel-side host socket is
torn down for the **parent** as well, breaking the parent's connection.

This violates POSIX `fork()` semantics and breaks the classic "prefork" server
pattern (parent `accept()`s a connection, forks a worker, the worker serves or
exits while the parent keeps listening).

Note that **file** descriptors are handled correctly: `vfsd` duplicates a
child's VFS descriptors on fork (sharing the underlying open file descriptions,
and therefore offsets, as POSIX requires). Only **socket** descriptors managed
by `networkd` are affected.

## Expected Behavior

Per POSIX, the system maintains two levels of state:

- a **per-process file descriptor table** (the small integers), and
- a shared, **reference-counted open file description** (the underlying object —
  here, the host socket).

On `fork()` the child receives a *copy of the descriptor table*; each child
descriptor points at the *same* open file description as the parent's, and that
description's reference count is incremented. `close()` is therefore
**per-process**: it removes only that process's table entry and decrements the
reference count. The other process's descriptor remains fully valid, and the
underlying socket is closed only when the **last** descriptor across **all**
processes is closed. This applies to sockets exactly as it does to regular
files and pipes.

So: a forked child closing (or exiting while holding) an inherited socket must
**not** disturb the parent's socket.

## How to Reproduce

Using the standalone deployment with host networking enabled
(`-allow-host-networking`), run a guest program that:

1. Creates a listening TCP socket and `accept()`s a client connection.
2. `fork()`s.
3. In the child: `close()` the accepted connection descriptor (or simply
   `_exit()`), then exit.
4. In the parent: after `waitpid()`, attempt to `send()`/`write()` a response on
   that same accepted connection.

Observed: the parent's `send()` fails (e.g. `ESHUTDOWN`/`EINVAL`) because the
child's close already tore down the shared host socket. Closing the *listening*
socket in the child instead breaks the parent's `accept()` loop in the same way.

Expected: the parent's `send()` succeeds; the child's close only dropped the
child's reference.

(This was first observed driving a CPython `http.server` on Nanvix: a request
[...]

Issue #354, opened 2024-12-23
[posix] Support `dup2()`


ppenna:
## Description

Implement `dup2()`.

## References

https://pubs.opengroup.org/onlinepubs/9799919799/functions/dup.html

ppenna:
Resolved in `dev` — closing as completed.

`dup2()` is implemented as a C-ABI libc binding in `src/libs/syscall/src/unistd/bindings/dup2.rs`, backed by the flat descriptor table.

Implemented by #2625.

_Identified during a sweep for issues fixed in `dev` but left open; please reopen if you consider this incomplete._

PR #2626, opened 2026-06-19
[tests] Add fork/exec/dup2/socket acceptance tests

esaurez:
## Summary

Adds four standalone Rust regression/acceptance tests under `src/tests/`, ported from the C
reproducers in `nanvix/posix-tests`. They confirm that Rust can express the same operations the
POSIX C tests cover, and they assert the **correct** behavior, so they currently **FAIL or time out**
and will pass once the corresponding capability is implemented/fixed — guarding it thereafter.

> These tests are **red by design** today. Opening as a draft: they pair with the fixes for the
> referenced issues and document the acceptance criteria for them.

| Test crate | What it asserts | Status today |
|---|---|---|
| `socket-fork-rust` | A child closing an inherited socket must not destroy the parent's socket (reference-counting across `fork()`) | **Fails** — panics at the parent's post-`close()` `getsockname()` |
| `pipe-dup2-rust` | `dup2()` must redirect a standard stream onto a pipe | **Fails** — `dup2()` is a no-op stub |
| `fork-exec-vfsd-test` (+ `fork-exec-vfsd-target`) | A `fork()`+`execv()`'d image must be able to do vfsd filesystem I/O | **Hangs/times out** — first vfsd read after exec blocks |

Related issues: `#2609` (socket refcount across fork), `#354` (`dup2`), and the fork+exec+vfsd hang.

## Why these are in Rust

`nanvix/posix-tests` exists for things that are awkward to express in the in-tree Rust harness (e.g.
dynamic loading). These three are **not** in that category — every operation they need has a native
Rust binding in `libsyscall` (`unistd::bindings::fork`, `do_execv`, `fcntl::open`,
`unistd::{read,write,close,pipe}`, `sys::socket::{socket,bind,getsockname}`), with `dup2` reachable
via its C ABI symbol. So they belong here as first-class standalone tests next to `test-fork-kcall`,
`waitpid-rust`, `execv-test`, and `network-rust`.

(The fork-pid-cache and pthread-reinit reproducers from posix-tests are intentionally **not** ported:
they are already covered by `test-fork-kcall` — `test_fork_pid_cache_invalidation` and
`test_fork_mutex_condvar`.)

## What each test does

- **`socket-fork-rust`** — Parent binds an `AF_INET` socket, `fork()`s, the child `close()`s its
  inherited copy and exits; after `waitpid()`, the parent probes its own socket with `getsockname()`.
  Today `networkd` shares the socket (fixed arithmetic guest→host mapping, no per-process refcount),
  so the child's `close()` tears it down for both and the parent's probe fails.
- **`pipe-dup2-rust`** — Forks; the child `dup2()`s a pipe's write end onto its standard output and
  writes a marker; the parent reads the marker back from the pipe. The redirection happens in the
  child so the parent's own stdout (carrying the `"ok"` sentinel) is undisturbed. `dup2()` is a stub
  returning `-1`, so the child reports failure and no marker arrives.
- **`fork-exec-vfsd-test` / `-target`** — The caller forks and the child `execv()`s `/target`
  (bundled into the test's ramfs). The target performs a vfsd read of `/target` and exits; the parent
  waits and then prints `"ok"`. `execv()` gives the image a new main-thread id and the vfsd
  push/pull rendezvous is keyed by `(pid, tid)`, so the target's first vfsd read never matches and
  the test times out.

## Wiring

- Workspace members in `Cargo.toml` (+ `Cargo.lock`).
- `ALL_GUEST_TESTS` and `STANDALONE_GUEST_BINARIES`; `fork-exec-vfsd-target` is added to
  `STANDALONE_RAMFS_ONLY_BINARIES` and bundled as `/target` in `fork-exec-vfsd-test.img`
  (`build/make/nanvixd.mk`, mirroring `execv-test.img`).
- Registered in `test/test-standalone.toml` and `test/test-standalone-windows.toml`
  (`socket-fork-rust` uses `-allow-host-networking`; `fork-exec-vfsd-test` uses
  `-ramfs ./bin/fork-exec-vfsd-test.img`).

## Validation

- All four crates pass `cargo check`, `clippy -D warnings` (incl. `as_conversions`, `unwrap_used`,
  `expect_used`), and `rustfmt --check`.
- A full standalone build produces the `.elf`, `.initrd`, and `fork-exec-vfsd-test.img` artifacts.
- `socket-fork-rust` and `pipe-dup2-rust` were run on standalone Nanvix and panic at their intended
  assertions (`nanvix/nanvix#2609` and `nanvix/nanvix#354` respectively), confirming the native
  operations work and the tests fail for the right reason.


esaurez:
Superseded: split into three independent acceptance-test PRs, one per test — #2627 (socket-across-fork / #2609), #2628 (dup2 redirection / #354), #2629 (fork+exec+vfsd I/O).

PR #2628, opened 2026-06-19
[tests] Add dup2 pipe-redirection acceptance test

esaurez:
## Summary

Adds `pipe-dup2-rust`, a standalone Rust **acceptance test** for `dup2()` redirection of a standard
stream (`nanvix/nanvix#354`), ported from the C reproducer in `nanvix/posix-tests`.

> Red by design today — opened as a draft. It asserts the **correct** behavior, so it fails until
> `dup2()` is implemented (it is currently a no-op stub returning `-1`), after which it passes and
> guards the behavior.

## What it does

The test `fork()`s; the child `dup2()`s a pipe's write end onto its standard output and writes a
marker; the parent reads the marker back from the pipe's read end. The redirection happens in the
child so the parent's own standard output (which carries the `"ok"` sentinel) is never disturbed —
mirroring the `pipe()` + `dup2()` + `exec()` pattern shells use to redirect a child's streams.

Native `pipe`/`read`/`write`/`close`/`fork` bindings are used; `dup2()` is reached through its C ABI
symbol (`libsyscall` has no safe wrapper for it yet, only the stub).

## Wiring

- Workspace member (`Cargo.toml` + `Cargo.lock`).
- `ALL_GUEST_TESTS` and `STANDALONE_GUEST_BINARIES`.
- Registered in `test/test-standalone.toml` and `test/test-standalone-windows.toml`.

## Validation

- `cargo check`, `clippy -D warnings` (incl. `as_conversions`/`unwrap_used`/`expect_used`), and
  `rustfmt --check` pass.
- Run on standalone Nanvix it panics at the intended assertion:
  `child could not redirect standard output via dup2() (... nanvix/nanvix#354)`.

Part of a set of three independent acceptance-test PRs split from the posix-tests reproducers
(socket-across-fork and fork+exec+vfsd I/O are the other two).


esaurez:
Rebased onto the latest dev (now includes the dup2() implementation). Rebuilt standalone and ran the test: it now **passes** (emits the \ok\ sentinel) — the acceptance test went red→green with the dup2() implementation, exactly as intended. Ready to come out of draft whenever you'd like.

PR #2815, opened 2026-07-01
Verify physical memory modules (specs)

tangruize:
Add the Verus specifications for the physical memory modules

### DIFF
--- a/src/kernel/src/hal/mem/types/address/aligned/page.spec.rs
+++ b/src/kernel/src/hal/mem/types/address/aligned/page.spec.rs
@@ -1,0 +1,20 @@
+use crate::hal::mem::spec_page_size;
+
+verus! {
+
+// Success condition for the validating constructor, stated purely on the abstract
+// address value. `from_address` validates, it does NOT normalize: success requires the
+// *input* to already be page-aligned.
+pub open spec fn spec_aligned(addr_view: int) -> bool {
+    addr_view % spec_page_size() == 0
+}
+
+impl<T: Address> PageAligned<T>
+{
+    pub open spec fn inv(&self) -> bool
+    {
+        spec_aligned(self@)
+    }
+}
+
+} // verus!
--- a/src/kernel/src/hal/mem/types/address/frame.rs
+++ b/src/kernel/src/hal/mem/types/address/frame.rs
@@ -5,6 +5,10 @@
 // Imports
 //==================================================================================================
 
+use vstd::prelude::*;
+#[cfg(verus_keep_ghost)]
+include!("frame.spec.rs");
+
 use crate::hal::mem::types::address::{
     Address,
     PageAddress,
@@ -23,6 +27,7 @@ use ::sys::error::Error;
 ///
 /// A type that represents a frame address.
 ///
+#[verus_verify(external_derive)]
 #[derive(Clone, Copy)]
 pub struct FrameAddress(PageAligned<PhysicalAddress>);
 
@@ -42,15 +47,54 @@ impl FrameAddress {
     pub fn into_page_address(self) -> PageAddress {
         PageAddress::new(PageAligned::into_virtual_address(self.0))
     }
+}
 
+// Raw-value and frame-number conversions of a frame address. The frame address denotes a single
+// page-aligned physical frame; these functions expose its two equivalent identities (raw physical
+// address and frame number) and the lossless, mutually-inverse mappings between them.
+#[verus_verify]
+impl FrameAddress {
+    // Constructs a frame address from a frame number. The frame's base address is
+    // `frame_number * PAGE_SIZE`, page-aligned by construction, so the call always succeeds and the
+    // result satisfies `inv()`.
+    #[verus_verify(external_body)]
+    #[verus_spec(result =>
+        ensures
+            result is Ok,
+            (result->Ok_0).inv(),
+            (result->Ok_0)@ == frame_number@ * spec_page_size(),
+    )]
     pub fn from_frame_number(frame_number: FrameNumber) -> Result<Self, Error> {
         Ok(Self(PageAligned::from_address(PhysicalAddress::from(frame_number))?))
     }
 
+    // Recovers the frame number of a frame address (`self@ / PAGE_SIZE`). Requires only that the
+    // address is page-aligned (so the round-trip `from_number(into_frame_number(self)) == self`
+    // holds); representability is automatic because the underlying
+    // `PhysicalAddress::into_frame_number` is total. The result is the exact inverse of
+    // `from_frame_number`.
+    #[verus_verify(external_body)]
+    #[verus_spec(result =>
+        requires
+            self.inv(),
+        ensures
+            result@ == self@ / spec_page_size(),
+            result@ * spec_page_size() == self@,
+    )]
     pub fn into_frame_number(self) -> FrameNumber {
         self.0.into_frame_number()
     }
 
+    // Succeeds only for page-aligned inputs. The contract exposes the validated raw address to
+    // verified callers.
+    #[verus_verify(external_body)]
+    #[verus_spec(result =>
+        ensures
+            match result {
+                Ok(fa) => fa.inv() && fa@ == raw_addr as int,
+                Err(_) => true,
+            },
+    )]
     pub fn from_raw_value(raw_addr: usize) -> Result<Self, Error> {
         Ok(Self(PageAligned::from_address(PhysicalAddress::from_raw_value(raw_addr)?)?))
     }
@@ -64,6 +108,12 @@ impl FrameAddress {
     ///
     /// The raw value of the target [`FrameAddress`].
     ///
+    // Dependency contract: the raw value is the abstract frame address. Verified against the
+    // `PageAligned::into_raw_value` dependency contract.
+    #[verus_spec(result =>
+        ensures
+            result as int == self@,
+    )]
     pub fn into_raw_value(self) -> usize {
         self.0.into_raw_value()
     }
--- a/src/kernel/src/hal/mem/types/address/frame.spec.rs
+++ b/src/kernel/src/hal/mem/types/address/frame.spec.rs
@@ -1,0 +1,25 @@
+verus! {
+
+/// The architectural page size.
+pub open spec fn spec_page_size() -> int {
+    ::arch::mem::PAGE_SIZE as int
+}
+
+impl View for FrameAddress
+{
+    type V = int;
+
+    closed spec fn view(&self) -> int
+    {
+        self.0@
+    }
+}
+
+impl FrameAddress {
+    pub open spec fn inv(&self) -> bool
+    {
+        self@ % spec_page_size() == 0
+    }
+}
+
+}
--- a/src/kernel/src/hal/mem/types/region.rs
+++ b/src/kernel/src/hal/mem/types/region.rs
@@ -5,6 +5,10 @@
 // Imports
 //==================================================================================================
 
+use vstd::prelude::*;
+#[cfg(verus_keep_ghost)]
+include!("region.spec.rs");
+
 use crate::hal::mem::types::{
     access::AccessPermission,
     address::{
@@ -31,6 +35,7 @@ use ::sys::error::{
 /// A type that represents the type of a memory region.
 ///
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[verus_verify]
 pub enum MemoryRegionType {
     /// Usable memory.
     Usable,
@@ -52,6 +57,7 @@ pub enum MemoryRegionType {
 /// Caching policy for MMIO memory regions. Controls the PWT (Page Write-Through) and PCD
 /// (Page Cache Disable) bits in the page table entries that back the region.
 ///
+#[verus_verify(external_derive)]
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub struct MmioCachePolicy {
     /// If `true`, the page is mapped with the Write-Through attribute (PWT=1).
@@ -129,6 +135,7 @@ impl MmioCachePolicy {
 ///
 /// A memory region.
 ///
+#[verus_verify(external_derive)]
 #[derive(Debug, Clone)]
 pub struct MemoryRegion<T: Address> {
     name: String,
@@ -195,11 +202,19 @@ impl<T: Address> MemoryRegion<T> {
     }
 
     /// Returns the first valid address that lies in the target memory region.
+    #[verus_spec(result =>
+        ensures
+            result@ == self@.start,
+    )]
     pub fn start(&self) -> T {
         self.start
     }
 
     /// Returns the size of the target memory region.
+    #[verus_spec(result =>
+        ensures
+            result as int == self@.size,
+    )]
     pub fn size(&self) -> usize {
         self.size
     }
@@ -275,6 +290,7 @@ impl<T: Address> Ord for MemoryRegion<T> {
 ///
 /// A memory region that has been truncated to a multiple of a page size.
 ///
+#[verus_verify(external_derive)]
 #[derive(Clone)]
 pub struct TruncatedMemoryRegion<T: Address>(MemoryRegion<PageAligned<T>>);
 
@@ -346,11 +362,19 @@ impl<T: Address> TruncatedMemoryRegion<T> {
     }
 
     /// Returns the first valid address that lies in the target memory region.
+    #[verus_spec(result =>
+        ensures
+            result@ == self@.start,
+    )]
     pub fn start(&self) -> PageAligned<T> {
         self.0.start()
     }
 
     /// Returns the size of the target memory region.
+    #[verus_spec(result =>
+        ensures
+            result as int == self@.size,
+    )]
     pub fn size(&self) -> usize {
         self.0.size()
     }
--- a/src/kernel/src/mm/phys/frame.rs
+++ b/src/kernel/src/mm/phys/frame.rs
@@ -44,6 +44,13 @@ use ::sys::{
     },
     mm::Address,
 };
+use ::vstd::prelude::*;
+
+#[cfg(verus_keep_ghost)]
+include!("frame.spec.rs");
+
+#[cfg(verus_keep_ghost)]
+include!("frame.proof.rs");
 
 //==================================================================================================
 // Inner
@@ -71,6 +78,7 @@ use ::sys::{
 static mut REFCOUNT_STORAGE: [u8; NFRAMES] = [0; NFRAMES];
 
 /// Private state of the frame allocator singleton.
+#[verus_verify]
 struct Inner {
     /// A bitmap that keeps track of free/used frames.
     bitmap: Bitmap,
@@ -92,6 +100,7 @@ struct Inner {
     refcount: &'static mut [u8],
 }
 
+#[verus_verify]
 impl Inner {
     ///
     /// # Description
@@ -103,6 +112,26 @@ impl Inner {
     /// Upon success, the address of the allocated frame is returned. Upon failure, an error is
     /// returned instead.
     ///
+    #[verus_verify(external_body)]
+    #[verus_spec(result =>
+        requires
+            old(self).inv(),
+        ensures
+            final(self).inv(),
+            match result {
+                Ok(frame) => {
+                    &&& frame.inv()
+                    &&& old(self)@.is_free(frame@)
+                    &&& final(self)@ == FrameAllocView {
+                        refcounts: old(self)@.refcounts.insert(frame@, 1int),
+                    }
+                },
+                Err(_) => {
+                    &&& final(self)@ == old(self)@
+                    &&& old(self)@.no_free_frames()
+                }
+            },
+    )]
     fn alloc(&mut self) -> Result<FrameAddress, Error> {
         let frame_number: usize = match self.bitmap.alloc() {
             Ok(index) => index,
@@ -147,6 +176,36 @@ impl Inner {
     /// Upon success, the base `FrameAddress` of the contiguous range is returned. Upon failure,
     /// an error is returned instead.
     ///
+    #[verus_verify(external_body)]
+    #[verus_spec(result =>
+        requires
+            old(self).inv(),
+            count > 0,
+        ensures
+            final(self).inv(),
+            match result {
+                Ok(base) => {
+                    &&& base.inv()
+                    &&& ({
+                        let frame_indices = Set::range(0, count as int);
+                        let frames = frame_indices.map_by(
+                            |i: int| base@ + i * spec_page_size(),
+                            |addr: int| (addr - base@) / spec_page_size(),
+                        );
+                        &&& old(self)@.all_free(frames)
+                        &&& final(self)@ == FrameAllocView {
+                            refcounts: old(self)@.refcounts.union_prefer_right(
+                                Map::new(frames, |addr: int| 1int)
+                            ),
+                        }
+                    })
+                },
+                Err(_) => {
+                    &&& final(self)@ == old(self)@
+                    &&& !old(self)@.exists_contiguous_free_run(count as int)
+                }
+            },
+    )]
     fn alloc_contiguous(&mut self, count: usize) -> Result<FrameAddress, Error> {
         // Reject impossible requests before delegating to the bitmap range allocator.
         if count > self.bitmap.number_of_bits() {
@@ -197,6 +256,31 @@ impl Inner {
     ///
     /// Upon success, `Ok(())` is returned. Upon failure, an error is returned instead.
     ///
+    #[verus_verify(external_body)]
+    #[verus_spec(result =>
+        requires
+            old(self).inv(),
+            frame.inv(),
+        ensures
+            final(self).inv(),
+            match result {
+                Ok(()) => {
+                    &&& old(self)@.is_allocated(frame@)
+                    // Releasing one reference simply decrements the count. When it
+                    // reaches zero the frame stays covered but becomes free, so the
+                    // map domain is unchanged in both the shared and last-owner cases.
+                    &&& final(self)@ == FrameAllocView {
+                        refcounts: old(self)@.refcounts.insert(
+                            frame@, old(self)@.refcounts[frame@] - 1
+                        ),
+                    }
+                },
+                Err(_) => {
+                    &&& final(self)@ == old(self)@
+                    &&& !old(self)@.is_allocated(frame@)
+                }
+            },
+    )]
     fn free(&mut self, frame: FrameAddress) -> Result<(), Error> {
         let frame_number: usize = frame.into_frame_number().into_raw_value();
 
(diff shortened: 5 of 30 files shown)


src/kernel/src/hal/mem/types/address/aligned/page.spec.rs
```diff
@@ -0,0 +1,20 @@
+use crate::hal::mem::spec_page_size;
+
+verus! {
+
+// Success condition for the validating constructor, stated purely on the abstract
+// address value. `from_address` validates, it does NOT normalize: success requires the
+// *input* to already be page-aligned.
+pub open spec fn spec_aligned(addr_view: int) -> bool {
```
jaylorch, line 8: Is this necessary? You haven't even used it six lines below, on line 14.

src/kernel/src/hal/mem/types/address/frame.rs
```diff
@@ -42,15 +46,51 @@ impl FrameAddress {
     pub fn into_page_address(self) -> PageAddress {
```
jaylorch, line 47: Why is there no spec for this function?

src/kernel/src/hal/mem/types/address/frame.spec.rs
```diff
@@ -0,0 +1,45 @@
+verus! {
+
+/// The architectural page size.
+pub open spec fn spec_page_size() -> int {
+    ::arch::mem::PAGE_SIZE as int
+}
+
+/// The integer index (raw value) of a frame number: its `arch` view.
+pub open spec fn spec_frame_raw_value(frame: FrameNumber) -> int {
```
jaylorch: This function is unnecessary. It's just a more verbose version of `@`.

src/kernel/src/hal/mem/types/address/frame.spec.rs
```diff
@@ -0,0 +1,45 @@
...
+    frame@
+}
+
+/// The largest representable frame index (the `arch` bound `FrameNumber::spec_max()`).
+pub open spec fn spec_max_frame_number() -> int {
```
jaylorch: This function is unnecessary. Its invocation is longer than its body!

src/kernel/src/hal/mem/types/address/frame.spec.rs
```diff
@@ -0,0 +1,45 @@
...
+    FrameNumber::spec_max() as int
+}
+
+/// The frame an address belongs to: `addr / FRAME_SIZE` (equivalently `addr >> FRAME_SHIFT`).
+pub open spec fn spec_frame_number(addr_view: int) -> int {
```
jaylorch: This function is unnecessary. It's short enough already.

src/kernel/src/hal/mem/types/address/frame.spec.rs
```diff
@@ -0,0 +1,45 @@
...
+    addr_view / spec_page_size()
+}
+
+/// The base address of a frame: `frame_index * FRAME_SIZE`.
+pub open spec fn spec_from_number(frame_view: int) -> int {
```
jaylorch: This function is unnecessary. It's short enough already.

src/kernel/src/hal/mem/types/region.rs
```diff
@@ -195,11 +199,22 @@ impl<T: Address> MemoryRegion<T> {
     }
 
     /// Returns the first valid address that lies in the target memory region.
+    // `T: Address` requires `Copy`, so this is a plain copy: the returned value has the same
```
jaylorch: This extra added comment material is just confusing, and doesn't add anything. I suggest removing it.

src/kernel/src/mm/phys/frame.rs
```diff
@@ -147,6 +170,34 @@ impl Inner {
...
+                        &&& old(self)@.all_free(frames)
+                        &&& final(self)@ == FrameAllocView {
+                            refcounts: old(self)@.refcounts.union_prefer_right(
+                                Map::new(frames, |addr: int| 1int)
+                            ),
+                        }
+                    })
+                },
+                Err(_) => {
+                    final(self)@ == old(self)@
```
jaylorch: This should be stronger: It should say that there is no sequence of free frames of length `count`.

src/kernel/src/mm/phys/frame.rs
```diff
@@ -246,6 +321,32 @@ impl Inner {
     ///
     /// Upon success, `Ok(())` is returned. Upon failure, an error is returned instead.
     ///
+    #[verus_spec(result =>
```
jaylorch, line 334: This and many other functions in this PR add a spec but don't add a proof. So they'll fail verification. They need to be marked as trusted, e.g., with `#[verifier::external_body]`.

src/kernel/src/mm/phys/mod.spec.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// free, with per-frame refcounts. This is the existing `FrameAllocView`.
+    pub frames: FrameAllocView,
+    /// The `PhysMemoryManager` singleton has been initialized with a fresh user page pool.
+    pub manager_ready: bool,
+}
+
+/// Current abstract state of the global physical-memory subsystem.
+///
+/// Uninterpreted accessor for module-level physical-memory state.
+pub uninterp spec fn phys_view() -> PhysModView;
```
jaylorch: THIS IS VERY WRONG. Verus assumes that all spec functions are deterministic. Since `phys_view` has no parameters, that means Verus assumes it's a constant. This will let it prove all sorts of untrue things about the physical view, because it can assume it never changes.

src/kernel/src/mm/phys/frame.rs
```diff
@@ -456,6 +636,19 @@ fn instance() -> &'static mut Inner {
...
 /// Must be called exactly once during boot, before any other function
 /// in this module.
+///
+/// Initializes module-level storage and publishes the corresponding abstract state.
+#[verus_verify(external_body)]
+#[verus_spec(result =>
+    ensures
+        match result {
+            Ok(_) => {
+                &&& crate::mm::phys::phys_view().initialized
```
jaylorch: `phys_view()` is treated as a constant here, so this postcondition is meaningless. This function probably shouldn't have a spec until we have a better story for writing specs that include the global state.

jaylorch: Indeed, all of the static functions in this file (the ones that don't take a `self`) probably shouldn't be verified, until we have a better way to handle global state.

src/kernel/src/mm/phys/kframe.rs
```diff
@@ -49,6 +53,20 @@ impl KernelFrame {
...
+    #[verus_spec(result =>
+        requires
+            base.inv(),
+        ensures
+            match result {
+                Ok(kf) => {
+                    &&& kf@ == base@
+                    &&& kf.inv()
+                },
+                Err(_) => true,
```
jaylorch: This should be `false`, since it never returns an error.

src/kernel/src/hal/mem/types/address/aligned/page.spec.rs
tangruize: fixed in https://github.com/nanvix/nanvix/pull/2815/changes/a422a58c9487f399cdbe68986171efdd6c9abf34, `PageAligned::inv` now uses `spec_aligned(self@)`. It's also used in `page.rs`, so I didn't remove it.

src/kernel/src/hal/mem/types/address/frame.rs
tangruize: This function isn't on the call stack of top-level functions, so it's outside this PR's spec scope. Do you think we should add a spec for it here?

src/kernel/src/hal/mem/types/address/frame.spec.rs
tangruize: Removed in https://github.com/nanvix/nanvix/pull/2815/changes/f9d7d316acdb7c13b49fa3a09143bb4595ddcfcb

src/kernel/src/hal/mem/types/address/frame.spec.rs
tangruize: Removed in https://github.com/nanvix/nanvix/pull/2815/changes/f9d7d316acdb7c13b49fa3a09143bb4595ddcfcb

src/kernel/src/hal/mem/types/address/frame.spec.rs
tangruize: Removed in https://github.com/nanvix/nanvix/pull/2815/changes/f9d7d316acdb7c13b49fa3a09143bb4595ddcfcb

src/kernel/src/hal/mem/types/address/frame.spec.rs
tangruize: Removed in https://github.com/nanvix/nanvix/pull/2815/changes/f9d7d316acdb7c13b49fa3a09143bb4595ddcfcb

src/kernel/src/hal/mem/types/region.rs
tangruize: Removed in https://github.com/nanvix/nanvix/pull/2815/commits/3b69bfe84be341e2a4a7b2d543b2c9fa4a7faaca. It was a leftover from an earlier version where the Address wasn't bound by Copy, so the proof had to reason about clone.

src/kernel/src/mm/phys/frame.rs
tangruize: Fixed in https://github.com/nanvix/nanvix/pull/2815/commits/1cc50401f4d2a80eeefa8b8d34c53055b7589175

src/kernel/src/mm/phys/frame.rs
tangruize: Fixed in https://github.com/nanvix/nanvix/pull/2815/commits/8be0177529fea4e2e82354d3114b5ca42059aa25. This commit removes `phys_view()` and `PhysModView` and also removes specs based on  `phys_view()`. The specs in manager.rs/kframe.rs/upool.rs are also removed because they rely on global state we cannot soundly verify yet.

src/kernel/src/mm/phys/mod.spec.rs
tangruize: See below comment and this commit https://github.com/nanvix/nanvix/commit/8be0177529fea4e2e82354d3114b5ca42059aa25

src/kernel/src/mm/phys/kframe.rs
tangruize: Since `kframe.rs`'s specs were removed in https://github.com/nanvix/nanvix/commit/8be0177529fea4e2e82354d3114b5ca42059aa25, no fix needed here.

src/kernel/src/mm/phys/frame.rs
tangruize: Agreed, the PR itself should pass Verus.

I added this commit https://github.com/nanvix/nanvix/pull/2815/commits/d899c07091e77c20965103eb9d4d83f4801e66a8. it brings `verus_verify` to this PR and marked some functions with external_body temporarily. The follow-up proof PR will remove the externel_body.

src/kernel/src/mm/phys/mod.spec.rs
```diff
@@ -0,0 +1,171 @@
...
+// PhysMemoryManager - Specifications
+//
+// This file contains specification functions and view types.
+
+verus! {
+
+use crate::hal::mem::spec_page_size;
+use vstd::map::*;
+
+pub uninterp spec fn byte_at_address(ptr: int) -> u8;
```
jaylorch: This is very wrong. Verus assumes that any function is deterministic. That means that it will assume that the byte at address `ptr` is the same forever.

src/kernel/src/mm/phys/mod.spec.rs
```diff
@@ -0,0 +1,171 @@
...
+pub struct ExLinkedList<T, A: Allocator>(::alloc::collections::LinkedList<T, A>);
+
+//==================================================================================================
+// Frame-set vocabulary (on the existing FrameAllocView)
+//==================================================================================================
+
+impl FrameAllocView {
+    /// The allocator tracks (covers) the frame at `addr` — it is one of the frames this
+    /// allocator knows about, allocated or free. Models `frame::is_covered`.
+    pub open spec fn covers(self, addr: int) -> bool {
```
jaylorch: There's no need to have both `is_covered` and `covers` if they're synonyms. It's just confusingly redundant.

src/kernel/src/mm/phys/mod.spec.rs
```diff
@@ -0,0 +1,171 @@
...
+        self.is_covered(addr)
+    }
+
+    /// The frame at `addr` is reserved: it has a positive reference count, hence `alloc()`
+    /// can never return it. This is the core caller-visible fact a "booked" frame satisfies.
+    pub open spec fn reserved(self, addr: int) -> bool {
```
jaylorch: There's no need to have `reserved` and `is_allocated` if they're synonyms. It's just confusingly redundant.

src/kernel/src/mm/phys/mod.spec.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// Reserve only the *covered* frames of `set`; skip the rest (coverage-gated). Models
+    /// the MMIO booking rule.
+    pub open spec fn book_covered(self, set: Set<int>) -> FrameAllocView {
+        self.book_all(set.filter(|a: int| self.covers(a)))
+    }
+}
+
+/// Page-aligned physical frame addresses covered by a region `[start, start + size)`.
+/// Mirrors the frame-set computation used by `frame::alloc_range`.
+pub open spec fn region_frame_addrs(start: int, size: int) -> Set<int> {
```
jaylorch, line 95: This is a nice function. So it should be used more often. I see other places in the code that could use this but don't.

src/kernel/src/mm/phys/mod.spec.rs
```diff
@@ -0,0 +1,171 @@
...
+    let end_frame_number = (start + size) / spec_page_size();
+    vstd::set_lib::set_int_range(start_frame_number, end_frame_number)
+        .map(|i: int| i * spec_page_size())
+}
+
+/// The set of physical frame addresses covered by all regions in a physical
+/// non-usable-memory list. Uninterpreted: `LinkedList` has no Verus model, so its contents
+/// cannot be folded over in spec; this names the abstract union of `region_frame_addrs`
+/// over the list, which is all the contract needs.
+pub uninterp spec fn phys_regions_frame_set(
```
jaylorch: Don't do it this way. Instead, model `LinkedList`.

src/kernel/src/mm/phys/mod.spec.rs
```diff
@@ -0,0 +1,171 @@
...
+    regions: &::alloc::collections::LinkedList<TruncatedMemoryRegion<PhysicalAddress>>,
+) -> Set<int>;
+
+/// The set of physical frame addresses covered by all MMIO regions in a list (after
+/// GVA->GPA translation). Uninterpreted for the same reason as `phys_regions_frame_set`.
+pub uninterp spec fn mmio_regions_frame_set(
```
jaylorch: Don't do it this way. Instead, model `LinkedList`.

src/kernel/src/mm/virt/identity_map.spec.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// address. Membership means: the page is **present, writable, supervisor-only**, and
+    /// reachable at its own physical address in the kernel address space. Permissions are uniform
+    /// across all mapped pages, so they are encoded by membership rather than stored per page.
+    pub mapped: Set<int>,
+}
+
+/// Current abstract state of the global kernel identity map.
+///
+/// Uninterpreted accessor for module-level identity-map state.
+pub uninterp spec fn identity_map_view() -> IdentityMapView;
```
jaylorch: This is very wrong. Verus assumes that every function is deterministic. So, given this function, it can infer that the identity map view never changes.

src/kernel/src/mm/virt/identity_map.rs
```diff
@@ -499,6 +504,31 @@ fn ensure_identity_mapped_range(
 /// - [`ErrorCode::OutOfMemory`]: No BSS page table slots available.
 /// - [`ErrorCode::BadAddress`]: The allocated page table frame number is out of range.
 ///
+#[verus_verify(external_body)]
+#[verus_spec(result =>
+    requires
+        identity_map_view().inv(),
```
jaylorch: Everything in this file seems predicated on the very wrong assumption that `identity_map_view()` can somehow change. But it is assumed by Verus to be constant.

tangruize: Removed identify_map_view in this commit: https://github.com/nanvix/nanvix/pull/2815/commits/f4078a7974eeda0cbc783c5d8295ad98aa51f580

src/kernel/src/mm/phys/mod.spec.rs
tangruize: Removed: https://github.com/nanvix/nanvix/pull/2815/changes/3218c0e0465bd7a3503c02d5b19936b540a396ee

src/libs/arch/src/x86/mem/paging/frame/number.rs
```diff
@@ -64,6 +71,12 @@ impl FrameNumber {
     /// - `Some(`[`FrameNumber`]`)`: Upon success.
     /// - `None`: If the value is greater than [`Self::MAX`].
     ///
+    #[verus_spec(result =>
+        ensures
+            value as int <= Self::spec_max() ==> (result is Some
```
jaylorch: To avoid the use of the ugly `Some_0`, I suggest `result matches Some(v) && v@ == value`.

src/libs/arch/src/x86/mem/paging/pde.spec.rs
```diff
@@ -0,0 +1,156 @@
+verus! {
+
+//==================================================================================================
+// Flag projection helpers
+//==================================================================================================
+
+// Each paging-control enum is two-valued (`0` = clear, `1 << SHIFT` = set), i.e. isomorphic to
+// `bool`. These helpers project the "set" variant to its spec-world `bool` form. Pattern matching
+// is confined here so the View and the constructor transitions stay declarative.
+pub open spec fn spec_present_set(p: PresentFlag) -> bool {
```
jaylorch: This and the functions following are so short that they're unnecessary. They just make things more complicated.

src/libs/arch/src/x86/mem/paging/frame/number.rs
tangruize: Fixed in https://github.com/nanvix/nanvix/pull/2815/commits/2cc51ed2aa40cacddf056889d9123465a269aa2e

src/libs/arch/src/x86/mem/paging/pde.spec.rs
tangruize: Fixed in https://github.com/nanvix/nanvix/pull/2815/commits/0e725071e327ccbd62d41c5d7428af610e9c77a3

src/libs/arch/src/x86/mem/paging/table.spec.rs
```diff
@@ -0,0 +1,101 @@
...
+//==================================================================================================
+
+/// Low `log2(PAGE_TABLE_LENGTH)` bits of `vaddr >> shift`.
+///
+/// Mirrors the exec masking `(vaddr >> shift) & (PAGE_TABLE_LENGTH - 1)` exactly.
+pub open spec fn spec_table_index(vaddr: usize, shift: usize) -> nat {
+    ((vaddr >> shift) & ((crate::mem::PAGE_TABLE_LENGTH - 1) as usize)) as nat
+}
+
+/// PD index: bits `[PGTAB_SHIFT, PGTAB_SHIFT + log2(LEN))` of `vaddr`.
```
jaylorch: This comment doesn't make much sense. For example, LEN doesn't seem to mean anything.

src/libs/arch/src/x86/mem/paging/table.spec.rs
```diff
@@ -0,0 +1,101 @@
...
+pub open spec fn spec_pd_index(vaddr: usize) -> nat {
+    spec_table_index(vaddr, crate::mem::PGTAB_SHIFT)
+}
+
+/// PT index: bits `[PAGE_SHIFT, PAGE_SHIFT + log2(LEN))` of `vaddr`.
```
jaylorch: This comment doesn't make much sense. For example, LEN doesn't seem to mean anything.

src/libs/arch/src/x86/mem/paging/table.spec.rs
jaylorch: The comment should instead describe what the function's output should be interpreted as meaning.

src/libs/arch/src/x86/mem/paging/table.spec.rs
jaylorch: The comment should instead describe what the function's output should be interpreted as meaning.

src/libs/arch/src/x86/mem/paging/table.rs
```diff
@@ -119,6 +159,9 @@ impl<E: TableEntry> Table<E> {
     /// `base` must be a valid, page-aligned address with at least one page of readable/writable
     /// memory.
     ///
+    #[verus_spec(result =>
```
jaylorch: The comment above suggests this function is missing at least one precondition.

src/libs/arch/src/x86/mem/paging/table.rs
```diff
@@ -152,6 +203,15 @@ impl<E: TableEntry> Table<E> {
     ///
     /// The memory at `base + index * size_of::<PteWord>()` must be valid for a volatile write.
     ///
+    /// Performs a volatile page-table store.
+    ///
+    /// The contract keeps only the index precondition because the table contents are mutable
```
jaylorch: Why is this even specified if we can't model its effect?

jaylorch: Indeed, why are we even specifying anything about page tables if we're writing a spec about physical memory? Shouldn't that wait until we deal with virtual memory?

src/libs/arch/src/x86/mem/paging/table.spec.rs
```diff
@@ -0,0 +1,101 @@
+verus! {
+
+//==================================================================================================
```
jaylorch: Why are we talking about page tables in a spec about physical memory?

src/libs/arch/src/x86/mem/constants.rs
```diff
@@ -6,6 +6,7 @@
 //==================================================================================================
 
 use sys::mm::Alignment;
+use vstd::prelude::*;
 
 #[allow(unused_imports)]
 use ::vstd::prelude::*;
```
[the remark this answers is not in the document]
jaylorch: I agree. This should be fixed.

src/libs/bump_allocator/src/lib.rs
```diff
@@ -114,6 +121,13 @@ use ::core::{
 ///
 /// Returns the aligned value, or `None` if `alignment` is zero or the computation overflows.
 ///
+#[verus_spec(result =>
+    ensures
+        match result {
```
jaylorch: This code would be more compact if it started with `match align_up_spec(...)` instead of `match result`.

src/libs/bump_allocator/src/lib.rs
```diff
@@ -240,6 +262,22 @@ impl<const N: usize, const A: usize, S: BssStorage> FixedSizeBumpAllocator<N, A,
     /// - [`BumpAllocError::OutOfBounds`] if computed slot exceeds storage bounds.
     /// - [`BumpAllocError::Misaligned`] if computed slot is not properly aligned.
     ///
+    /// Materializes a slot reference from backend-provided storage.
+    #[verus_verify(external_body)]
+    #[verus_spec(result =>
```
jaylorch: The body of this function uses weak-memory atomic operations to effectively mutate `self` even though it's only passed an immutable reference to it. So I don't understand how you plan to verify its operation. You're going to need something like an atomic invariant and a tracked object.

src/libs/bump_allocator/src/lib.spec.rs
```diff
@@ -0,0 +1,163 @@
...
+// BumpView - abstract pool model
+//==================================================================================================
+
+/// Abstract view of a `FixedSizeBumpAllocator<N, A, S>`.
+///
+/// Pure ghost description of the slot pool: no atomics, no raw pointers, no cursor
+/// mechanism. Carries the fixed pool geometry plus the single dynamic quantity
+/// `allocated` (the number of slots handed out so far).
+#[verifier::ext_equal]
+pub struct BumpView {
```
jaylorch: This abstraction doesn't hide anything about the implementation. The caller doesn't need to know all this internal detail of the bump allocator. It just cares that references it receives are aligned and the right size.

src/libs/bump_allocator/src/lib.spec.rs
```diff
@@ -0,0 +1,163 @@
...
+        // (d) Addresses do not wrap the usize space.
+        &&& self.base >= 0
+        &&& self.base + (self.storage_size as int) <= usize::MAX as int + 1
+        // (e) Monotone-capacity ceiling.
+        &&& self.allocated <= self.capacity
+    }
+
+    /// The three geometric guarantees the kernel's `unsafe` soundness relies on,
+    /// stated over the whole pool: every slot is aligned, in-bounds, and distinct.
+    pub open spec fn geometry_ok(self) -> bool {
```
jaylorch: This exposes too much detail of the implementation. This kind of things should be part of the implementation rather than the abstraction.

src/libs/sys/src/sys/mm/address/mod.rs
```diff
@@ -46,9 +51,21 @@ where
     /// - `Ok(Self)`: The new address.
     /// - `Err(Error::BadAddress)`: If the provided address is invalid.
     ///
+    #[verus_spec(result =>
+        ensures
+            match result {
+                Ok(a) => a@ == raw_addr as int,
+                Err(e) => e.code == crate::error::ErrorCode::BadAddress,
```
jaylorch: It seems like this won't be strong enough. It allows returning an error for arbitrary reasons. Perhaps the trait should also include a spec function that says whether the address is valid, and that will correspond to whether this function returns Ok or Err.

jaylorch (commented):
See above line-by-line comments

src/libs/bump_allocator/src/lib.rs
jaylorch: Not to mention the need to verify weak memory operations, which Verus doesn't currently support.

src/libs/arch/src/x86/mem/paging/table.spec.rs
tangruize: The table.spec.rs is removed as page tables are not in physical mm's verification scope: https://github.com/nanvix/nanvix/pull/2815/commits/c516e17859a923a612d380c941a2af1bf74d18d0

src/libs/arch/src/x86/mem/paging/table.spec.rs
tangruize: The table.spec.rs is removed as page tables are not in physical mm's verification scope: https://github.com/nanvix/nanvix/pull/2815/commits/c516e17859a923a612d380c941a2af1bf74d18d0

src/libs/arch/src/x86/mem/paging/table.rs
tangruize: This spec is removed as page tables are not in physical mm's verification scope: https://github.com/nanvix/nanvix/pull/2815/commits/c516e17859a923a612d380c941a2af1bf74d18d0

src/libs/arch/src/x86/mem/paging/table.rs
tangruize: Yes, you're right. I've removed these specs, see commit https://github.com/nanvix/nanvix/pull/2815/commits/c516e17859a923a612d380c941a2af1bf74d18d0

src/libs/arch/src/x86/mem/paging/table.spec.rs
tangruize: These were leftovers from removing the phy-mm wrapper function. removed now: https://github.com/nanvix/nanvix/pull/2815/commits/c516e17859a923a612d380c941a2af1bf74d18d0

src/libs/bump_allocator/src/lib.rs
tangruize: The bump allocator was reachable only in virt-mm and is outside the scope of this phy-mm PR, so I removed it: https://github.com/nanvix/nanvix/pull/2815/commits/c71dd7a8e052f1e62d9f4facc4b077777319e718

src/libs/bump_allocator/src/lib.rs
tangruize: Agreed.

The bump allocator was reachable only in virt-mm and is outside the scope of this phy-mm PR, so I removed it: https://github.com/nanvix/nanvix/pull/2815/commits/c71dd7a8e052f1e62d9f4facc4b077777319e718

src/libs/sys/src/sys/mm/address/mod.rs
tangruize: Strengthened in this commit: https://github.com/nanvix/nanvix/pull/2815/commits/dc1630e4619135fec41f7251e3f555561589ff64

src/kernel/src/hal/mem/types/address/frame.rs
jaylorch: Ah. When I wrote that comment, I was under the mistaken impression that this PR was just the high-level trusted spec.

jaylorch (dismissed):

PR #2884, opened 2026-07-07
[net-backend] Add non-blocking socket support

liam0215:
Add NetBackend::set_nonblocking() plus NetError::is_would_block() and is_in_progress() so callers can drive host sockets without blocking. The mode is opt-in: sockets are still created blocking, so existing callers are unaffected.

Move host errno normalization into the shared error crate and update net-backend to use it. Linux and Winsock errno values are now translated to Nanvix errno numbering before conversion to ErrorCode, avoiding raw host errno values being reinterpreted as unrelated Nanvix errors. This is needed because Linux/Winsock and Nanvix use different numeric layouts for many socket errors, such as EINPROGRESS, EALREADY, and EISCONN.

Handle the Windows non-blocking connect() special case separately: generic WSAEWOULDBLOCK still maps to TryAgain, while connect() maps it to OperationInProgress. EALREADY is also treated as an in-progress connect state.

This is groundwork for the decoupled networkd epoll reactor, which multiplexes non-blocking host sockets and parks operations that report would-block or in-progress.

### DIFF
--- a/src/libs/net-backend/src/platform/unix.rs
+++ b/src/libs/net-backend/src/platform/unix.rs
@@ -85,13 +85,6 @@ pub(crate) fn socket_failed(result: RawSocket) -> bool {
     result == INVALID_SOCKET
 }
 
-/// Maps a platform errno to the POSIX-like errno values used by `ErrorCode::try_from`.
-/// On Unix, errno is already POSIX.
-#[inline]
-pub(crate) fn normalize_errno(errno: i32) -> i32 {
-    errno
-}
-
 //==================================================================================================
 // Socket Close
 //==================================================================================================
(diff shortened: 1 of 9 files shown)


src/libs/net-backend/src/platform/unix.rs
```diff
@@ -85,11 +85,59 @@ pub(crate) fn socket_failed(result: RawSocket) -> bool {
...
+        libc::EPROTOTYPE => ::sysapi::errno::EPROTOTYPE,
+        libc::ESHUTDOWN => ::sysapi::errno::ESHUTDOWN,
+        libc::ESOCKTNOSUPPORT => ::sysapi::errno::ESOCKTNOSUPPORT,
+        libc::ETIMEDOUT => ::sysapi::errno::ETIMEDOUT,
+        libc::ETOOMANYREFS => ::sysapi::errno::ETOOMANYREFS,
+        libc::EUSERS => ::sysapi::errno::EUSERS,
+        other => other,
+    }
+}
+
```
ppenna: We should actually have this as a function in the `errno` crate and translate all error codes. As it is, `other` can be wrongly converted.

src/libs/net-backend/src/platform/windows.rs
```diff
@@ -92,53 +92,66 @@ pub(crate) fn socket_failed(result: RawSocket) -> bool {
...
+        winsock::WSAEPROTONOSUPPORT => ::sysapi::errno::EPROTONOSUPPORT,
+        winsock::WSAEPROTOTYPE => ::sysapi::errno::EPROTOTYPE,
+        winsock::WSAESHUTDOWN => ::sysapi::errno::ESHUTDOWN,
+        winsock::WSAESOCKTNOSUPPORT => ::sysapi::errno::ESOCKTNOSUPPORT,
+        winsock::WSAETIMEDOUT => ::sysapi::errno::ETIMEDOUT,
+        winsock::WSAETOOMANYREFS => ::sysapi::errno::ETOOMANYREFS,
+        winsock::WSAEWOULDBLOCK => ::sysapi::errno::EAGAIN,
+        winsock::WSAEUSERS => ::sysapi::errno::EUSERS,
+        winsock::WSANOTINITIALISED => ::sysapi::errno::EBADFD, // closest match
         _ => errno,
```
ppenna: See my comment above.

liam0215: What about for the connect-specific error conversions?

PR #3032, opened 2026-08-04
[vfs] chdir only on dir

ada-x64:
Ensure that chdir errors when attempting to chdir onto a file. Also hooks up a regression test.

### DIFF
--- a/src/libs/vfs/src/filesystem.rs
+++ b/src/libs/vfs/src/filesystem.rs
@@ -286,14 +286,12 @@ pub(crate) fn rename(cwd: &str, old_path: &str, new_path: &str) -> Result<(), Fa
 /// - [`Fat32Error::NotInitialized`] if the filesystem hasn't been initialized.
 /// - [`Fat32Error::InvalidPath`] if the path is malformed.
 /// - [`Fat32Error::NotFound`] if no mount handles this path.
+/// - [`Fat32Error::NotADirectory`] if the path is not a directory.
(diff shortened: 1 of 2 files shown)


src/tests/integration/test-c-chdir/main.c
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT License.
+
+#include <assert.h>
+#include <errno.h>
+#include <limits.h>
+#include <string.h>
+#include <unistd.h>
+
+// Regression test. Assert that we cannot chdir onto a file.
+// See https://github.com/nanvix/cpython/issues/516
```
ppenna: Please remove this comment. Issues should be mentioned only on `TODO` and `FIXME` comments.

src/tests/integration/test-c-chdir/main.c
```diff
@@ -0,0 +1,25 @@
...
+
+    assert(chdir("/marker.txt") != 0);
+    assert(errno == ENOTDIR);
+
+    // A failed chdir must leave the cwd unchanged.
+    char after[PATH_MAX];
+    assert(getcwd(after, sizeof(after)) != NULL);
+    assert(strcmp(before, after) == 0);
+    return 0;
+}
```
ppenna: Add a test case that also covers cross-mount paths would be nice. Also, we could likely reuse an existing testing harness for this test. Check where test for `chdir` are currently defined.

ada-x64: https://github.com/nanvix/nanvix/issues/3033
Leaving this for a follow-up

PR #3038, opened 2026-08-05
[vfsd] hostfs-aware chdir

### LINKED ISSUE #3033: [vfsd] Support chdir to hostfs mount
## Context

`chdir` only resolves paths within the guest VFS mount table. Every other hostfs-aware syscall (`open`, `stat`, `mkdir`, `rename`, …) forwards `/mnt` paths to hostfsd, but `handle_chdir` does not — so `chdir("/mnt/...")` fails with `NoSuchEntry` even when the directory exists on the host.

## Proposed Solution

Make `chdir` hostfs-aware, mirroring `handle_fstatat_with_hostfs`:

- Route `chdir` through a `handle_chdir_with_hostfs` variant.
- For `is_hostfs_path` targets, forward a path-stat to hostfsd and defer.
- On completion, set the cwd if the target is a directory, else return `ENOTDIR`.

Non-hostfs paths keep the existing behavior. `fchdir` over hostfs and cwd-under-`/mnt` unmount are out of scope (follow-ups).

ada-x64:
This commit enables hostfs aware chdir.

Closes #3033.
Follow-up to #3032.

### DIFF
--- a/src/daemons/vfsd/src/pending.rs
+++ b/src/daemons/vfsd/src/pending.rs
@@ -20,10 +20,16 @@ extern crate alloc;
 
 use crate::error::{
     build_error,
+    fat32_to_error_code,
     send_response,
 };
-use ::alloc::collections::BTreeMap;
+use ::alloc::{
+    collections::BTreeMap,
+    string::String,
+};
 use ::hostfs_api::{
+    file_kind,
+    LstatResponse,
     OperationId,
     OperationIdAllocator,
 };
@@ -38,6 +44,8 @@ use ::sys::{
         ThreadIdentifier,
     },
 };
+use ::syscall::unistd::message::ChangeDirectoryResponse;
+use ::vfs::fd::vfs_set_cwd;
 
 //==================================================================================================
 // Pending Operation Descriptor
@@ -97,6 +105,12 @@ pub(crate) enum PendingOpKind {
     /// Path-based stat that follows the final symbolic link (default `stat(2)` semantics).
     /// Shares the `lstat` response wire format and completion path.
     PathStat,
+    /// chdir onto a hostfs path — a path-based stat whose completion commits the cwd
+    /// when the target is a directory (else `ENOTDIR`). Reuses the `PathStat` wire form.
+    Chdir {
+        /// Absolute hostfs path to become the cwd once confirmed to be a directory.
+        path: String,
+    },
     /// getdents — directory listing over hostfs.
     ///
     /// hostfsd returns one entry per IKC round-trip, so a single guest `getdents`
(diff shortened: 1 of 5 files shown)


src/daemons/vfsd/src/pending.rs
```diff
@@ -1164,3 +1172,43 @@ fn complete_lstat(source_tid: ThreadIdentifier, response_payload: &[u8; Message:
...
 }
+
+/// Completes a deferred hostfs `chdir`: the pending path-stat has returned, so
+/// commit the cwd when the target is a directory, else surface `ENOTDIR`.
+fn complete_chdir(
+    source_tid: ThreadIdentifier,
+    response_payload: &[u8; Message::PAYLOAD_SIZE],
+    path: alloc::string::String,
+) {
+    use ::syscall::unistd::message::ChangeDirectoryResponse;
```
ppenna: Please review all your PR. Symbols must be imported, not have their path hardcoded.

src/libs/vfs/src/fd/mod.rs
```diff
@@ -1495,6 +1495,24 @@ pub fn vfs_chdir(path: &str) -> Result<(), Fat32Error> {
...
+/// Commits the current working directory to a hostfs path, bypassing local
+/// mount-table validation but still enforcing an absolute, normalized cwd.
+///
+/// Unlike [`vfs_chdir`], this performs NO existence/type check: vfsd uses it to
+/// finalize a chdir onto a hostfs path after hostfsd has confirmed the target is
+/// a directory (hostfs paths live outside the local FAT mount table, so
+/// [`vfs_chdir`] would reject them). `path` is normalized here so the stored cwd
+/// keeps the same absolute, canonical form as every other code path; a relative
+/// or malformed `path` is rejected with [`Fat32Error::InvalidPath`].
+pub fn vfs_set_cwd(path: &str) -> Result<(), Fat32Error> {
```
ppenna, line 1507: Path is already resolved, and this may be confusing on the logic bellow. Please create a follow up issue that is about introducing a ResolvedPath structure and change the vfsd code to use it.

ada-x64: #3039

src/daemons/vfsd/src/pending.rs
```diff
@@ -97,6 +97,12 @@ pub(crate) enum PendingOpKind {
     /// Path-based stat that follows the final symbolic link (default `stat(2)` semantics).
     /// Shares the `lstat` response wire format and completion path.
     PathStat,
+    /// chdir onto a hostfs path — a path-based stat whose completion commits the cwd
+    /// when the target is a directory (else `ENOTDIR`). Reuses the `PathStat` wire form.
+    Chdir {
+        /// Absolute hostfs path to become the cwd once confirmed to be a directory.
+        path: alloc::string::String,
```
ppenna: This should be an import.

src/daemons/vfsd/src/pending.rs
```diff
@@ -1164,3 +1180,40 @@ fn complete_lstat(source_tid: ThreadIdentifier, response_payload: &[u8; Message:
         },
     }
 }
+
+/// Completes a deferred hostfs `chdir`: the pending path-stat has returned, so
+/// commit the cwd when the target is a directory, else surface `ENOTDIR`.
+fn complete_chdir(
+    source_tid: ThreadIdentifier,
+    response_payload: &[u8; Message::PAYLOAD_SIZE],
+    path: String,
```
ppenna, line 1189: No need to pass as a `String`. Ownership is not required and copy can be avoided. Use `&str`

PR #3042, opened 2026-08-06
[vfs] Add ResolvedPath struct

### LINKED ISSUE #3039: [vfsd] - Introduce ResolvedPath
## Context

#3038 - We introduce hostfs-aware chdir semantics. However, we are relying on raw string data for paths.

## Proposed Solution

We should introduce a `ResolvedPath` struct to more accurately capture the intended semantics. In particular, the new `vfs_set_cwd` API should take a `ResolvedPath` as input rather than verifying that the path has been resolved at each call. This moves the responsibility away from the function and to the caller.

ada-x64:
Adds a `ResolvedPath` newtype marking paths that have been through
`vfs_resolve_path`, so callees stop re-validating. Closes #3039.

`Resolved` means *provenance*: the path is absolute because it was anchored
against a dirfd or the cwd. It deliberately does **not** mean normalized —
hostfs paths go to hostfsd verbatim, which resolves `..` physically. Collapsing
them lexically here would diverge whenever a component is a symlink.

Three consequences worth review:

- `vfs_set_cwd` takes `ResolvedPath` and returns `()`. The absolute check is now
  the type's job. Normalization stays (otherwise `cd /mnt/foo/..` is stored
  verbatim and grows), which means it has to be infallible — hence:
- `..` past the root now clamps to `/` instead of returning `InvalidPath`.
  POSIX defines `/..` as `/`. Guest-visible: `stat("/..")` was EINVAL, now
  succeeds.
- `vfs_resolve_path("")` now returns `None`. It previously resolved to the cwd,
  so `open("")` handed back a directory fd and `chdir("")` silently succeeded.

`normalize_absolute` takes `&ResolvedPath` rather than `&str` so its
precondition is carried by the type instead of asserted. That puts one upward
`mount` -> `fd` import in the crate; the alternative is `debug_assert!` on a
`&str`.

Follow-up: #3043 (backend-routed path type).

### DIFF
--- a/src/libs/vfs/src/fd/mod.rs
+++ b/src/libs/vfs/src/fd/mod.rs
@@ -34,6 +34,7 @@ use crate::{
         LineDiscipline,
         TerminalSignal,
     },
+    mount::normalize_absolute,
     process::{
         current_cwd,
         current_pid,
@@ -53,6 +54,7 @@ use ::alloc::{
         BTreeMap,
         BTreeSet,
     },
+    format,
     string::String,
     sync::Arc,
     vec::Vec,
@@ -960,6 +962,23 @@ pub fn vfs_poll(fd: c_int, events: c_short) -> Result<c_short, Fat32Error> {
     Ok(revents)
 }
 
+/// An absolute VFS path produced by [`vfs_resolve_path`].
+///
+/// Carries the proof that a `dirfd` + `path` pair has already been anchored, so
+/// callees need not re-check it. The path is absolute but not necessarily
+/// lexically normalized: `.`/`..` are left for whoever owns the path to
+/// interpret, since hostfs resolves them against the real filesystem.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ResolvedPath(String);
+
+impl core::ops::Deref for ResolvedPath {
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
 /// Resolves a `dirfd` + `path` pair into an absolute VFS path.
 ///
 /// If `path` is absolute, it is returned as-is (dirfd is ignored per POSIX).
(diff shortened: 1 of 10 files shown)


src/libs/vfs/src/fd/mod.rs
```diff
@@ -960,6 +962,23 @@ pub fn vfs_poll(fd: c_int, events: c_short) -> Result<c_short, Fat32Error> {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ResolvedPath(String);
+
+impl core::ops::Deref for ResolvedPath {
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
ppenna: We intentionally should not provide this, we dont want arbitrary casts

src/libs/vfs/src/fd/mod.rs
```diff
@@ -960,6 +962,23 @@ pub fn vfs_poll(fd: c_int, events: c_short) -> Result<c_short, Fat32Error> {
...
 }
 
+/// An absolute VFS path produced by [`vfs_resolve_path`].
+///
+/// Carries the proof that a `dirfd` + `path` pair has already been anchored, so
+/// callees need not re-check it. The path is absolute but not necessarily
+/// lexically normalized: `.`/`..` are left for whoever owns the path to
+/// interpret, since hostfs resolves them against the real filesystem.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ResolvedPath(String);
```
ppenna: This should be in a module such that we prevent arbitrary structure construction.

src/libs/vfs/src/path.rs
ppenna, line 1: For a layered architecture, I believe this should be a submodule of `fd`

PR #3047, opened 2026-08-07
[ci] static benchmark reporting site

ada-x64:
## Benchmark reporting via a static site

Replaces the markdown-table PR comment and the 780-line matplotlib plotting
script with a static site (`www/bench/`) published to GitHub Pages, plus a
single PR comment linking to it.

### How it works

- Benchmark history CSVs move from `data/{baremetal,windows-baremetal}/` to
  `www/bench/data/{linux,windows}-baremetal/`, so the site serves them directly.
- `scripts/benchmark.py ci-url` builds a link whose `?d=` payload carries the
  run's results base64-encoded. The site overlays that commit on top of the
  runner's persisted history. With no `--commit`, it renders history only —
  which is what you get browsing the site bare.
- `deploy-bench.yml` publishes `www/bench/` on changes under that path.
- The two finalize jobs each emit a URL; a new `benchmark-report` job posts one
  comment with both platform links.

### Removed

- `scripts/plot-performance.py` and the matplotlib dependency.
- `.github/actions/benchmark-report/` and `post-benchmark-comment.yml`. The
  latter existed to get a write token for fork PRs; the benchmark gates already
  skip forks entirely, so the comment can be posted inline.
- `benchmark.py ci-summary` and its table-rendering helpers, now unused.

Regression gating (`ci-gate`) and persistence (`benchmark-persist`) are
unchanged apart from the new paths and an explicit `source-dir`.

### Notes for reviewers

- `benchmark-persist` gained a `source-dir` input: artifacts now land in a
  per-platform subdirectory instead of the repo root.
- The baseline commits no longer carry `[skip ci]`, otherwise the push that
  updates the data would never redeploy the site. `ci.yml` now ignores
  `www/bench/data/**` on dev pushes to compensate.
- `deploy-bench.yml` only runs on dev pushes and manual dispatch. There is no
  PR preview: Pages has a single deployment target, so a preview would publish
  over the live site. This part gets exercised after merge.
- The regression gate now reads the baseline from the working tree instead of
  checking it out of dev. The gates already require the branch to be up to date
  with dev, so the two copies are the same, and this stops the gate from
  breaking whenever the baseline paths move.


scripts/plot-performance.py
ppenna, line 1: We should keep script so people can build plots offline.

ada-x64: https://github.com/nanvix/nanvix/pull/3047/changes#diff-85558aea93e4748f8ec3ffe53c8a44a78b1230c411ea9281def7866ce02c8753R90-R100

Plots can be built offline by running a simple local HTTP server, as demonstrated here. I removed this script because it was re-implemented in this PR. Maybe a simple script to do this would be nice?

PR #3060, opened 2026-08-10
[vfs] Fix POSIX rename replace-existing semantics

ada-x64:
Fix for nanvix/cpython#501.

`os.rename()` / `os.replace()` returned `EEXIST` when the destination already existed. POSIX requires replacing the target. The underlying rust-fatfs `rename` refuses to overwrite, so the VFS layer now removes the destination first after validating type compatibility (file↔dir mismatch → proper errno, non-empty dir → `ENOTEMPTY`, identity rename → no-op).

Non-atomic unlink+rename is documented; in practice unreachable on the in-memory FAT backend.

Adds C test covering subdirectory renames and replace-existing (`renameat_subdir.c`).

### DIFF
--- a/src/libs/vfs/src/filesystem.rs
+++ b/src/libs/vfs/src/filesystem.rs
@@ -250,9 +250,17 @@ pub(crate) fn read_dir(cwd: &str, path: &str) -> Result<Vec<DirEntry>, Fat32Erro
 /// # Errors
 ///
 /// - [`Fat32Error::NotInitialized`] if the filesystem hasn't been initialized.
+/// - [`Fat32Error::ReadOnly`] if the mount is read-only.
 /// - [`Fat32Error::NotFound`] if `old_path` doesn't exist.
-/// - [`Fat32Error::AlreadyExists`] if `new_path` already exists.
-/// - [`Fat32Error::InvalidPath`] if paths are on different mounts.
+/// - [`Fat32Error::NotADirectory`] if source is a directory but destination is a file.
+/// - [`Fat32Error::NotAFile`] if source is a file but destination is a directory.
+/// - [`Fat32Error::NotEmpty`] if destination is a non-empty directory (via `rmdir`).
+/// - [`Fat32Error::InvalidPath`] if paths are on different mounts, or if path resolution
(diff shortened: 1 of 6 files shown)


src/libs/vfs/src/filesystem.rs
```diff
@@ -271,7 +279,42 @@ pub(crate) fn rename(cwd: &str, old_path: &str, new_path: &str) -> Result<(), Fa
...
+                } else {
+                    fat.unlink(&new_rel)?;
+                }
+            },
+            Err(Fat32Error::NotFound) => {},
+            Err(e) => return Err(e),
+        }
+
+        fat.rename(&old_rel, &new_rel)
     })
```
ppenna, line 325: I am not sure we are correctly handling filenames containing slash characters. Please investigate and create follow-up issue

src/tests/integration/test-c-file/renameat_subdir.c
```diff
@@ -0,0 +1,181 @@
...
+
+    test_rename_root();
+    test_rename_within_subdir();
+    test_rename_subdir_to_root();
+    test_rename_across_subdirs();
+    test_rename_replace_existing();
+    test_rename_identity();
+
+    fprintf(stderr, "renameat subdir: all passed\n");
+}
```
ppenna, line 287: Please add  test cases for renaming dot and dot dot.

src/libs/vfs/src/filesystem.rs
ada-x64: https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap03.html#tag_03_146

> The bytes composing the name shall not contain the <NUL> or <slash> characters.

Added a test to ensure that an attempt to construct a path with <slash> fails. Also added some tests for non-portable POSIX filename symbols (\, ?, *, etc)

src/tests/integration/test-c-file/renameat_subdir.c
ada-x64: Fixed a bug where trailing dot and dotdot were being lexically parsed, so e.g. renaming 'd/.' and 'd/..' would pass.

ppenna (approved):

PR #3097, opened 2026-08-12
[libc] Fix NULL times in the utime family to set current time

### LINKED ISSUE #3066: [posix] utimensat with NULL times returns EINVAL instead of setting current time
When called with times=None, cpython invokes utimensat(fd, NULL, NULL, 0)
to request the current time. Nanvix rejects the NULL times pointer with
EINVAL instead of treating it as "set to now". Affects all modes (both
standalone and hosted).

Distinct from the FAT mtime-not-persisted issue (cpython#514) — this
is about the utimensat syscall rejecting a valid NULL argument.

Fix: in the utimensat handler, treat NULL times as "set to current
time" per POSIX.

Affects: test_os (test_utime_current, test_utime_current_old).

Related: nanvix/cpython#529

ada-x64:
Closes #3066

## What
`utimensat`, `futimens`, `utimes`, and `utime` all rejected a NULL `times` pointer with `EINVAL`. POSIX specifies NULL means "set both access and modification times to the current time." This broke cpython's `os.utime(path, times=None)` (nanvix/cpython#529), affecting `test_os::test_utime_current` / `test_utime_current_old`.

Distinct from the FAT mtime-not-persisted issue (cpython#514) — this is purely the syscalls rejecting a valid NULL argument.

## Change
- Extracted a shared `resolve_times()` helper mapping NULL → `UTIME_NOW` for both entries; `utimensat` and `futimens` use it.
- `utimes`/`utime` forward NULL down to the fixed `utimensat` (they are thin adapters that convert their struct type to `timespec` then call `utimensat`).
- Added a repro asserting NULL succeeds for all four.

## Verification
Verified against the full lifecycle; the `test-c-file` POSIX suite passes and traces confirm NULL converts to `UTIME_NOW`.

### DIFF
--- a/src/libs/libc_sys_stat/src/lib.rs
+++ b/src/libs/libc_sys_stat/src/lib.rs
@@ -175,6 +175,26 @@ pub unsafe extern "C" fn fchmodat(
     }
 }
 
+/// Resolves a raw `times` pointer into an owned `[timespec; 2]`.
+///
+/// A NULL pointer maps to `UTIME_NOW` for both entries, per POSIX ("set both
+/// timestamps to the current time"). A non-NULL pointer must reference exactly
+/// two elements; otherwise [`None`] is returned.
+///
+/// # Safety
+///
+/// The caller must ensure a non-NULL `times` points to two valid `timespec`s.
+unsafe fn resolve_times(times: *const timespec) -> Option<[timespec; 2]> {
+    if times.is_null() {
+        let now: timespec = timespec {
+            tv_sec: 0,
+            tv_nsec: sys_stat::UTIME_NOW,
+        };
+        return Some([now; 2]);
+    }
+    slice::from_raw_parts(times, 2).try_into().ok()
+}
+
 ///
 /// # Description
 ///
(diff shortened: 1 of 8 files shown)


src/libs/libc_sys_stat/src/lib.rs
```diff
@@ -175,6 +175,26 @@ pub unsafe extern "C" fn fchmodat(
...
+/// # Safety
+///
+/// The caller must ensure a non-NULL `times` points to two valid `timespec`s.
+unsafe fn resolve_times(times: *const timespec) -> Option<[timespec; 2]> {
+    if times.is_null() {
+        let now: timespec = timespec {
+            tv_sec: 0,
+            tv_nsec: sys_stat::UTIME_NOW,
+        };
+        return Some([now; 2]);
```
ppenna, line 193: "Only a process with the effective user ID equal to the user ID of the file, or with write access to the file, or with appropriate privileges may use futimens() or utimensat() with a null pointer as the times argument or with both tv_nsec fields set to the special value UTIME_NOW. "

Although Nanvix is single user, for completeness we may want to include a check and fail otherwise.

ada-x64: As you mention the user check is redundant (and not implemented elsewhere), and 'appropriate privileges' just equates to writability. So really all we need to do is check for writability. 

I suggest we introduce this in the follow-up PR #3098. We have `check_writable` for mkdir, rmdir, open, and rename, which guards _right_ before mutation. We could add a similar check here once we're actually mutating. I've added a comment at `vfs_utimensat`, which is where the future guard should live.

EDIT: Ideally, this sort of guard lives on the type level...

Issue #3104, opened 2026-08-14
[kernel] Per-process euid and utimensat permission enforcement


ada-x64:
`utimensat`/`futimens` apply times with no permission check. POSIX requires the
caller to own the file, have write access (for the NULL / `UTIME_NOW` case), or
be privileged; otherwise `EPERM`/`EACCES`. Nanvix cannot enforce this because it
has no per-process user identity:

- `src/libs/syscall/src/unistd/syscall/geteuid.rs` — `geteuid()` returns the
  `ROOT` constant with no kcall; `getuid()` likewise.
- `setuid`/`seteuid` exist only as binding no-ops (no kcall, no state).
- `src/daemons/vfsd/src/pending.rs` — `st_uid` is hardcoded `0`.
- The kernel `ProcessState` (`src/kernel/src/pm/process/state/mod.rs:248`) stores
  `capabilities` but no uid.

The enforcement point is vfsd, not libc: the VFS owns file metadata and can
check atomically. vfsd already receives the kernel-attested `source` thread, so
it knows *who* called — it just cannot ask *what uid* they are.

The euid must come from a source vfsd can trust (the kernel). A caller-supplied
euid in the IPC request is spoofable and would make the check meaningless.

## Work

1. Kernel owns the euid. Add `euid: UserIdentifier` to `ProcessState`, default
   `ROOT`, inherited on fork/spawn — mirroring the existing `capabilities`
   field and its getters/setters. Wire a real `geteuid` kcall so a process's own
   `geteuid()` reflects kernel state. `setuid`/`seteuid` kcalls can stay stubbed
   until multiuser needs them.

2. vfsd learns the caller's euid. Add a pid-keyed query kcall
   (`euid_of(pid) -> UserIdentifier`) that vfsd calls with the attested `source`
   pid. (Alternative: kernel-stamped IPC credentials on every message — broader,
   benefits all daemons, deferred.)

3. Enforce in `vfs_utimensat`. `handle_utimensat`
   (`src/daemons/vfsd/src/handler/long.rs:308`) passes the caller euid into
   `vfs_utimensat`, which applies the POSIX rule against the file's
   `st_uid`/`st_mode`: `euid == 0` → allow; `euid == st_uid` → allow;
   NULL/`UTIME_NOW` && write access → allow; else `EPERM` (`EACCES` for the
   write-access branch).

## Notes

- #3098 lands only the `vfs_utimensat` permission helper behind a `caller_euid()`
[...]

Issue #3103, opened 2026-08-14
[vfs] fstat on an open fd reports the FAT epoch, not real times


ada-x64:
Path-based `stat`/`fstatat` return real file timestamps (#3098), but `vfs_fstat`
on an open descriptor reports the FAT epoch (1980-01-01) for
`atim`/`mtim`/`ctim` (`src/libs/vfs/src/fd/mod.rs`). So `open`+`fstat` disagrees
with `stat` on the same file — a POSIX conformance gap (the two should return
identical `struct stat`).

Cause: `rust-fatfs` exposes the time getters (`modified`/`accessed`/`created`)
only on a `DirEntry`, reached by scanning the parent directory. An open `File`
has no such getter and retains no path, so `vfs_fstat` cannot reach the entry and
falls back to the epoch.

Options:

- Store the mount-relative path on the FAT file handle at open time and route
  `vfs_fstat` through the same parent-dir scan `vfs_stat` uses. Pure-nanvix, but
  a path re-lookup: breaks if the file was renamed/unlinked after `open` (real
  `fstat` is inode-based and path-independent).
- Add a time getter on the open handle in the FAT backend (needs a change to the
  `rust-fatfs` dependency, which exposes no such API today).

Related: #3098

Issue #3099, opened 2026-08-13
[vfs] Removed the hardcoded stat epoch


ada-x64:
## Problem

The same synthetic timestamp `1_704_067_200` (2024-01-01T00:00:00Z) is redefined
independently in five places across three crates, with no single source of
truth:

- `src/libs/vfs/src/fd/mod.rs:1110` — `populate_stat_fields` (FAT files/dirs)
- `src/libs/vfs/src/fd/mod.rs:1146` — `populate_pipe_stat_fields` (pipes)
- `src/libs/vfs/src/fd/mod.rs:1179` — `populate_console_stat_fields` (console)
- `src/daemons/vfsd/src/pending.rs:160` — `STAT_FIXED_EPOCH` (hostfsd path)
- `src/tests/integration/test-rust-vfs-test/src/main.rs:65` — `VFS_FIXED_TIMESTAMP`

Each is its own `const`. Changing the value means editing five sites, and the
test hardcodes the same magic number it is meant to validate.

## Proposed solution

After #3098 lands, FAT files carry real on-disk times, so the largest user
disappears. Every remaining synthetic user has a *real* time available, so the
constant can be deleted rather than centralized:

- pipes / console — capture the wall clock (`pm::__kcall_gettime`) at object
  creation, store it on the object, and report that. A pipe's `ctim` *is* when
  it was created.
- hostfsd — forward the real host `st_atim`/`st_mtim`/`st_ctim` through the
  protocol instead of dropping them and synthesizing. The current comment at
  `pending.rs:150` notes hostfsd simply does not forward these fields yet.
- mount root — real directory time / creation.

If every path has a real source, there is no fallback constant to keep in sync.

## Notes

- The hostfsd `STAT_FIXED_EPOCH` is a distinct backend from the FAT path; it is
  synthetic because hostfsd drops the host timestamps, not because FAT lacks
  them.
- Depends on a real wall clock (#668) for creation-time capture; `CLOCK_REALTIME`
  returns seconds-since-boot under linuxd/microvm, while nanvixd returns real
  time via `pm::__kcall_gettime`.
- Not bundled into #3098: that issue fixes the FAT stat/utimensat path; this one
  chases the remaining synthetic constants.

Related: #3098, #668, #3066, #1608

PR #3106, opened 2026-08-15
[vfs] Real file timestamps for stat/utimensat

### LINKED ISSUE #3098: [vfs] Persist and report real file timestamps for stat/utimensat
The VFS neither stores nor reports file timestamps. `stat`/`fstat` return a
hardcoded `FIXED_EPOCH = 1_704_067_200` (2024-01-01) for atim/mtim/ctim, and
`utimensat` writes nothing. So `os.utime(path, ...)` then `stat` always reports
the same constant, and `UTIME_NOW` is never resolved.

This is the persistence half #3066 carved out. #3066 fixed the libc layer (NULL
times no longer returns `EINVAL`); this tracks making the resulting timestamp
real. Every layer is stubbed to a constant:

- `src/libs/fat32/src/fat/time.rs:22` — `NanvixTimeProvider` always returns FAT
  epoch 1980
- `src/libs/fat32/src/fat/filesystem.rs:543` — `FatStat` carries only `size` +
  `is_dir`, no time fields
- `src/libs/vfs/src/fd/mod.rs:1108` — `populate_stat_fields` hardcodes
  `FIXED_EPOCH`
- `src/libs/vfs/src/fd/mod.rs:2129` — `vfs_utimensat` ignores `_times` (a
  validate-only stub from #1608)

Not a FAT limitation: FAT32 stores modify/access/create times, and `rust-fatfs`
already exposes them via `File::modified()` etc. and the `TimeProvider` hook we
pin to 1980. Resolution is coarse (2s write, date-only access) but fine for
"set to now".

Depends on a real wall clock (#668): `CLOCK_REALTIME` returns seconds-since-boot
under linuxd/microvm, though nanvixd appears to return real time
(`SystemTime { seconds: 1786568525 }` ≈ 2026 via `pm::__kcall_gettime`).

Fix: give `FatStat` time fields sourced from `rust-fatfs`, make `vfs_utimensat`
resolve `UTIME_NOW`/`UTIME_OMIT` against the clock and apply them, and have
`populate_stat_fields` return the stored times. Needs a `fat32` regression test
(set mtime, read it back).

Downstream: cpython `test_utime_current`/`_current_old` stay skipped until this
lands — re-tag their skips from #3066 (EINVAL) to this issue / #668.

Related: nanvix/cpython#514, nanvix/cpython#529, #668, #3066, #1608, #1761

ada-x64:
# [vfs] Real file timestamps for stat/utimensat

Closes #3098.

The VFS stubbed all file times to a fixed epoch: `stat`/`fstat` returned a
hardcoded `1704067200`, `utimensat` ignored its `times`, and the FAT
`TimeProvider` was pinned to 1980. This makes timestamps real.

## Changes

- `NanvixTimeProvider` is now wall-clock backed via `gettime`; host/`std` test
  builds fall back to the FAT epoch for determinism.
- Added Unix<->FAT date conversions (Howard Hinnant civil/days), clamped to the
  FAT year range [1980, 2107].
- `FatStat` carries `atime`/`mtime`/`ctime`; `Fat::stat` reads them from the
  directory entry and `Fat::set_times` writes them.
- Times thread through the VFS `Stat`/`VfsStat` layers; `vfs_utimensat` resolves
  `UTIME_NOW`/`UTIME_OMIT` and applies the result.
- New `fat32` round-trip tests; `test-rust-vfs-test` now sets a time and reads
  it back.

## Known limitations

- `fstat` on an open fd still reports the FAT epoch: fatfs exposes time getters
  only on a directory entry, not on an open handle. Path-based `stat` carries the
  real times. Tracked in #3103.
- Directory `utimensat` is a no-op (fatfs has no writable directory time entry).

## Follow-ups

- Per-process euid + utimensat permission enforcement (#3104, blocked by this).
- `Stat`/`VfsStat` type consolidation (drafted).
- Timestamp/epoch consolidation (#3099).

Verified against the full lifecycle.

### DIFF
--- a/src/libs/vfs/src/descriptor/vfs_stat.rs
+++ b/src/libs/vfs/src/descriptor/vfs_stat.rs
@@ -42,26 +42,31 @@ impl VfsStat {
     }
 
     /// Returns the file size in bytes.
+    #[must_use]
     pub fn size(&self) -> u64 {
         self.size
     }
 
     /// Returns whether this entry is a directory.
+    #[must_use]
     pub fn is_dir(&self) -> bool {
         self.is_dir
     }
 
     /// Returns the last access time (Unix seconds).
+    #[must_use]
     pub fn atime(&self) -> i64 {
         self.atime
     }
 
     /// Returns the last modification time (Unix seconds).
+    #[must_use]
     pub fn mtime(&self) -> i64 {
         self.mtime
     }
 
     /// Returns the creation time (Unix seconds).
+    #[must_use]
     pub fn ctime(&self) -> i64 {
         self.ctime
     }
--- a/src/libs/vfs/src/descriptor/vfs_stat.rs
+++ b/src/libs/vfs/src/descriptor/vfs_stat.rs
@@ -17,6 +17,12 @@ pub struct VfsStat {
     size: u64,
     /// Whether this entry is a directory.
     is_dir: bool,
+    /// Last access time (Unix seconds).
+    atime: i64,
+    /// Last modification time (Unix seconds).
+    mtime: i64,
+    /// Creation time (Unix seconds).
+    ctime: i64,
 }
 
 //==================================================================================================
@@ -25,8 +31,14 @@ pub struct VfsStat {
 
 impl VfsStat {
     /// Creates a new `VfsStat`.
-    pub fn new(size: u64, is_dir: bool) -> Self {
-        Self { size, is_dir }
+    pub fn new(size: u64, is_dir: bool, atime: i64, mtime: i64, ctime: i64) -> Self {
+        Self {
+            size,
+            is_dir,
+            atime,
+            mtime,
+            ctime,
+        }
     }
 
     /// Returns the file size in bytes.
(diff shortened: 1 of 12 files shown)


src/libs/vfs/src/fd/mod.rs
```diff
@@ -2134,21 +2162,47 @@ pub fn vfs_fchownat(
...
+        Ok(()) => now.seconds() as i64,
+        Err(_) => FAT_EPOCH_SECS,
+    }
+}
+
+/// Host test builds have no kernel clock; use the FAT epoch.
+#[cfg(feature = "std")]
+fn wall_clock_secs() -> i64 {
+    FAT_EPOCH_SECS
 }
```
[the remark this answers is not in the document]
ada-x64: Acknowledged, but out of scope. Deduplicating would require moving the implementation to a shared crate. Making VFS depend on the FAT helper would invert the abstraction and would not generalize to other filesystem backends.

src/libs/vfs/src/descriptor/vfs_stat.rs
ppenna, line 1: Please follow up an issue that is about merging `VfsStat` and `Stat` structures. It makes nonsense to have them replicated.

ada-x64: Already taken care of in a follow-up PR :)

https://github.com/nanvix/nanvix/pull/3110/changes#diff-2c09ed53c4abe8f7b9a276c87bcd5b8f95a865f06f26359a823c8a23879ef72d

PR #3110, opened 2026-08-18
[vfs] Replace synthetic descriptor timestamps

ada-x64:
# [vfs] Replace synthetic descriptor timestamps

Part of #3099. Depends on #3106.

Pipes and console descriptors currently report a hardcoded timestamp for every
`stat` time. Capture the wall clock when these objects are created and report
that instant instead.

## Changes

- Add a shared VFS wall-clock helper backed by `gettime`.
- Store creation time on pipe ends and console handles.
- Report creation time through `fstat` for pipes and consoles.
- Consolidate VFS-to-POSIX stat field updates behind one conversion.
- Dispatch `vfs_fstat` exhaustively by handle type.

## Known limitation

Open FAT file descriptors still report the FAT epoch because the current
upstream `fatfs::File` API does not expose directory-entry timestamps.
Re-querying by pathname would violate open-file identity after rename or unlink.


src/libs/vfs/src/fd/mod.rs
```diff
@@ -1229,32 +1215,43 @@ pub fn vfs_fstat(fd: c_int, buf: &mut ::sysapi::sys_stat::stat) -> Result<(), Fa
...
+            buf.update_from_vfs(&info);
+        },
+        VfsFileHandle::Pipe(end) => {
+            populate_pipe_stat_fields(buf, end.pipe_id(), end.created());
+        },
+        VfsFileHandle::Console(handle) => {
+            populate_console_stat_fields(buf, handle.stream(), handle.created());
+        },
+        VfsFileHandle::Socket(_) => {
+            // TODO: Report socket metadata instead of treating the descriptor as a regular file.
```
ppenna, line 1250: Please create a follow-up issue for this.

ppenna (approved):

src/libs/vfs/src/fd/mod.rs
ada-x64: https://github.com/nanvix/nanvix/issues/3114

PR #3143, opened 2026-08-27
[hostfsd] Forward Access Checks

### LINKED ISSUE #3132: [hostfs] Add permission mutation and access checks
## Context

These compiled tests are disabled under hostfs:

- `test_chmod()`
- `test_fchmodat()`
- `test_fchmod()`
- `test_lchmod()`
- `test_access()`
- `test_faccessat()`

Path operations fall through to the local FAT VFS and report `ENOENT`. Descriptor `fchmod()` is a success-returning stub. `handle_faccessat()` discards the requested mode and flags, while `vfs_accessat()` checks only existence. Hostfs has no permission or access-check requests.

`test_faccessat()` also exposes a hostfs open bug: `open(O_CREAT | O_RDONLY)` returns `EINVAL`. `test_lchmod()` unnecessarily depends on hard-link support. The standalone root identity affects expected access decisions.

References:

- https://pubs.opengroup.org/onlinepubs/9799919799/functions/fchmodat.html
- https://pubs.opengroup.org/onlinepubs/9799919799/functions/fchmod.html
- https://pubs.opengroup.org/onlinepubs/9799919799/functions/lchmod.html
- https://pubs.opengroup.org/onlinepubs/9799919799/functions/faccessat.html
- https://pubs.opengroup.org/onlinepubs/9799919799/functions/access.html

## Proposed Solution

- Add pathname and descriptor permission mutation to the hostfs protocol and backend.
- Implement `chmod()`, `fchmodat()`, `fchmod()`, and `lchmod()` over those operations.
- Implement access checks with the requested mode, flags, and intended identity semantics.
- Support `O_CREAT | O_RDONLY` in hostfs open translation.
- Make `test_lchmod()` independent of hard links.
- Strengthen `test_fchmod()` so a success-returning stub cannot pass.

## Related Issues

- #3043 tracks structural routing of resolved paths to their backend.
- #2110 tracks cleanup of permission no-op stubs.
- #3123 defines the standalone root-only identity model.
- #3104 tracks future trusted per-process identity and permission enforcement.

## Acceptance Criteria

- All six tests above pass against hostfs and perform observable checks.
- Mode changes are visible through `stat()` and `fstat()`.
- Access checks honor `F_OK`, `R_OK`, `W_OK`, and `X_OK` under the documented identity model.
- Directory-relative, no-follow, invalid-descriptor, and invalid-flag errors are covered.

## Related Pull Request

Discovered while enabling hosted C tests in #3127:
https://github.com/nanvix/nanvix/pull/3127

ada-x64:
Closes #3132.

### Summary

- Forward `access()` and `faccessat()` operations to hostfsd.
- Evaluate requested access using guest-visible owner permission bits.
- Preserve and validate `AT_EACCESS` and `AT_SYMLINK_NOFOLLOW`.
- Support `O_CREAT | O_RDONLY` on Unix while retaining a read-only descriptor.
- Re-enable the existing hostfs access tests.

### Dependencies

Stacked on the permission-mutation PR because the existing access tests use `chmod()` and `fchmodat()` to prepare their permission states.

### Verification

- Workspace checks, formatting, linting, and spellcheck pass.
- Unit tests pass.
- The focused `test-c-file` POSIX test passes.

### DIFF
--- a/src/daemons/hostfsd/src/handler.rs
+++ b/src/daemons/hostfsd/src/handler.rs
@@ -68,10 +68,18 @@ use std::{
 };
 
 #[cfg(unix)]
-use std::os::unix::fs::{
-    MetadataExt,
-    PermissionsExt,
+use std::{
+    ffi::CString,
+    os::unix::{
+        ffi::OsStrExt,
+        fs::{
+            MetadataExt,
+            PermissionsExt,
+        },
+        io::FromRawFd,
+    },
 };
+
 fn timestamp_to_system_time(seconds: i64, nanoseconds: i64) -> io::Result<SystemTime> {
     if !(0..1_000_000_000).contains(&nanoseconds) {
         return Err(io::Error::from(io::ErrorKind::InvalidInput));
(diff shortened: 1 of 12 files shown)


src/daemons/hostfsd/src/handler.rs
```diff
@@ -113,26 +121,59 @@ fn file_times(times: &[::sysapi::time::timespec; 2]) -> io::Result<FileTimes> {
...
+                let _ = mode;
+                let mut create_options: OpenOptions = options.clone();
+                create_options.create_new(true);
+                create_options.open(host_path)
+            }
+        } else {
+            let mut create_options: OpenOptions = options.clone();
+            create_options.create_new(true);
+            create_options.open(host_path)
+        };
```
ppenna, line 181: Shouldn't file creation be based on the `create` flag?

src/daemons/hostfsd/src/handler.rs
```diff
@@ -1088,6 +1138,80 @@ impl HostFsHandler {
...
+        let owner_mode: u32 = S_IRUSR
+            | if metadata.permissions().readonly() {
+                0
+            } else {
+                S_IWUSR
+            }
+            | if metadata.is_dir() { S_IXUSR } else { 0 };
+        let denied: bool = (mode & R_OK != 0 && owner_mode & S_IRUSR == 0)
+            || (mode & W_OK != 0 && owner_mode & S_IWUSR == 0)
+            || (mode & X_OK != 0 && owner_mode & S_IXUSR == 0);
```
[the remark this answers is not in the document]
ada-x64: POSIX checks each requested permission independently: `W_OK` checks write permission, while `X_OK` checks execute/search permission. Search permission is additionally required for pathname-prefix components, not implicitly for the target directory when only `W_OK` is requested. Thus a mode `0200` directory may pass `W_OK` but fail `W_OK | X_OK`, matching Unix behavior. I believe the existing check is correct.

src/daemons/hostfsd/src/handler.rs
ada-x64: It is. the `create` parameter is generated from the O_CREAT flag at handler.rs:593 (handle_long_open). That being said, the boundary here is unclear and we should not be mixing parameters like this. I've pushed an update to derive the creation flags in _this_ function instead of its caller.

Issue #3122, opened 2026-08-24
[vfs] Provide a minimal POSIX device namespace


ada-x64:
## Context

POSIX requires `/dev`, `/dev/null`, `/dev/tty`, and `/dev/console` to exist. It defines `/dev/null` as an empty data source and infinite data sink: reads return EOF and writes are discarded.

Nanvix already has device-like behavior for the console. `vfsd` seeds descriptors 0, 1, and 2 with `ConsoleHandle` values backed by a shared `LineDiscipline`, and `fstat()` reports them as character devices. The paths `/dev/null`, `/dev/tty`, and `/dev/console` do not exist, however. `ttyname_r()` returns `/dev/tty`, but opening that path fails unless a consumer happens to place a regular file there.

FAT32 cannot represent device-special files, `mkramfs` copies only directories and regular files, and `mknod()` intentionally returns `ENOTSUP`. A regular FAT file at `/dev/null` would also be incorrect because writes would consume storage and later reads could return the written data.

Reference: [POSIX Directory Structure and Devices](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap10.html#tag_10_01).

## Proposed behavior

Add a small synthetic device namespace in the VFS, independent of mounted filesystem implementations. Path resolution should recognize built-in `/dev` entries before dispatching to FAT32, hostfs, or another filesystem implementation.

Initially expose:

- `/dev/null`: reads return EOF; writes discard all bytes and report success.
- `/dev/tty`: opens the calling process's controlling terminal.
- `/dev/console`: opens the system console directly.

The current Nanvix system has one console terminal shared by the standard console handles, and `procd` already owns its foreground process-group state. The initial `/dev/tty` and `/dev/console` implementations may therefore refer to the same underlying console device while preserving their distinct semantics: `/dev/tty` is selected through the caller's controlling-terminal relationship, whereas `/dev/console` names the system console independently of standard-descriptor redirection.

The synthetic entries should produce VFS handles with behavior implemented by the device namespace. They should not implement the filesystem-backend trait used by FAT32, hostfs, or future persistent filesystems.

## Work

- Add synthetic path resolution for `/dev` and its built-in entries.
- Add a null-device handle and read/write/poll/seek behavior.
- Reuse or adapt the existing console handle so `/dev/tty` and `/dev/console` support the access modes allowed by `open()`.
[...]

PR #3148, opened 2026-09-01
[vfs] F: Add devfs

ada-x64:
## Summary

Add a synthetic `/dev` namespace owned by the VFS rather than a mounted filesystem backend.

The namespace is resolved before FAT32 or hostfs dispatch. It synthesizes metadata for `/dev`, includes it in root-directory enumeration, and prevents filesystem operations from creating, replacing, renaming, or removing reserved namespace entries.

This is the base of a stacked series. #3149 adds `/dev/null`; subsequent changes add `/dev/console` and `/dev/tty`.

## Testing

- Passed formatting, linting, and spellchecking.
- Passed compilation checks.
- Verified as the base of the stacked device implementation and its integration tests.

## Tracking

Partial implementation of #3122.

### DIFF
--- a/src/libs/vfs/src/devfs.rs
+++ b/src/libs/vfs/src/devfs.rs
@@ -7,25 +7,67 @@
 // Imports
 //==================================================================================================
 
-use crate::mount::{
-    anchor_path,
-    normalize_anchored,
+use crate::{
+    filesystem::{
+        DirEntry,
+        Stat,
+    },
+    mount::{
+        anchor_path,
+        normalize_anchored,
+    },
 };
 use ::alloc::{
     string::String,
     vec::Vec,
 };
-use ::fat32::Fat32Error;
+use ::fat32::{
+    Fat32Error,
+    FAT_EPOCH_SECS,
+};
+use ::sys::pm::{
+    GroupIdentifier,
+    UserIdentifier,
+};
+use ::sysapi::{
+    sys_stat::{
+        file_mode,
+        file_type,
+        stat as PosixStat,
+    },
+    sys_types::{
+        gid_t,
+        uid_t,
+    },
+    time::timespec,
+};
 
 //==================================================================================================
 // Constants
 //==================================================================================================
 
-/// Synthetic device identifier for the device namespace.
+/// Synthetic device identifier for devfs.
+///
+/// IDs 1 through 3 identify the VFS file namespace, pipefs, and console.
 const DEVICE_NAMESPACE_ID: u64 = 4;
 
+/// Name of the devfs root directory.
+const DIRECTORY_NAME: &str = "dev";
+/// Absolute path of the devfs root directory.
+const DIRECTORY_PATH: &str = "/dev";
+/// Absolute prefix for devfs entries.
+const DIRECTORY_PREFIX: &str = "/dev/";
+
 /// Stable inode identifier for `/dev`.
-pub(crate) const DIRECTORY_INODE: u64 = 1;
+///
+/// Inode zero is reserved, so `/dev` uses the first available identifier.
+const DIRECTORY_INODE: u64 = 1;
+
+/// Preferred I/O block size reported for devfs entries.
+const STAT_BLOCK_SIZE: i64 = ::arch::mem::PAGE_SIZE as i64;
+
+/// Stable timestamp used until VFS defines backend-neutral synthetic timestamps.
+const STAT_TIMESTAMP_SECS: i64 = FAT_EPOCH_SECS;
 
 //==================================================================================================
 // Structures
@@ -33,7 +75,7 @@ pub(crate) const DIRECTORY_INODE: u64 = 1;
 
 /// Metadata for a synthetic device-namespace entry.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
-pub(crate) struct DeviceMetadata {
+struct DeviceMetadata {
     /// Synthetic device identifier.
     device: u64,
     /// Stable inode identifier.
@@ -59,26 +101,9 @@ pub(crate) enum DevicePath {
 // Implementations
 //==================================================================================================
 
-impl DeviceMetadata {
-    /// Returns the synthetic device identifier.
-    pub(crate) fn device(&self) -> u64 {
-        self.device
-    }
-
-    /// Returns the stable inode identifier.
-    pub(crate) fn inode(&self) -> u64 {
-        self.inode
-    }
-
-    /// Returns whether this entry is a directory.
-    pub(crate) fn is_directory(&self) -> bool {
-        self.is_directory
-    }
-}
-
 impl DevicePath {
     /// Returns metadata for an existing namespace entry.
-    pub(crate) fn metadata(self) -> Option<DeviceMetadata> {
+    fn metadata(self) -> Option<DeviceMetadata> {
         match self {
             DevicePath::Directory => Some(DeviceMetadata {
                 device: DEVICE_NAMESPACE_ID,
@@ -95,7 +120,7 @@ impl DevicePath {
 //==================================================================================================
 
 /// Resolves a path in the synthetic device namespace.
-pub(crate) fn resolve(path: &str, cwd: &str) -> Result<Option<DevicePath>, Fat32Error> {
+pub(crate) fn resolve(cwd: &str, path: &str) -> Result<Option<DevicePath>, Fat32Error> {
     let anchored: String = anchor_path(path, cwd)?;
     validate_components(&anchored)?;
     let normalized: String = normalize_anchored(&anchored);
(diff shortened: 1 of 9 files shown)


src/libs/vfs/src/filesystem.rs
```diff
@@ -103,6 +112,15 @@ pub(crate) fn file_raw_region(cwd: &str, path: &str) -> Option<(*const u8, usize
...
     let requires_dir: bool = path.ends_with('/');
+    if let Some(metadata) = device_namespace::metadata(path, cwd)? {
+        return Ok(Stat::new(
+            0,
+            metadata.is_directory(),
+            FAT_EPOCH_SECS,
+            FAT_EPOCH_SECS,
+            FAT_EPOCH_SECS,
+        ));
+    }
```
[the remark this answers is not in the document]
ada-x64: Addressed in https://github.com/nanvix/nanvix/pull/3149

src/libs/vfs/src/fd/mod.rs
```diff
@@ -1221,6 +1237,29 @@ fn populate_console_stat_fields(
     };
 }
 
+/// Populates stat fields for a synthetic device-namespace entry.
+fn populate_device_stat_fields(buf: &mut ::sysapi::sys_stat::stat, metadata: DeviceMetadata) {
+    buf.st_size = 0;
+    buf.st_nlink = if metadata.is_directory() { 2 } else { 1 };
+    buf.st_dev = metadata.device();
+    buf.st_ino = metadata.inode();
+    buf.st_mode = file_type::S_IFDIR | file_mode::S_IRWXU;
```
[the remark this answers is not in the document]
ada-x64: Addressed in https://github.com/nanvix/nanvix/pull/3149

src/libs/vfs/src/filesystem.rs
```diff
@@ -280,26 +313,45 @@ pub(crate) fn unlink(cwd: &str, path: &str) -> Result<(), Fat32Error> {
...
+                .into_iter()
+                .enumerate()
+                .map(|(index, entry)| {
+                    DirEntry::new(entry.name, index as u64 + 1, entry.is_dir, entry.size)
+                })
+                .collect())
+        })?,
+        Err(Fat32Error::NotFound) if normalized == "/" => Vec::new(),
+        Err(error) => return Err(error),
+    };
```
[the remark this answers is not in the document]
ada-x64: /dev has a distinct st_dev than the FAT mount (it's a different namespace), so collision is irrelevant.

src/libs/vfs/src/devfs.rs
```diff
@@ -0,0 +1,187 @@
...
+    vec::Vec,
+};
+use ::fat32::Fat32Error;
+
+//==================================================================================================
+// Constants
+//==================================================================================================
+
+/// Synthetic device identifier for the device namespace.
+const DEVICE_NAMESPACE_ID: u64 = 4;
```
ppenna, line 52: Can you add rationale behind the value `4`?

src/libs/vfs/src/devfs.rs
```diff
@@ -0,0 +1,187 @@
...
+
+/// Stable inode identifier for `/dev`.
+pub(crate) const DIRECTORY_INODE: u64 = 1;
```
ppenna: Add rationale saying that inode 0 is reserved.

src/libs/vfs/src/filesystem.rs
```diff
@@ -280,26 +313,45 @@ pub(crate) fn unlink(cwd: &str, path: &str) -> Result<(), Fat32Error> {
...
+    if normalized == "/" {
+        entries.retain(|entry| entry.name() != "dev");
+        entries.push(DirEntry::new(
+            String::from("dev"),
```
ppenna: We should make `dev` as a const string.

src/libs/vfs/src/filesystem.rs
```diff
@@ -103,6 +112,15 @@ pub(crate) fn file_raw_region(cwd: &str, path: &str) -> Option<(*const u8, usize
...
 pub(crate) fn stat(cwd: &str, path: &str) -> Result<Stat, Fat32Error> {
     let requires_dir: bool = path.ends_with('/');
+    if let Some(metadata) = device_namespace::metadata(path, cwd)? {
+        return Ok(Stat::new(
+            0,
+            metadata.is_directory(),
+            FAT_EPOCH_SECS,
+            FAT_EPOCH_SECS,
+            FAT_EPOCH_SECS,
+        ));
```
ppenna: Why not synthesize this on `devfs` itself and expose helper?

src/libs/vfs/src/fd/mod.rs
```diff
@@ -1221,6 +1237,29 @@ fn populate_console_stat_fields(
...
+    buf.st_mtim = timespec {
+        tv_sec: FAT_EPOCH_SECS,
+        tv_nsec: 0,
+    };
+    buf.st_ctim = timespec {
+        tv_sec: FAT_EPOCH_SECS,
+        tv_nsec: 0,
+    };
+}
+
```
ppenna: This belongs on `devfs`, no?

src/libs/vfs/src/devfs.rs
```diff
@@ -0,0 +1,287 @@
...
+/// Stable inode identifier for `/dev`.
+///
+/// Inode zero is reserved, so `/dev` uses the first available identifier.
+const DIRECTORY_INODE: u64 = 1;
+
+/// Preferred I/O block size reported for devfs entries.
+const STAT_BLOCK_SIZE: i64 = ::arch::mem::PAGE_SIZE as i64;
+
+/// Stable timestamp used until VFS defines backend-neutral synthetic timestamps.
+const STAT_TIMESTAMP_SECS: i64 = FAT_EPOCH_SECS;
```
ada-x64, line 70: Kept the existing FAT epoch behavior for now, but isolated it behind a named devfs timestamp policy and documented why. We should open a follow-up issue for a backend-neutral synthetic timestamp after this stack lands.

src/libs/vfs/src/devfs.rs
```diff
@@ -0,0 +1,287 @@
...
+        st_blksize: STAT_BLOCK_SIZE,
+        st_blocks: 0,
+        st_atim: timestamp,
+        st_mtim: timestamp,
+        st_ctim: timestamp,
+    }))
+}
+
+/// Returns the `/dev` entry injected into the VFS root directory.
+pub(crate) fn directory_entry() -> DirEntry {
```
ada-x64, line 192: Moved root-entry construction and devfs directory enumeration behind devfs helpers. The directory name, inode, and internal metadata representation can now remain private instead of being reconstructed by `filesystem`.

src/libs/vfs/src/devfs.rs
```diff
@@ -0,0 +1,287 @@
...
+/// Absolute prefix for devfs entries.
+const DIRECTORY_PREFIX: &str = "/dev/";
+
+/// Stable inode identifier for `/dev`.
+///
+/// Inode zero is reserved, so `/dev` uses the first available identifier.
+const DIRECTORY_INODE: u64 = 1;
+
+/// Preferred I/O block size reported for devfs entries.
+const STAT_BLOCK_SIZE: i64 = ::arch::mem::PAGE_SIZE as i64;
```
ada-x64, line 67: Derived `st_blksize` from `arch::mem::PAGE_SIZE` instead of embedding 4096, so the reported block size follows the target architecture.

src/libs/vfs/src/devfs.rs
```diff
@@ -0,0 +1,287 @@
...
+        0,
+        metadata.is_directory,
+        STAT_TIMESTAMP_SECS,
+        STAT_TIMESTAMP_SECS,
+        STAT_TIMESTAMP_SECS,
+    )))
+}
+
+/// Synthesizes POSIX metadata for an existing devfs path.
+pub(crate) fn posix_stat(cwd: &str, path: &str) -> Result<Option<PosixStat>, Fat32Error> {
```
ada-x64, line 166: Changed devfs POSIX metadata synthesis to return a complete `stat`, including ownership, rather than mutating part of a caller-initialized buffer. This gives the complete representation one owner.

src/libs/vfs/src/devfs.rs
```diff
@@ -0,0 +1,287 @@
...
+        }
+    }
+}
+
+//==================================================================================================
+// Standalone Functions
+//==================================================================================================
+
+/// Resolves a path in the synthetic device namespace.
+pub(crate) fn resolve(cwd: &str, path: &str) -> Result<Option<DevicePath>, Fat32Error> {
```
ada-x64, line 123: Standardized the devfs API on the surrounding VFS convention of `(cwd, path)`. Since both parameters are `&str`, consistent ordering prevents argument reversals that would otherwise compile.

Issue #3121, opened 2026-08-24
[testing] Remove __NANVIX_STANDALONE__ C guard


ada-x64:
## Context

We now _only_ support standalone mode. We need to strip out the __NANVIX_STANDALONE__ guard and ensure that everything still operates smoothly.

PR #3155, opened 2026-09-08
[tests] Restore Unix Socket Tests

ada-x64:
## Summary

- Restore the missing Unix-domain socket C tests.
- Run pathname-based socket coverage through hostfs on supported hosts.
- Use routed socket I/O for connected socket pairs.

## Testing

Verified against the full Linux and Windows lifecycles.

## Related

Part of #3121.

Stacked on #3135.

### DIFF
--- a/src/tests/integration/test-c-network/common.c
+++ b/src/tests/integration/test-c-network/common.c
@@ -12,7 +12,6 @@
 #include <netinet/in.h>
 #include <string.h>
 #include <sys/socket.h>
-#include <sys/un.h>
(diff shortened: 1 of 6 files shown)


src/tests/integration/test-c-network/common.c
ppenna, line 1: @ada-x64 where were these removed?

ada-x64: This is pathname cleanup for the socket tests. It's been moved to `unix.c` lines 126-144

Issue #3160, opened 2026-09-09
[ipc] Add reliable correlated transactions


ada-x64:
## Summary

Nanvix implements several cross-daemon state transitions as correlated request/acknowledgement protocols. Current examples include:

- cloning VFS state during `fork()`;
- applying close-on-exec during `exec()`;
- detaching the controlling terminal during `setsid()`;
- forwarding VFS operations to hostfsd.

These protocols independently retain caller state and match replies with request identifiers, but local IPC does not provide a reliable transaction abstraction. A successful send establishes only that the kernel queued a message. It does not establish that the responder processed it or that the acknowledgement reached the requester.

If a responder exits, hangs, or applies an operation but fails to send its acknowledgement, a caller may remain blocked indefinitely. Retrying is not generally safe because the requester cannot distinguish an operation that was not processed from one that committed but lost its acknowledgement.

## Desired behavior

Introduce a reusable correlated IPC transaction abstraction with explicit outcomes for:

- response received;
- request not dispatched;
- responder terminated before completion;
- deadline or cancellation after the request may have been processed;
- malformed or unauthenticated response.

The API must distinguish definite failure before dispatch from indeterminate completion after dispatch. An indeterminate operation must not be reported as safely retryable unless its protocol is explicitly idempotent.

## Required invariants

- [ ] Completion state is reserved before a request becomes visible to its responder.
- [ ] Pending transactions have a static resource limit.
- [ ] Responses are matched by request identifier and authenticated responder identity.
- [ ] Responses retain the originating process and thread routing context.
- [ ] Stale, duplicate, and forged responses cannot complete another transaction.
- [ ] Responder termination resolves affected callers instead of leaving them blocked forever.
- [ ] Caller termination reclaims pending state without affecting another caller.
- [ ] Request identifier reuse cannot let an old response complete a newer transaction.
- [ ] Timeout and cancellation expose indeterminate completion where appropriate.
- [ ] Retried operations are idempotent or expose a way to query their committed result.

## Proposed direction

Define a common transaction lifecycle resembling:

```text
reserve(responder, deadline) -> transaction
send(transaction, request) -> dispatched | not_dispatched
receive(transaction) -> reply | peer_closed | deadline_expired | cancelled
reply(transaction, response)
```

[...]

PR #3161, opened 2026-09-09
Add /dev/tty 

ada-x64:
## Summary

Add `/dev/tty` as an alias for the calling process's controlling terminal.

- expose `/dev/tty` through devfs;
- preserve controlling-terminal association across `fork()`;
- detach a new session from its controlling terminal during `setsid()`;
- return `ENXIO` when a process without a controlling terminal opens `/dev/tty`;
- reject terminal process-group operations from detached sessions with `ENOTTY`;
- add namespace, open-mode, metadata, detachment, and error-path coverage.

## Workaround

`setsid()` currently uses a dedicated procd-to-vfsd acknowledgement barrier. This ensures the call does not return before vfsd has detached `/dev/tty`, but it duplicates transaction machinery used by other cross-daemon operations and does not fully address lost acknowledgements or responder failure.

Generic reliable IPC transactions are tracked in
https://github.com/nanvix/nanvix/issues/3160.

## Testing

- Full build
- Full test suite
- Format, lint, and spell checks


src/libs/vfs/src/devfs.rs
```diff
@@ -80,16 +80,27 @@ const NULL_INODE: u64 = 2;
...
+const TTY_SPECIAL_DEVICE_ID: u64 = 2;
+
 /// Name of the console device.
 const CONSOLE_NAME: &str = "console";
 /// Absolute path of the console device.
 const CONSOLE_PATH: &str = "/dev/console";
 
 /// Stable inode identifier for `/dev/console`.
-const CONSOLE_INODE: u64 = 3;
+const CONSOLE_INODE: u64 = 4;
```
ppenna, line 100: These are hard-coded everywhere... We need to create an enum to back this static allocation. Can you create a follow-up issue for that?

ppenna (approved):

Issue #3101, opened 2026-08-13
[vfs] Consolidate path validation into typed resolution


ada-x64:
## Context

Path validation is scattered across every layer. A single `*at` call is
anchored, normalized, mount-matched, routed, and stat-checked in five different
files, and most of that work is redone 2–3 times because the only thing tying it
together is a bare `String`. `ResolvedPath(String)` (#3042) marks provenance but
nothing else, so `vfs_open_resolved` just calls `vfs_open(path.as_str())` and
re-resolves from scratch.

Four distinct concerns are tangled into that one string:

- **provenance** — the `(dirfd, path)` pair came from a syscall and travels
  together, still *unresolved*: relative to its `dirfd`, not yet joined.
- **normalization** — lexical `.`/`..`/`//` collapse. Correct for FAT, *wrong*
  for hostfs (hostfsd resolves `..` physically, against real symlinks).
- **backend routing** — `is_hostfs_path`, a `/mnt` prefix test repeated at nine
  call sites in `hostfs_handlers.rs`. hostfs is a *sidechannel* backend handled
  in `vfsd`, not a true VFS backend (those are the in-`vfs` impls like fatfs);
  routing only decides which of the two a path takes.
- **permissions** — splits in two: mount capability (RO/RW, known from the mount
  table, synchronous) and object metadata (exists, is-dir, ancestor-is-dir,
  ownership), which needs `stat` — and for hostfs that is an async IPC round trip
  through the pending queue, so it cannot run inside a resolution lock scope.

Because these share one type, the wrong policy can be applied to the wrong
backend and only fails at the point of use (a hostfs path reaching the FAT layer
compiles and returns `NotFound` at runtime). This is the groundwork #3043 builds
on: it names the same problem, but its fix — folding routing into a single
backend-tagged enum in `vfs` — is the *move* this issue deliberately leaves
alone (see Out of scope).

## Proposed solution

Split the one string into a short pipeline of types, each carrying exactly one
proof, so the scattered checks collapse into construction:

```
(dirfd, path)  from syscall
  ▼
AnchoredPath { dirfd, path }              vfs; raw provenance, NOT resolved
  ▼   vfsd routes on the joined path (is_hostfs_path):
  ├─ hostfs sidechannel ─▶ HostResolvedPath(String)  hostfs-api (+ i32 errno) ─▶ hostfsd
[...]

PR #3167, opened 2026-09-14
[vfs] Add typed path resolution foundations

ada-x64:
## Summary

Introduce the first stage of typed path resolution for #3101.

This PR establishes backend-owned path proof types and centralizes VFS anchoring, normalization, mount matching, and devfs classification. Existing filesystem consumers remain buildable through narrow compatibility adapters and will be migrated in follow-up PRs.

This is the first PR in the planned rollout for #3101.

## Changes

- Add fat32::FatResolvedPath.
  - Private representation.
  - Checked construction of normalized FAT-local paths.
  - Supports the existing empty-string mount-root representation.
  - Rejects absolute, unnormalized, and otherwise invalid FAT-local paths.

- Add hostfs_api::HostResolvedPath.
  - Private representation.
  - Recognizes only /mnt and /mnt/... namespace boundaries.
  - Preserves the hostfs-relative tail verbatim.
  - Provides checked wire decoding using the existing errno convention.

- Add typed VFS path provenance through AnchoredPath.
  - Preserves the raw dirfd and path supplied at the syscall boundary.
  - Classifies absolute, AT_FDCWD-relative, and directory-fd-relative paths as explicit private states.
  - Defers empty-path and descriptor validation until resolution.
  - Preserves ENOENT, EBADF, and ENOTDIR behavior.

- Add a checked VFS resolver core.
  - Anchors each input once.
  - Performs FAT lexical normalization.
  - Uses longest-prefix mount matching.
  - Produces FatResolvedPath.
  - Preserves exact mount-root behavior.
  - Preserves existing PathCache keys, hits, and mount-table invalidation.

- Add anchored devfs resolution.
  - Preserves devfs-before-FAT precedence.
  - Consolidates anchoring, traversal validation, normalization, and classification.
  - Preserves DevicePath variants, metadata behavior, trailing-slash handling, directory reads, and concrete-device traversal errors.
  - Verifies that devfs matches do not populate the FAT cache and that FAT fallthrough retains cache hit/miss parity.

## Resolution model

    (dirfd, raw path)
            |
            v
       AnchoredPath
            |
            +----> devfs validation/normalization ----> DevicePath
            |
            +----> VFS normalization/mount match -----> FatResolvedPath

HostResolvedPath is introduced here as the checked hostfs proof type, but the vfsd/hostfsd routing and IPC migration is intentionally deferred to a later stage.

## Temporary compatibility

The following compatibility surfaces remain intentionally and are marked for removal under #3101:

- vfs_resolve_path() absolute-string projection
- Vfs::resolve_legacy()
- devfs::resolve()
- devfs::owns()
- private devfs metadata adapter
- devfs::stat()
- devfs::posix_stat()
- devfs::read_dir()
- the temporary dead-code allowance on devfs::resolve_anchored()

These adapters delegate to the new shared resolution paths and do not feed proved backend paths back through anchoring or mount matching.

## Out of scope

This PR does not:

- Migrate the twelve VFS/FAT filesystem operations.
- Migrate vfsd hostfs routing, pending state, IPC, or hostfsd sandbox entry.
- Remove all generic/string compatibility APIs.
- Introduce a backend routing enum or redesign PathCache.
- Move metadata or object checks into path constructors.

Those changes are reserved for the remaining rollout stages.


src/libs/fat32/src/path.rs
```diff
@@ -0,0 +1,110 @@
...
+    ///
+    /// A checked resolved path. The empty path represents the filesystem root.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Fat32Error::InvalidPath`] if `path` is absolute, contains a null byte, or contains
+    /// empty, current-directory, or parent-directory components.
+    pub fn new(path: String) -> Result<Self, Fat32Error> {
+        if path.contains('\0')
+            || path.starts_with('/')
```
ppenna, line 63: Can't absolute paths start with `/`?

ada-x64: This type describes file paths that are already resolved relative to a mount, with the prefix stripped. Typical usage looks like this:

```rust
let request: AnchoredPath =
    AnchoredPath::new(AT_FDCWD, String::from("/data/directory/file"))?;

let resolved: FatResolvedPath = vfs.resolve(request)?;
assert_eq!(resolved.as_str(), "directory/file");
```

Typically we're not directly invoking this but rather getting it from the `AnchoredPath` type.

Also added some more doc comments and a test to explain some tricky semantics.

ppenna: Docs helped.

ppenna (approved):

